PR #1651, opened 2023-01-22
Wallet refactors

ap-justin:
With error handling refactors
* #1712 

### Notable changes
* fully hardcoded `chain` data since most of it is hardcoded anyway `name` `chainId`, `blockExplorer`. `rpc` and `lcd` doesn't need to be in sync with what's on AWS. Tokens on the other hand, should stay there for values like `approved` `min_donation_amount`, and tokens supported.

```ts

export type Chain = {
  type: ConnectedToChainType["type"]; //to determine which type tx to perform
  name: string;
  rpc: string;
  lcd: string;
  txExplorer: string;
};

```
* Wallet context interface was simplified from 
```ts
export type WalletState = undefined | {
  walletIcon: string;
  displayCoin: Token;
  coins: Token[];
  giftcardCoins: Token[];
  address: string;
  chain: Chain;
  providerId: ProviderId;
  supportedChains: BaseChain[];
};
```

to 
```ts
export type ContextState =
  | "loading" /** consolidate all LoadingWallet*/
  | ConnectedWallet
  | DisconnectedWallet[];

```

* wallet hooks interface simplified from 

```ts
 {
    connection,
    disconnect,
    switchChain,
    isLoading,
    providerInfo,
    supportedChains: SUPPORTED_CHAINS,
  }

```
to 

```ts
export type WalletState = ConnectedToChainType | Disconnected | Loading;

//where Connected pertains to
type Connected = {
  status: "connected";
  address: string;
  chainId: string;
  disconnect(): void;
};

```
this means, data like `balances`, `chain-info` (which makes connecting slow) is not part of wallet connection process anymore. Instead these are retrieved by components that need them.

* Created `<WalletGuard/>` component that provides asserted type `ConnectedWallet` and exposes customizable fallback for `loading` `disconnected` or `chain_unsupported`

* Smart contract classses now only accepts pre-asserted `CosmosWallet` instead of `WalletState | undefined`

* Admin transactors now uses pre-asserted (in Admin/Guard ) wallet instead of raw wallet state from WalletContext

* `<ChainSelector/>` is utilized both on `on-demand` chain switching and in cases where user connected to the wrong network. That is, instead of `error` and disconnecting the wallet. The user has a chance to switch to the correct chain.

### Packages
* added `react-toastify` to top level hooks like in `WalletContext` may show errors first hand and not rely on `ModalContext`
* remove `ethers` and instead replaced with individual modules. This removed the need for `util` polyfill. Also, bundle was reduced by `20kb`
* remove most cosmjs packages and just use more browser friendly keplr wallet apis  + lcd api   `-400kb`
* remove most @ethersproject packages and just use eip1193 ( injecteced provider ) + eip1484 (json rpc specs) ` -200kb + removal of web3-engine node polyfills`

### TESTS
* Wallet suite on all wallet types
 - [x] connect
 - [x] disconnect
 - [x] select wallet
* Donate - improved types
  - [x] juno native
  - [x] juno ibc
  - [x] terra native
  - [x] terra ibc
  - [x] terra cw20
  - [x] evm native
  - [x] evm erc20
  - [x] any with KYC 

* Registration
 - [x] wallet registration
 
 * Contracts
 - [x] multi msgs
 - [x] single msg


ap-justin:
will reopen once issue with keplr `cosmos_signDirect` is resolved

PR #1783, opened 2023-02-15
Update Active Countries to be a non-required field in both Registration (added as new field) and in Update profile

misicnenad:
Ticket(s):
- https://github.com/AngelProtocolFinance/angelprotocol-web-app/issues/1754
- https://github.com/AngelProtocolFinance/angelprotocol-web-app/issues/1752

Based on changes found in @SovereignAndrey's draft PR https://github.com/AngelProtocolFinance/angelprotocol-web-app/pull/1756/files

- start registration process http://localhost:4200/register
- get to Documentation step (step 2)
- verify a non-required "Active Countries" field is present
- complete step 2
- verify submitting is successful with or without the "Active Countries" filled
- connect Charity 1 wallet
- go to Edit profile page http://localhost:4200/admin/1/edit-profile
- verify "Active Countries" field is not required anymore

## UI changes for review

image
image


SovereignAndrey (approved):

src/pages/Registration/Steps/Documentation/schema.ts
```diff
@@ -42,4 +42,5 @@ export const schema = Yup.object().shape<SchemaShape<FormValues>>({
   hasAgreedToTerms: Yup.bool().isTrue(
     "Please confirm that you agree to our Terms and Conditions"
   ),
+  activeInCountriesOpts: Yup.array(),
```
ap-justin, line 45: this is not needed, but in case we still want to add this rule, we should at least provide error message `active countries must be array` so that user would see some reason just in case this rule is preventing the form from being submitted - though I'm not sure this error would make sense to user since this  `activeInCountries` not being an `array` is development error

ap-justin (changes requested):

src/pages/Registration/Steps/Documentation/schema.ts
misicnenad: This would be a perfect use case for some logger service, storing info/errors/warnings on all things going on in the webapp (some of which we do not want to necessarily show to the user).
Even though this would be a "development error", it would still even make sense to display an error message here if for nothing else, then to be able to catch bugs *faster* by having the app immediately display the error to us devs or to users (as opposed to having users being unable to submit some form due to some ninja bug and us having to spend time searching for the source).

SovereignAndrey: @nenad That's def a fair point. Having the check doesn't hurt anything. AWS would have its own validation checks too as a fallback, but defining that here could be good to define that explicitly.

nenad: Wrong Nenad. @misicnenad

PR #1840, opened 2023-02-21
Add transition to backdrop and Dialog

### LINKED ISSUE #1827: Add Transition to Modal (+ backdrop) for smoother UI
CU #: 33t9cmv - https://app.clickup.com/t/33t9cmv
Created on: Nov 16 2022 
Original Creator: Nenad Misic

## Description
<p>See <a href="https://headlessui.com/react/dialog" target="_blank">HeadlessUI Dialog example</a> for inspiration</p>

abbasali1994:
Closes #1827 

-

## UI changes for review

### DIFF
--- a/src/components/Panel/index.tsx
+++ b/src/components/Panel/index.tsx
@@ -1,0 +1,18 @@
+import { Dialog, Transition } from "@headlessui/react";
+import React, { FC } from "react";
+
+const Panel: FC<any> = (props) => (
+  <Transition.Child
+    as={React.Fragment}
+    enter="transform ease-out duration-300"
+    enterFrom="opacity-0"
+    enterTo="opacity-100"
+    leave="transform ease-in duration-300"
+    leaveFrom="opacity-100"
+    leaveTo="opacity-0"
+  >
+    <Dialog.Panel {...props}>{props.children}</Dialog.Panel>
+  </Transition.Child>
+);
+
+export default Panel as typeof Dialog.Panel;
--- a/src/contexts/ModalContext.tsx
+++ b/src/contexts/ModalContext.tsx
@@ -93,19 +93,7 @@ export default function ModalContext(
               aria-hidden="true"
             />
           </Transition.Child>
-          <Transition.Child
-            as={Fragment}
-            enter="ease-out duration-300"
-            enterFrom="opacity-0"
-            enterTo="opacity-100"
-            leave="ease-in duration-300"
-            leaveFrom="opacity-100"
-            leaveTo="opacity-0"
-          >
-            <div>
-              {state?.Modal /** should always be wrapped with Dialog.Panel */}
-            </div>
-          </Transition.Child>
+          {state?.Modal /** should always be wrapped with Panel */}
         </Dialog>
       </Transition>
       {props.children}
--- a/src/contexts/ModalContext.tsx
+++ b/src/contexts/ModalContext.tsx
@@ -1,5 +1,6 @@
-import { Dialog } from "@headlessui/react";
+import { Dialog, Transition } from "@headlessui/react";
 import {
+  Fragment,
   PropsWithChildren,
   ReactNode,
   createContext,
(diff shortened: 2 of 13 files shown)


src/contexts/ModalContext.tsx
```diff
@@ -76,11 +77,37 @@ export default function ModalContext(
...
+          <Transition.Child
+            as={Fragment}
+            enter="ease-out duration-300"
+            enterFrom="opacity-0"
+            enterTo="opacity-100"
+            leave="ease-in duration-300"
+            leaveFrom="opacity-100"
+            leaveTo="opacity-0"
+          >
+            <div>
```
ap-justin, line 105:
```suggestion
        
```

src/contexts/ModalContext.tsx
```diff
@@ -76,11 +77,37 @@ export default function ModalContext(
...
+            leaveFrom="opacity-100"
+            leaveTo="opacity-0"
+          >
+            <div
+              className="z-10 fixed inset-0 bg-black/50"
+              aria-hidden="true"
+            />
+          </Transition.Child>
+          <Transition.Child
+            as={Fragment}
```
ap-justin, line 97:
```suggestion

```

ap-justin (changes requested):
`<Transition.Child>{panel}<Transition.Child>` might be suboptimal since `<Transition.Child/>` is already rendered (along with ModalContext)- and not along with `Dialog.Panel` (injected)

better would be to create a component wrapper 
```
function Panel(){
  return <Transtion><Dialog.Panel>{children}</Dialog.Panel></Transition>
}
```
and replace every `Dialog.Panel` with HOC `<Panel/>`

this way no need for additonal div `<div>{state.Modal}</div>`

src/contexts/ModalContext.tsx
abbasali1994: I will check this 
 I tried a similar approach but it didnt work cause since transition.child wasnt rendered by default on render transition effect didnt occur, 
 I will try the suggested approach

abbasali1994: @ap-justin tried your approach but faced similar issue, since Transition.Child is not rendered already the transition effect is not visible , I think for now we need to go with this approach

ap-justin: @abbasali1994, not sure why it's not working for you - works fine with me 

the extra `div` is problematic
* in the current solution, `backdrop` seem to go in front of `modal` and transition is not smooth
https://www.loom.com/share/3d1f7a2f4b1d4b979195371c4de837d2
* when applied with other transition effects e.g. `scale` , modal breaks
https://www.loom.com/share/29374750e0df4fa0b8c6c3412fe3006e

working when `<Transition/>` is rendered along with `<Dialog.Panel/>`
* opacity transition accurate
https://www.loom.com/share/015763346e92461cb2bbe48f09a37b7c

* scale transition correct
https://www.loom.com/share/963923e149e74a1bbfe5a5ddf313b808

plus to extra `div` which is kind of out of place, since most `panels` are `fixed` positioned
```
<div>
 <modal fixed />
</div>
```

ap-justin: @abbasali1994, see example #1850

SovereignAndrey (dismissed):

src/contexts/ModalContext.tsx
abbasali1994: @ap-justin yes justin i figured out what i was doing out wrong yesterday and updated the PR sorry didnt push the commit, 
i hv taken reference from your PR and modified it furthur, Please check now

ap-justin (approved):

src/components/Panel/index.tsx
```diff
@@ -0,0 +1,18 @@
+import { Dialog, Transition } from "@headlessui/react";
+import React, { FC } from "react";
+
+const Panel: FC<any> = (props) => (
```
misicnenad, line 4: nit: `Panel` doesn't need to be in a separate folder, e.g. no need for a `/components/Panel/index.tsx` structure. It can just be placed in a file `/components/Panel.tsx`.

SovereignAndrey (approved):
Nice work on this @abbasali1994! :smile:

src/components/Panel/index.tsx
misicnenad, line 4: @abbasali1994 @ap-justin I would suggest we align naming that we use for modals.

We call these "modals", in our code we have "ModalContext", "useModalContext", "show/closeModal", "KadoModal", "TxModal" etc, so we should keep this naming convention when referring to related components -> this component should be renamed `"Panel" -> "Modal"`.

misicnenad (changes requested):

PR #1841, opened 2023-02-21
Disable Continue unless agreed to TnC

### LINKED ISSUE #1825: Disable "Continue" button on Donate KYC step if "terms & conditions" aren't accepted
CU #: 37q4wkj - https://app.clickup.com/t/37q4wkj
Created on: Dec 6 2022 
Original Creator: Nenad Misic

## Description
<p></p>

abbasali1994:
Closes #1825 

-

## UI changes for review

### DIFF
--- a/src/components/KYC/Form/Controls.tsx
+++ b/src/components/KYC/Form/Controls.tsx
@@ -65,7 +65,12 @@ export default function Controls({
           Continue
         </button>
       ) : (
-        <button className="btn-orange btn-donate" type="button" onClick={skip}>
+        <button
+          className="btn-orange btn-donate"
+          type="button"
+          onClick={skip}
+          disabled={!hasAgreedToTerms}
+        >
           Continue
         </button>
       )}


src/components/KYC/Form/Controls.tsx
```diff
@@ -65,7 +65,12 @@ export default function Controls({
           Continue
         </button>
       ) : (
-        <button className="btn-orange btn-donate" type="button" onClick={skip}>
+        <button
+          className="btn-orange btn-donate"
+          type="button"
+          onClick={skip}
+          disabled={!hasAgreedToTerms}
```
ap-justin, line 72: @abbasali1994, @SovereignAndrey 

there's already existing check for this - which button is `enabled` at first which submitting without clicking the checkbox would make it focus there
image


I believe this is better UX than button `disabled` at first and user has to figure 
if need to fill-out form or not 
Screenshot 2023-02-21 at 6 40 21 PM


if user starts filling-out form and the endowment doesn't require KYC, the form would sense that user wants to provide KYC anyway so it will make all the fields required
Screenshot 2023-02-21 at 6 44 04 PM


 which could be prevented if 
 * user just clicks `submit`
 * if KYC not required, get focused on `check-box`
 * if KYC required, gets prompted with required fields

ap-justin (changes requested):

src/components/KYC/Form/Controls.tsx
SovereignAndrey: @ap-justin The current logic on the site is not what reflective of the latest requirements and real use has shown there are issues with it. Users have reported back that the existing logic results in them missing the T&C box and just clicking submit. This ticket will help cue users in to the fact that they need to do _something_ in order to move forward by having the submit button disabled to start.

SovereignAndrey (approved):

ap-justin (approved):

PR #1851, opened 2023-02-23
Display `tokens_on_hand` Balance in Endow Profile Page

### LINKED ISSUE #1843: The example charity current balance is different on marketplace page than in actual account
see the loom video here - https://www.loom.com/share/b6e63e3397ad44ad9601820896de23f7

jp-mariano:
Ticket(s):
- Closes #1843

There's a discrepancy when an endowment admin checks their balance from the Admin page and when they check it from the Profile page. The Admin page balance shows the `tokens_on_hand` balance which makes sense so that they will know if they can still withdraw funds. The Profile page balance shows the `donations_received` balance which is a cumulative metric, it does not take into account the withdrawn funds. And before the [implementation of the new fields](https://github.com/AngelProtocolFinance/devops/pull/200) in AWS, we don't have this `tokens_on_hand` fields in our DB records.

To address this discrepancy in the balances displayed, we have to use the new "on-hand" fields returned by the fetch endpoint. These are `on_hand_liq, on_hand_lock, on_hand_overall`.

- Connect an admin wallet
- Open "your" endowment's Admin page in a new tab & on current tab, open "your" Profile page
- Make sure that the balances shown in the Admin page is the same with the Profile page

## UI changes for review
None

### DIFF
--- a/src/pages/Profile/Body/GeneralInfo/DetailsColumn/Balances.tsx
+++ b/src/pages/Profile/Body/GeneralInfo/DetailsColumn/Balances.tsx
@@ -2,13 +2,13 @@ import { humanize } from "helpers";
 import { useProfileContext } from "../../../ProfileContext";
 
 export default function Balances() {
-  const { total_liq, total_lock, overall } = useProfileContext();
+  const { on_hand_liq, on_hand_lock, on_hand_overall } = useProfileContext();
 
   return (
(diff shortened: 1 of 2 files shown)


ap-justin (approved):

SovereignAndrey (approved):
Nice work! :)

PR #1853, opened 2023-02-23
Rename `Panel -> Modal`, to align with already existing naming convention

misicnenad:
Copy/pasting (late) reviews on https://github.com/AngelProtocolFinance/angelprotocol-web-app/pull/1840

https://github.com/AngelProtocolFinance/angelprotocol-web-app/pull/1840#discussion_r1115306573 :
> nit: `Panel` doesn't need to be in a separate folder, e.g. no need for a `/components/Panel/index.tsx` structure. It can just be placed in a file `/components/Panel.tsx`.

https://github.com/AngelProtocolFinance/angelprotocol-web-app/pull/1840#discussion_r1115308711 :
> @abbasali1994 @ap-justin I would suggest we align naming that we use for modals.
> 
> We call these "modals", in our code we have "ModalContext", "useModalContext", "show/closeModal", "KadoModal", "TxModal" etc, so we should keep this naming convention when referring to related components -> this component should be renamed `"Panel" -> "Modal"`.

We already have an established naming convention and we should keep using it for consistency and ease of understanding both each other and the code.


SovereignAndrey (approved):

src/components/Modal.tsx
```diff
@@ -11,8 +11,15 @@ const Panel: FC<any> = (props) => (
...
   </Transition.Child>
 );
 
-export default Panel as typeof Dialog.Panel;
+const TitleFn: FC<any> = (props) => <Dialog.Title {...props} />;
+
+const ModalFnRoot = ModalFn as typeof Dialog.Panel;
+const Title = TitleFn as typeof Dialog.Title;
+
+const Modal = Object.assign(ModalFnRoot, { Title });
```
misicnenad, line 23: Add `Modal.Title` as a wrapper around `Dialog.Title` simply for consistency, i.e. it is less confusing to see:
```tsx
import Modal from "components/Modal";
import { Dialog } from "@headlessui/react";
...
<Modal>
  <Modal.Title>...</Modal.Title>
  ...
</Modal>
```
then seeing:
```tsx
import Modal from "components/Modal";
...
<Modal>
  <Dialog.Title>...</Dialog.Title>
  ...
</Modal>
```
Also one less import.

misicnenad: Based off of official @headlessui implementation of same components we're wrapping
For their implementation, see: https://github.com/tailwindlabs/headlessui/blob/main/packages/%40headlessui-react/src/components/dialog/dialog.tsx#L643

ap-justin (approved):

SovereignAndrey (approved):

Issue #1708, opened 2023-02-06
As an Endowment manager, I want to decide who (multi-sig or delegate) governs which parameters of my Endowment


SovereignAndrey:
**CU Ticket:** https://app.clickup.com/t/865bm380x
**Assigned To:** 
**Version:** v1.8

## COS
- [ ] styling matches Figma
- [ ] User can change the controlling entity/delegate wallet for various settings fields (if the `SettingsPermissions.modifiable == `true``)
- [ ] Hitting "Submit changes" creates a proposal to the Endowment multisig to send the correct update settings 

# Fetching the Settings Controller for an Endowment
You will need to query the Account Settings Controller contract endpoint: 
```rust
endowment_controller { id: u32 }
``` 

This will provide you with an `EndowmentController` struct for a reply, with `SettingsPermissions` struct for each settings field (ie. a row in the settings management table you're building here): 
```rust
pub struct EndowmentController {
    pub endowment_controller: SettingsPermissions,
    pub beneficiaries_allowlist: SettingsPermissions,
    pub contributors_allowlist: SettingsPermissions,
    pub maturity_allowlist: SettingsPermissions,
    pub donation_match_active: SettingsPermissions,
    pub earnings_fee: SettingsPermissions,
    pub withdraw_fee: SettingsPermissions,
    pub deposit_fee: SettingsPermissions,
    pub aum_fee: SettingsPermissions,
    pub kyc_donors_only: SettingsPermissions,
    pub name: SettingsPermissions,
    pub image: SettingsPermissions,
    pub logo: SettingsPermissions,
    pub categories: SettingsPermissions,
    pub ignore_user_splits: SettingsPermissions,
    pub split_to_liquid: SettingsPermissions,
}
```

The `SettingsPermissions` payload will have the following shape: 
```rust
pub struct SettingsPermissions {
    owner_controlled: bool,  // CW3 Multisig Owner
    gov_controlled: bool, // SubDao Gov
    modifiable: bool, // Might not be updatable (this is the "locked forever" setting when set to `true`)
    delegate: Option<Delegate>, 
}
```
The Delegate struct looks like: 
```rust
pub struct Delegate {
    address: Addr,
    expires: Option<u64>, // datetime int of delegation expiry
}
```

# Updating the Endowment Settings Controller
[...]

PR #1885, opened 2023-03-07
Create AIF Permissions page

misicnenad:
Ticket(s):
- https://github.com/AngelProtocolFinance/angelprotocol-web-app/issues/1708

## Relevant Figma designs
https://github.com/AngelProtocolFinance/angelprotocol-web-app/issues/1708#issuecomment-1457755066

**(VERIFY CHARITY UI/UX)**
- connect Charity 1 wallet
- go to http://localhost:4200/admin/1/permissions
- verify UI/UX disabled
- verify "Governance" and "Actions" columns do not appear
**(VERIFY AST/NORMAL ENDOW UI/UX)**
- connect AST 2 wallet
- go to http://localhost:4200/admin/27/permissions
- verify UI/UX matches Figma
- verify "Governance" and "Actions" columns appear
- verify settings which have "LOCKED FOREVER" set are disabled
- update permissions
- verify all requirements are met (see comment section of relevant ticket https://github.com/AngelProtocolFinance/angelprotocol-web-app/issues/1708)

## UI changes for review

**NOTE: it was decided that checkboxes should be used instead of radio buttons for appropriate fields (see https://github.com/AngelProtocolFinance/angelprotocol-web-app/issues/1708#issuecomment-1463628336 for more details).**

**NOTE 2: updated the desktop reset btn's label to read just "Reset" (instead of "Reset changes") as it makes it reusable across both desktop and mobile. Otherwise would've had to either create a mobile version of the buttons section or to use screen size hook to check which label to display. Just setting it to "Reset" on both seemed more straightforward.**

#### Light
Charity:
image
AST/normal endow.:
image
#### Dark
Charity:
image
AST/normal endow.:
image

#### Light
Charity:
image
AST/normal endow.:
image
#### Dark
Charity:
image
AST/normal endow.:
image

### DIFF
--- a/src/pages/Admin/Charity/Permissions/Form.tsx
+++ b/src/pages/Admin/Charity/Permissions/Form.tsx
@@ -10,13 +10,14 @@ export default function Form() {
   const errorName = getTypedKeys(errors).find((errKey) => !!errors[errKey]);
 
   return (
-    <form className="grid gap-6 w-full" onSubmit={submit} onReset={reset}>
+    <form className="grid gap-6 w-full" onSubmit={submit}>
       <Table className="max-lg:hidden" />
       <MobileTable className="lg:hidden" />
 
       <div className="flex flex-wrap items-center gap-3">
         <button
-          type="reset"
+          type="button"
+          onClick={reset}
           disabled={isSubmitting}
           className="btn-outline-filled py-2 px-8 rounded"
         >
--- a/src/pages/Admin/Charity/Permissions/Form.tsx
+++ b/src/pages/Admin/Charity/Permissions/Form.tsx
@@ -3,11 +3,14 @@ import Table from "./Table";
 import useSubmit from "./useSubmit";
 
 export default function Form() {
-  const { isSubmitting, reset, submit } = useSubmit();
+  const { isSubmitting, modifiable, reset, submit } = useSubmit();
 
   return (
     <form onSubmit={submit} onReset={() => reset()}>
-      <fieldset className="grid gap-6 w-full" disabled={isSubmitting}>
+      <fieldset
+        className="grid gap-6 w-full"
+        disabled={isSubmitting || !modifiable}
+      >
         <Table className="max-md:hidden" />
         <MobileTable className="md:hidden" />
 
--- a/src/pages/Admin/Charity/Permissions/Form.tsx
+++ b/src/pages/Admin/Charity/Permissions/Form.tsx
@@ -12,7 +12,7 @@ export default function Form() {
   return (
     <form className="grid gap-6 w-full" onSubmit={submit} onReset={reset}>
       <Table className="max-lg:hidden" />
-      <MobileTable className="lg:hidden" disabled={isSubmitting} />
+      <MobileTable className="lg:hidden" />
 
       <div className="flex flex-wrap items-center gap-3">
         <button
--- a/src/pages/Admin/Charity/Permissions/Form.tsx
+++ b/src/pages/Admin/Charity/Permissions/Form.tsx
@@ -10,11 +10,7 @@ export default function Form() {
   const errorName = getTypedKeys(errors).find((errKey) => !!errors[errKey]);
 
   return (
-    <form
-      className="grid gap-6 w-full"
-      onSubmit={submit}
-      onReset={() => reset()}
-    >
+    <form className="grid gap-6 w-full" onSubmit={submit} onReset={reset}>
       <Table className="max-lg:hidden" disabled={disabled} />
       <MobileTable className="lg:hidden" disabled={disabled} />
 
@@ -38,7 +34,7 @@ export default function Form() {
             errors={errors}
             name={`${errorName}.delegate_address`}
             as="span"
-            className="max-lg:hidden field-error static text-sm"
+            className="max-lg:hidden field-error static text-sm flex-wrap"
           />
         )}
       </div>
--- a/src/pages/Admin/Charity/Permissions/Form.tsx
+++ b/src/pages/Admin/Charity/Permissions/Form.tsx
@@ -14,7 +14,7 @@ export default function Form() {
       <Table className="max-lg:hidden" disabled={disabled} />
       <MobileTable className="lg:hidden" disabled={disabled} />
 
-      <div className="flex items-center gap-3">
+      <div className="flex flex-wrap items-center gap-3">
         <button
           type="reset"
           disabled={disabled}
@@ -34,7 +34,7 @@ export default function Form() {
             errors={errors}
             name={`${errorName}.delegate_address`}
             as="span"
-            className="max-lg:hidden field-error static text-sm flex-wrap"
+            className="field-error static text-sm"
           />
         )}
       </div>
--- a/src/pages/Admin/Charity/Permissions/Form.tsx
+++ b/src/pages/Admin/Charity/Permissions/Form.tsx
@@ -2,10 +2,8 @@ import MobileTable from "./MobileTable";
 import Table from "./Table";
 import useSubmit from "./useSubmit";
 
-export default function Form({ disabled = false }) {
-  const { isSubmitting, reset, submit } = useSubmit();
-
-  const isDisabled = disabled || isSubmitting;
+export default function Form() {
+  const { disabled, reset, submit } = useSubmit();
 
   return (
     <form
@@ -13,20 +11,20 @@ export default function Form({ disabled = false }) {
       onSubmit={submit}
       onReset={() => reset()}
     >
-      <Table className="max-md:hidden" disabled={isDisabled} />
-      <MobileTable className="md:hidden" disabled={isDisabled} />
+      <Table className="max-md:hidden" disabled={disabled} />
+      <MobileTable className="md:hidden" disabled={disabled} />
 
       <div className="flex gap-3">
         <button
           type="reset"
-          disabled={isDisabled}
+          disabled={disabled}
           className="btn-outline-filled py-2 px-8 rounded"
         >
           Reset
         </button>
         <button
           type="submit"
-          disabled={isDisabled}
+          disabled={disabled}
           className="btn-orange py-2 px-8 rounded"
         >
           Submit changes
--- a/src/pages/Admin/Charity/Permissions/Form.tsx
+++ b/src/pages/Admin/Charity/Permissions/Form.tsx
@@ -6,30 +6,20 @@ export default function Form() {
   const { isSubmitting, reset, submit } = useSubmit();
 
   return (
-    <form
-      className="grid gap-6 w-full"
-      onSubmit={submit}
-      onReset={() => reset()}
-    >
-      <Table className="max-md:hidden" />
-      <MobileTable className="md:hidden" />
+    <form onSubmit={submit} onReset={() => reset()}>
+      <fieldset className="grid gap-6 w-full" disabled={isSubmitting}>
+        <Table className="max-md:hidden" />
+        <MobileTable className="md:hidden" />
 
-      <div className="flex gap-3">
-        <button
-          type="reset"
-          className="btn-outline-filled py-2 px-8 rounded"
-          disabled={isSubmitting}
-        >
-          Reset
-        </button>
-        <button
-          type="submit"
-          className="btn-orange py-2 px-8 rounded"
-          disabled={isSubmitting}
-        >
-          Submit changes
-        </button>
-      </div>
+        <div className="flex gap-3">
+          <button type="reset" className="btn-outline-filled py-2 px-8 rounded">
+            Reset
+          </button>
+          <button type="submit" className="btn-orange py-2 px-8 rounded">
+            Submit changes
+          </button>
+        </div>
+      </fieldset>
     </form>
   );
 }
--- a/src/pages/Admin/Charity/Permissions/Form.tsx
+++ b/src/pages/Admin/Charity/Permissions/Form.tsx
@@ -5,26 +5,26 @@ import Table from "./Table";
 import useSubmit from "./useSubmit";
 
 export default function Form() {
-  const { disabled, errors, reset, submit } = useSubmit();
+  const { isSubmitting, errors, reset, submit } = useSubmit();
 
   const errorName = getTypedKeys(errors).find((errKey) => !!errors[errKey]);
 
   return (
     <form className="grid gap-6 w-full" onSubmit={submit} onReset={reset}>
-      <Table className="max-lg:hidden" disabled={disabled} />
-      <MobileTable className="lg:hidden" disabled={disabled} />
+      <Table className="max-lg:hidden" disabled={isSubmitting} />
+      <MobileTable className="lg:hidden" disabled={isSubmitting} />
 
       <div className="flex flex-wrap items-center gap-3">
         <button
           type="reset"
-          disabled={disabled}
+          disabled={isSubmitting}
           className="btn-outline-filled py-2 px-8 rounded"
         >
           Reset
         </button>
         <button
           type="submit"
-          disabled={disabled}
+          disabled={isSubmitting}
           className="btn-orange py-2 px-8 rounded"
         >
           Submit changes
--- a/src/pages/Admin/Charity/Permissions/Form.tsx
+++ b/src/pages/Admin/Charity/Permissions/Form.tsx
@@ -2,8 +2,10 @@ import MobileTable from "./MobileTable";
 import Table from "./Table";
 import useSubmit from "./useSubmit";
 
-export default function Form() {
-  const { disabled, reset, submit } = useSubmit();
+export default function Form({ disabled = false }) {
+  const { isSubmitting, reset, submit } = useSubmit();
+
+  const isDisabled = disabled || isSubmitting;
 
   return (
     <form
@@ -11,20 +13,20 @@ export default function Form() {
       onSubmit={submit}
       onReset={() => reset()}
     >
-      <Table className="max-md:hidden" disabled={disabled} />
-      <MobileTable className="md:hidden" disabled={disabled} />
+      <Table className="max-md:hidden" disabled={isDisabled} />
+      <MobileTable className="md:hidden" disabled={isDisabled} />
 
       <div className="flex gap-3">
         <button
           type="reset"
-          disabled={disabled}
+          disabled={isDisabled}
           className="btn-outline-filled py-2 px-8 rounded"
         >
           Reset
         </button>
         <button
           type="submit"
-          disabled={disabled}
+          disabled={isDisabled}
           className="btn-orange py-2 px-8 rounded"
         >
           Submit changes
--- a/src/pages/Admin/Charity/Permissions/Form.tsx
+++ b/src/pages/Admin/Charity/Permissions/Form.tsx
@@ -15,8 +15,8 @@ export default function Form() {
       onSubmit={submit}
       onReset={() => reset()}
     >
-      <Table className="max-md:hidden" disabled={disabled} />
-      <MobileTable className="md:hidden" disabled={disabled} />
+      <Table className="max-lg:hidden" disabled={disabled} />
+      <MobileTable className="lg:hidden" disabled={disabled} />
 
       <div className="flex items-center gap-3">
         <button
@@ -38,7 +38,7 @@ export default function Form() {
             errors={errors}
             name={`${errorName}.delegate_address`}
             as="span"
-            className="field-error static text-sm"
+            className="max-lg:hidden field-error static text-sm"
           />
         )}
       </div>
--- a/src/pages/Admin/Charity/Permissions/Form.tsx
+++ b/src/pages/Admin/Charity/Permissions/Form.tsx
@@ -3,26 +3,33 @@ import Table from "./Table";
 import useSubmit from "./useSubmit";
 
 export default function Form() {
-  const { isSubmitting, modifiable, reset, submit } = useSubmit();
+  const { disabled, reset, submit } = useSubmit();
 
   return (
-    <form onSubmit={submit} onReset={() => reset()}>
-      <fieldset
-        className="grid gap-6 w-full"
-        disabled={isSubmitting || !modifiable}
-      >
-        <Table className="max-md:hidden" />
-        <MobileTable className="md:hidden" />
+    <form
+      className="grid gap-6 w-full"
+      onSubmit={submit}
+      onReset={() => reset()}
+    >
+      <Table className="max-md:hidden" disabled={disabled} />
+      <MobileTable className="md:hidden" disabled={disabled} />
 
-        <div className="flex gap-3">
-          <button type="reset" className="btn-outline-filled py-2 px-8 rounded">
-            Reset
-          </button>
-          <button type="submit" className="btn-orange py-2 px-8 rounded">
-            Submit changes
-          </button>
-        </div>
-      </fieldset>
+      <div className="flex gap-3">
+        <button
+          type="reset"
+          disabled={disabled}
+          className="btn-outline-filled py-2 px-8 rounded"
+        >
+          Reset
+        </button>
+        <button
+          type="submit"
+          disabled={disabled}
+          className="btn-orange py-2 px-8 rounded"
+        >
+          Submit changes
+        </button>
+      </div>
     </form>
   );
 }
--- a/src/pages/Admin/Charity/Permissions/Form.tsx
+++ b/src/pages/Admin/Charity/Permissions/Form.tsx
@@ -1,9 +1,13 @@
+import { ErrorMessage } from "@hookform/error-message";
+import { getTypedKeys } from "helpers";
 import MobileTable from "./MobileTable";
 import Table from "./Table";
 import useSubmit from "./useSubmit";
 
 export default function Form() {
-  const { disabled, reset, submit } = useSubmit();
+  const { disabled, errors, reset, submit } = useSubmit();
+
+  const errorName = getTypedKeys(errors).find((errKey) => !!errors[errKey]);
 
   return (
     <form
@@ -14,7 +18,7 @@ export default function Form() {
       <Table className="max-md:hidden" disabled={disabled} />
       <MobileTable className="md:hidden" disabled={disabled} />
 
-      <div className="flex gap-3">
+      <div className="flex items-center gap-3">
         <button
           type="reset"
           disabled={disabled}
@@ -29,6 +33,14 @@ export default function Form() {
         >
           Submit changes
         </button>
+        {errorName && (
+          <ErrorMessage
+            errors={errors}
+            name={`${errorName}.delegate_address`}
+            as="span"
+            className="field-error static text-sm"
+          />
+        )}
       </div>
     </form>
   );
--- a/src/pages/Admin/Charity/Permissions/Form.tsx
+++ b/src/pages/Admin/Charity/Permissions/Form.tsx
@@ -11,7 +11,7 @@ export default function Form() {
 
   return (
     <form className="grid gap-6 w-full" onSubmit={submit} onReset={reset}>
-      <Table className="max-lg:hidden" disabled={isSubmitting} />
+      <Table className="max-lg:hidden" />
       <MobileTable className="lg:hidden" disabled={isSubmitting} />
 
       <div className="flex flex-wrap items-center gap-3">
--- a/src/pages/Admin/Charity/Permissions/MobileTable.tsx
+++ b/src/pages/Admin/Charity/Permissions/MobileTable.tsx
@@ -4,6 +4,7 @@ import { useAdminResources } from "pages/Admin/Guard";
 import { useGetWallet } from "contexts/WalletContext";
 import { DrawerIcon } from "components/Icon";
 import { CheckField, Field } from "components/form";
+import { getTypedKeys } from "helpers";
 import { FormField, FormValues } from "./schema";
 
 const formValues: FormValues["initialValues"] = {
@@ -13,9 +14,6 @@ const formValues: FormValues["initialValues"] = {
   donationSplitParams: {} as FormField,
   profile: {} as FormField,
 };
-function getTypedKeys<T extends object>(obj: T): Array<keyof T> {
-  return Object.keys(obj) as Array<keyof T>;
-}
 const FORM_KEYS = getTypedKeys(formValues);
 
 export default function MobileTable({ className = "", disabled = false }) {
--- a/src/pages/Admin/Charity/Permissions/MobileTable.tsx
+++ b/src/pages/Admin/Charity/Permissions/MobileTable.tsx
@@ -4,14 +4,24 @@ import { useAdminResources } from "pages/Admin/Guard";
 import { useGetWallet } from "contexts/WalletContext";
 import { DrawerIcon } from "components/Icon";
 import { CheckField, Field } from "components/form";
-import { FormValues } from "./schema";
+import { FormField, FormValues } from "./schema";
+
+const formValues: FormValues["initialValues"] = {
+  accountFees: {} as FormField,
+  beneficiaries_allowlist: {} as FormField,
+  contributors_allowlist: {} as FormField,
+  donationSplitParams: {} as FormField,
+  profile: {} as FormField,
+};
+function getTypedKeys<T extends object>(obj: T): Array<keyof T> {
+  return Object.keys(obj) as Array<keyof T>;
+}
+const FORM_KEYS = getTypedKeys(formValues);
 
 export default function MobileTable({ className = "", disabled = false }) {
   const { endow_type } = useAdminResources<"charity">();
   const { wallet } = useGetWallet();
-  const { getValues, setValue, watch } = useFormContext<FormValues>();
-
-  const { initialValues, ...formValues } = getValues();
+  const { setValue, watch } = useFormContext<FormValues>();
 
   return (
     <div className={`${className} border border-prim rounded-t`}>
@@ -20,9 +30,7 @@ export default function MobileTable({ className = "", disabled = false }) {
         <div className="py-3 px-4">Action</div>
       </div>
 
-      {(
-        Object.keys(formValues) as (keyof Omit<FormValues, "initialValues">)[]
-      ).map((formField) => (
+      {FORM_KEYS.map((formField) => (
         <Disclosure
           key={`mob-table-row-${formField}`}
           as="div"
--- a/src/pages/Admin/Charity/Permissions/MobileTable.tsx
+++ b/src/pages/Admin/Charity/Permissions/MobileTable.tsx
@@ -1,6 +1,5 @@
 import { Disclosure } from "@headlessui/react";
 import { useFormContext } from "react-hook-form";
-import { useAdminResources } from "pages/Admin/Guard";
 import { useGetWallet } from "contexts/WalletContext";
 import { DrawerIcon } from "components/Icon";
 import { CheckField, Field } from "components/form";
@@ -18,7 +17,6 @@ const formValues: UpdateableFormValues = {
 const FORM_KEYS = getTypedKeys(formValues);
 
 export default function MobileTable({ className = "", disabled = false }) {
-  const { endow_type } = useAdminResources<"charity">();
   const { wallet } = useGetWallet();
   const { watch } = useFormContext<FormValues>();
 
@@ -114,13 +112,11 @@ export default function MobileTable({ className = "", disabled = false }) {
                     }}
                     disabled={!delegate || isDisabled}
                   />
-                  {endow_type === "normal" ? (
-                    /** Color #54595F is hardcoded because this is the only place where it's necessary */
-                    <div className="flex justify-between items-center p-4">
-                      <span className="font-bold uppercase">Actions</span>
-                      <LockButton disabled={isDisabled} name={fieldName} />
-                    </div>
-                  ) : null}
+                  {/** Color #54595F is hardcoded because this is the only place where it's necessary */}
+                  <div className="flex justify-between items-center p-4">
+                    <span className="font-bold uppercase">Actions</span>
+                    <LockButton disabled={isDisabled} name={fieldName} />
+                  </div>
                 </Disclosure.Panel>
               </>
             );
--- a/src/pages/Admin/Charity/Permissions/MobileTable.tsx
+++ b/src/pages/Admin/Charity/Permissions/MobileTable.tsx
@@ -35,18 +35,23 @@ export default function MobileTable({ className = "", disabled = false }) {
           className="text-sm odd:bg-orange-l6 dark:even:bg-blue-d6 dark:odd:bg-blue-d7 w-full border-b last:border-0 border-prim"
         >
           {({ open }) => {
-            const modifiable = watch(`${formField}.modifiable`);
             const delegate = watch(`${formField}.delegate`);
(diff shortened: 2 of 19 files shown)


src/components/TableSection.tsx
```diff
@@ -22,6 +22,10 @@ export default function TableSection(props: HeadProps | BodyProps) {
 
 export function Cells(props: CellProps) {
   const cells = React.Children.map(props.children, (child, index) => {
+    if (!child) {
+      return null;
```
misicnenad, line 26: Automatically ignored by the table component.

src/pages/Admin/Charity/Permissions/useSubmit.ts
```diff
@@ -0,0 +1,63 @@
...
+        msgs: [proposalMsg],
+        ...propMeta,
+      });
+    } catch (error) {
+      handleError(error);
+    }
+  }
+
+  return {
+    disabled:
```
ap-justin, line 58: @misicnenad, @SovereignAndrey, @abbasali1994 

this might be a discussion on general handling of `disabled` state of a form

image

as a developer, when I go to this page and see everything disabled - I know what to do. 

* I need to connect my wallet
* I have to be on the right endowment  / must have the right authorization

but for someone who doesn't know that internals of this. It might be confusing to see why everything is `disabled` and user might try bunch of interactions just to make things `enabled`

Solution to this:
1. If we want everything disabled at first, we should provide helpful text/tooltips why the form is disabled like

or 
2. Just enable the form initially, and when user tries to submit be prompted with messages like
`Your wallet needs to be connected to submit this transaction`
`Your wallet is not authorized to modify settings` 
etc..

this is also considering admin list in wallet dropdown is not perfectly in 
image
sync so a `user` might be on a page where he is not a member of

src/pages/Admin/Charity/Permissions/Table.tsx
```diff
@@ -0,0 +1,141 @@
...
+                    }
+                    disabled={!value.delegate || isDisabled}
+                  />
+
+                  {isNormal ? (
+                    <button
+                      type="button"
+                      className="btn-red py-1 px-2 rounded font-semibold text-xs uppercase text-white tracking-wider"
+                      disabled={isDisabled}
+                      onClick={() => {
```
ap-justin, line 122: image

`lock-forever` should trigger validation, we could get this scenario that an invalid value is locked

ap-justin (changes requested):

src/pages/Admin/Charity/Permissions/useSubmit.ts
```diff
@@ -0,0 +1,63 @@
...
+      isSubmitting || endow_type === "charity" || !propMeta.isAuthorized,
```
ap-justin, line 59: should also check if `EndowmentController.endowment_controller` is delegated to `sender`. In this case, even if `sender` is not the `owner` (single wallet or a member of multisig wallet (CW3)), he/she could still submit this `{ update_endowment_controller }` msg

ap-justin (changes requested):

src/pages/Admin/Charity/Permissions/MobileTable.tsx
```diff
@@ -0,0 +1,127 @@
...
+                  {endow_type === "normal" ? (
+                    <div className="flex justify-between items-center p-4">
+                      <span className="font-bold uppercase">Actions</span>
+                      <button
+                        type="button"
+                        className="btn-red p-2 rounded font-semibold text-xs uppercase text-white tracking-wider"
+                        {...register(`${formField}.modifiable`)}
+                        disabled={isDisabled}
+                      >
+                        {modifiable ? "Lock forever" : "Locked forever"}
```
ap-justin, line 115: image

`Lock  forever` button doesn't do anything on mobile table

ap-justin (changes requested):

src/pages/Admin/Charity/Permissions/MobileTable.tsx
```diff
@@ -0,0 +1,127 @@
...
+            const initDelegate = watch(`initialValues.${formField}.delegate`);
+            const init_delegate_address = watch(
+              `initialValues.${formField}.delegate_address`
+            );
+            const isDisabled =
+              !modifiable ||
+              disabled ||
+              (initDelegate &&
+                !!init_delegate_address &&
+                init_delegate_address !== wallet?.address);
```
ap-justin, line 43: whether connected wallet has control been deligated to shouldn't affect how he would interact to individual settings, nor `modifiable` in the `whole` endowment controller context

NOTE: this is with the execption for `EndowmentController.endowment_controller` setting, for which if delegated to a single entity  `Z` would mean `Mr.Z` is the only one who could modify all other individual settings

consider this scenario
1. initialy, `owner` ( whether single wallet or multisig wallet say `AdminX` ) is allowed to modify `EndowmentController`  as specified in `EndowmentController.endowment_controller` field which is defaulted to 
```rust
pub fn default(endow_type: &EndowmentType) -> Self {
        SettingsPermissions {
            owner_controlled: true,
            gov_controlled: false,
            modifiable: match endow_type {
                EndowmentType::Charity => false,
                _ => true,
            },
            delegate: None,
        }
    }
```

2. let's say `A`, `B`, `C` are members of `AdminX`. and `A` creates a proposal to delegate all inividual settings to `A` and set all fields to `not modifiable`

```ts
//PERMISSIONS 
1.`fees` --> delegate --> `A`
2.`beneficiaries` --> delegate --> `A`
3. `contributors` --> delegate --> `A`
```
and everyone approves, making the changes effective

3.  this would mean that `B, C` could not anymore modify permissions `1`, `2`, `3`, though technically they should be able to since they are still a member of the `owner` specified in `Endomwent.endowment_controller`

ap-justin (changes requested):

src/pages/Admin/Charity/Permissions/useSubmit.ts
```diff
@@ -0,0 +1,63 @@
...
+
+      const settingsController = new SettingsController(wallet);
+      const msg = createUpdateEndowmentControllerMsg(id, diff);
+      const embeddedMsg =
+        settingsController.createEmbeddedUpdateEndowmentControllerMsg(msg);
+
+      const meta: ProposalMeta = { type: "endow_controller" };
+
+      const adminContract = new CW3(wallet, cw3);
+      const proposalMsg = adminContract.createProposalMsg(
```
ap-justin, line 41: should have a provision where if `wallet.address === EndowmentController.endowment_controller.delegate || wallet.address === endowent.owner`, should just send the msg to `SettingsController` contract

ap-justin (changes requested):

src/types/contracts/settingsController.ts
```diff
@@ -0,0 +1,34 @@
...
+  beneficiaries_allowlist: SettingsPermissions;
+  contributors_allowlist: SettingsPermissions;
+  maturity_allowlist: SettingsPermissions;
+  earnings_fee: SettingsPermissions;
+  deposit_fee: SettingsPermissions;
+  withdraw_fee: SettingsPermissions;
+  aum_fee: SettingsPermissions;
+};
+
+export type UpdateEndowmentControllerMsg = Partial<EndowmentController> & {
```
ap-justin, line 32: though `type EndowmentController` and `type UpdateEndowmentControllerMsg` looks the same, they are actually separate structs in the contracts. so it's better to define them separately here.  Moreover, they do have some differences on them 
e.g `type EndowmentController` have `profile` field while `UpdateEndowmentControllerMsg` have `donation_match_active` field

ap-justin (changes requested):

src/pages/Admin/Charity/Permissions/MobileTable.tsx
```diff
@@ -0,0 +1,127 @@
...
+import { DrawerIcon } from "components/Icon";
+import { CheckField, Field } from "components/form";
+import { FormValues } from "./schema";
+
+export default function MobileTable({ className = "", disabled = false }) {
+  const { endow_type } = useAdminResources<"charity">();
+  const { wallet } = useGetWallet();
+  const { getValues, register, watch } = useFormContext<FormValues>();
+
+  const { initialValues, ...formValues } = getValues();
```
ap-justin, line 14: nit:  though `getValues` doesn't trigger rerender, it's quite uintuitive that `formFields` are rendered by `formValues` which it also updates
`formValues` -> `formFields` -> `formValues`

might be better to just create static `formFields` directly from FormValues
```ts
const formFields:{K in keyof FormValues: K } = {/** ... */}
const fieldNames = Object.keys(formFields)
```

src/pages/Admin/Charity/Permissions/Table.tsx
```diff
@@ -0,0 +1,141 @@
...
+      </TableSection>
+      <TableSection
+        type="tbody"
+        rowClass="dark:bg-blue-d6 divide-x divide-prim"
+        selectedClass="bg-orange-l5 dark:bg-blue-d4"
+      >
+        {(
+          Object.keys(formValues) as (keyof Omit<FormValues, "initialValues">)[]
+        ).map((formField) => (
+          <Controller
```
ap-justin, line 45: nit: instead of `<Controller/>` (controlled), and manually setting `onChange` on each `<input/>` could abstract a row
```tsx
//Row.tsx
const { register } = useFormContext()
return <>
   ....
   <input {...register("name.delegate")}/> 
   <input {...register("name.address")}/> 
  </>
```

src/types/contracts/settingsController.ts
SovereignAndrey: @ap-justin @misicnenad This is changing as we speak! I've been working on harmonizing that and some other bug fixes this morning. Stand by for an update later this afternoon. :+1:

SovereignAndrey: **UPDATE:**
`profile` is removed now from the controller and update msg struct & `donation_match_active` has been added to both. The "profile related" fields (ie. `kyc_donors_only, name, image, logo, categories`) have had the default values changed for Charity type endowments so that they ARE modifiable should they choose to do so.

src/pages/Admin/Charity/Permissions/useSubmit.ts
misicnenad: @SovereignAndrey @ap-justin might be useful for disconnected/unauthorized users to have a top banner on each page with a tooltip stating `"Your wallet needs to be connected and authorized to modify settings"` (a combination of the two messages from @ap-justin's comment, but we could have separate messages ofc).
It could be similar to the one in the Withdraw page:
image

src/pages/Admin/Charity/Permissions/Table.tsx
misicnenad: @SovereignAndrey @ap-justin 
Maybe we can make it so that clicking "Lock forever" just sets the appropriate form value, but does not **lock** the table row until the form is submitted and the changes are executed.
In other words, make it so that a table row is _actually locked_ only if the initial `modifiable` field is set to `false`. This would allow for two things:
- no need to manually trigger validation
- **even more important** - would allow the user to go back and forth between "Lock forever" and "LockED forever", which they would be able to do until they actually submit the form (current UX is bad as once the settings are set as "locked" but NOT SUBMITTED, they cannot be switched back until the whole page is reloaded - what if the user locked the settings by mistake or something dumb as just wanting to see what would happen). In this case, once the form is submitted and executed, the rows are locked *for real*.

src/pages/Admin/Charity/Permissions/useSubmit.ts
misicnenad: @SovereignAndrey can you confirm this is also one of the conditions to be able to send the msg?
Can you also confirm that none of the current settings can update `EndowmentController.endowment_controller`?

src/pages/Admin/Charity/Permissions/MobileTable.tsx
misicnenad: @ap-justin this is related to the comment above https://github.com/AngelProtocolFinance/angelprotocol-web-app/pull/1885#discussion_r1136474481 ?

src/pages/Admin/Charity/Permissions/useSubmit.ts
SovereignAndrey: > @SovereignAndrey @ap-justin might be useful for disconnected/unauthorized users to have a top banner on each page with a tooltip stating `"Your wallet needs to be connected and authorized to modify settings"` (a combination of the two messages from @ap-justin's comment, but we could have separate messages ofc). It could be similar to the one in the Withdraw page: image

That would be good. It's also a bigger issue than just on this page alone for how we ought to display forms to non-connected wallet users. Ideally, we have separate, non-form based views for when wallets are not connected.

misicnenad:
Converting to draft PR as the intended requirements might not have been met https://github.com/AngelProtocolFinance/angelprotocol-web-app/issues/1708#issuecomment-1469519395

src/pages/Admin/Charity/Permissions/useSubmit.ts
misicnenad: Waiting on Andrey to confirm the exact requirements here, see https://github.com/AngelProtocolFinance/angelprotocol-web-app/pull/1885#issuecomment-1469530382.

src/pages/Admin/Charity/Permissions/MobileTable.tsx
misicnenad: Waiting on Andrey to confirm the exact requirements here, see https://github.com/AngelProtocolFinance/angelprotocol-web-app/pull/1885#issuecomment-1469530382.

src/types/contracts/settingsController.ts
misicnenad: @ap-justin 
> though type EndowmentController and type UpdateEndowmentControllerMsg may look the same, they are actually separate structs in the contracts

In light of recent changes to types, we can keep these types as-is until they start drastically differing from each other.

src/pages/Admin/Charity/Permissions/MobileTable.tsx
misicnenad: Sure.

src/types/contracts/settingsController.ts
misicnenad: > The "profile related" fields (ie. kyc_donors_only, name, image, logo, categories) have had the default values changed for Charity type endowments so that they ARE modifiable should they choose to do so

@SovereignAndrey does that mean we **should** have the "Actions" (lock(ed) forever btn) column displayed for charities as well (previous requirement was to hide it for them https://github.com/AngelProtocolFinance/angelprotocol-web-app/issues/1708#issuecomment-1463675056)?

src/pages/Admin/Charity/Permissions/useSubmit.ts
misicnenad: @SovereignAndrey @ap-justin since this is a problem related to _all_ admin pages, it's safe to say that this requirement is out of scope of this PR and should be tackled separately.

src/pages/Admin/Charity/Permissions/useSubmit.ts
misicnenad: As per the most recent comments https://github.com/AngelProtocolFinance/angelprotocol-web-app/issues/1708#issuecomment-1469557854, `wallet.address === EndowmentController.endowment_controller.delegate` is the only relevant check, updating the code.

src/pages/Admin/Charity/Permissions/MobileTable.tsx
misicnenad: > whether connected wallet has control been deligated to shouldn't affect how he would interact to individual settings, nor `modifiable` in the `whole` endowment controller context
> 
> NOTE: this is with the execption for `EndowmentController.endowment_controller` setting, for which if delegated to a single entity `Z` would mean `Mr.Z` is the only one who could modify all other individual settings
> 
> consider this scenario
> 
> 1. initialy, `owner` ( whether single wallet or multisig wallet say `AdminX` ) is allowed to modify `EndowmentController`  as specified in `EndowmentController.endowment_controller` field which is defaulted to
> 
> ```rust
> pub fn default(endow_type: &EndowmentType) -> Self {
>         SettingsPermissions {
>             owner_controlled: true,
>             gov_controlled: false,
>             modifiable: match endow_type {
>                 EndowmentType::Charity => false,
>                 _ => true,
>             },
>             delegate: None,
>         }
>     }
> ```
> 
> 2. let's say `A`, `B`, `C` are members of `AdminX`. and `A` creates a proposal to delegate all inividual settings to `A` and set all fields to `not modifiable`
> 
> ```ts
> //PERMISSIONS 
> 1.`fees` --> delegate --> `A`
> 2.`beneficiaries` --> delegate --> `A`
> 3. `contributors` --> delegate --> `A`
> ```
> 
> and everyone approves, making the changes effective
> 
> 3. this would mean that `B, C` could not anymore modify permissions `1`, `2`, `3`, though technically they should be able to since they are still a member of the `owner` specified in `Endomwent.endowment_controller`

@SovereignAndrey can you comment on this flow, seems like an important (potential) edge case?

src/types/contracts/settingsController.ts
SovereignAndrey: > > The "profile related" fields (ie. kyc_donors_only, name, image, logo, categories) have had the default values changed for Charity type endowments so that they ARE modifiable should they choose to do so
> 
> @SovereignAndrey does that mean we **should** have the "Actions" (lock(ed) forever btn) column displayed for charities as well (previous requirement was to hide it for them [#1708 (comment)](https://github.com/AngelProtocolFinance/angelprotocol-web-app/issues/1708#issuecomment-1463675056))?

Yes, we should show it after all, since not everything it locked forever.

src/pages/Admin/Charity/Permissions/MobileTable.tsx
SovereignAndrey: > @SovereignAndrey can you comment on this flow, seems like an important (potential) edge case?

@misicnenad @ap-justin This is not an issue, as in the above scenario **no one** would be able to modify the `EndowmentSettings` values for the `fees, beneficiaries, or contributors` once the proposal passed, since that proposal has made those fields **not** modifiable. The modifiable check if done first before ownership & message sender ID checks, so it applies regardless of whomever might have/had the ability to change a given field/attribute. Making `A` a delegate in the same proposal was a non-event since they were immediately locked out at the same time.

src/pages/Admin/Charity/Permissions/Table.tsx
SovereignAndrey: > @SovereignAndrey @ap-justin Maybe we can make it so that clicking "Lock forever" just sets the appropriate form value, but does not **lock** the table row until the form is submitted and the changes are executed. In other words, make it so that a table row is _actually locked_ only if the initial `modifiable` field is set to `false`. This would allow for two things:
> 
> * the user would be able to keep updating other fields in the same row -> just because they want the fields locked in the future, does not mean they want them locked _immediately_ - they might still want to change other fields in the same row and THEN lock them (once submitted)
> * would allow the user to go back and forth between "Lock forever" and "LockED forever", which they would be able to do until they actually submit the form (current UX is bad as once the settings are set as "locked" but NOT SUBMITTED, they cannot be switched back until the whole page is reloaded - what if the user locked the settings by mistake or something dumb as just wanting to see what would happen). In this case, once the form is submitted and executed, the rows are locked _for real_.

Agreed, that we should allow them to configure the settings (owner and locked status (assuming still modifiable)) to however they wish before submitting and send everything at once. That said locking a previously unlocked field should "grey out" the row or something to signal it's all set in stone at that point, but they should be able to toggle that back to "unlocked" and change the ownership checks / delegate fields if desired. 

We'll want to make sure that if a field is originally locked on contract then it can NOT be toggled back to unlocked. Allowing that would mean a submitted message / proposal would fail since it'd be trying to update a non-modifiable field in the EndowmentController. 

* I will need to update the contract's `set_permissions` function to check the existing `modifiable` state there _before_ setting new values. We currently only look at modifiable when it comes to changing the field's setting and not the controller as well.

misicnenad:
@SovereignAndrey @ap-justin do you have an idea how we could display error messages for invalid wallet address in "Delegate Address" field(s)?
Trying things out and need some suggestions.

An idea for Desktop view would be to display an error message to the right of the "Submit changes" button:
image
What do you think about this?
Do you have an idea how we could display this on mobile too?

src/pages/Admin/Charity/Permissions/Form.tsx
```diff
@@ -0,0 +1,47 @@
...
+
+  const errorName = getTypedKeys(errors).find((errKey) => !!errors[errKey]);
+
+  return (
+    <form
+      className="grid gap-6 w-full"
+      onSubmit={submit}
+      onReset={() => reset()}
+    >
+      <Table className="max-lg:hidden" disabled={disabled} />
```
misicnenad, line 18: Used webapp width for breakpoints instead of tailwind container queries as there seem to be some bugs in it, see example below.

Notice a couple of things:
- using named @container like `@container/perm`
- `Table` should be displayed only when `@container/perm` min width is `md:`
- `MobileTable` should be hidden only when `@container/perm` min width is `md:`
- component with `@container/perm` class name is clearly less than `md:` in width
- `Table` is still appearing
image

src/pages/Admin/Charity/Permissions/MobileTable.tsx
```diff
@@ -0,0 +1,143 @@
...
+                    } w-12 place-self-center`}
+                    isOpen={open}
+                  />
+                  <p className="text-sm p-4 text-left h-full truncate">
+                    {name}
+                  </p>
+                </Disclosure.Button>
+                <Disclosure.Panel className="w-full font-work divide-y divide-prim border-t border-prim">
+                  <div className="grid gap-4 py-3 px-4">
+                    <CheckField<FormValues>
```
misicnenad, line 82: There's a bug with checkboxes when switching between desktop and mobile, it will be addressed in https://github.com/orgs/AngelProtocolFinance/projects/2/views/1?pane=issue&itemId=23101212

SovereignAndrey:
> @SovereignAndrey @ap-justin do you have an idea how we could display error messages for invalid wallet address in "Delegate Address" field(s)?
> Trying things out and need some suggestions.
> 
> An idea for Desktop view would be to display an error message to the right of the "Submit changes" button and give the invalid input a red border:
> image
> What do you think about this?
> Do you have an idea how we could display this on mobile too?

Couldn't mobile use the same setup? Show the error below the submit button and open the drawers for rows with the errors. 

Good enough for now I think. Some internal reviews by the broader team and comments will help drive the final design and any changes needed(if any). :+1:

ap-justin (changes requested):

src/pages/Admin/Charity/Permissions/Table.tsx
```diff
@@ -0,0 +1,129 @@
...
+          );
+          const initModifiable = watch(`initialValues.${fieldName}.modifiable`);
+          const modifiable = watch(`${fieldName}.modifiable`);
+
+          const isDisabled =
+            disabled ||
+            !initModifiable ||
+            (initDelegate &&
+              !!initDelegateAddress &&
+              initDelegateAddress !== wallet?.address);
```
ap-justin, line 71: @misicnenad,
[previous conversation](https://github.com/AngelProtocolFinance/angelprotocol-web-app/pull/1885#discussion_r1136490308) has been resolved on the basis of this new information ( regarding - `modifiable` field) 


>  We'll want to make sure that if a field is originally locked on contract then it can NOT be toggled back to unlocked. Allowing that would mean a submitted message / proposal would fail since it'd be trying to update a non-modifiable field in the EndowmentController.


and is partly covered in this check
```ts
const isDisabled =
            disabled ||
            !initModifiable ||
            (initDelegate &&
              !!initDelegateAddress &&
              initDelegateAddress !== wallet?.address);
```

but disabling the field based on who has current control delegated is still a mistake 

going back to this example (but without the `modifiable` part)

> let's say A, B, C are members of AdminX. and A creates a proposal to delegate all inividual settings to A ~~and set all fields to not modifiable~~

 ```
 //PERMISSIONS 
 1.`fees` --> delegate --> `A`
 2.`beneficiaries` --> delegate --> `A`
 3. `contributors` --> delegate --> `A`
```

this would mean that `B`,`C` can't propose to change the delegated address anymore even though it's `not locked`


NOTE: if `EndowmentController.endow_controller` row field is delegated to `A` then in this context (editing endowment controller as a `whole`) it makes sense to gray out this particular row - but it's not present atm in the UI 

but for other fields, checks should be done when editing those fields per [this comment](https://github.com/AngelProtocolFinance/angelprotocol-web-app/issues/1708#issuecomment-1469557854)


> This form is about updating the EndowmentController as a whole, and not the individual fields. The contract/delegate controllers of all of the other individual fields in that EndowmentController struct are only applicable when we are updating those fields themselves

**Review by ap-justin — CHANGES_REQUESTED** [review]

File: src/pages/Admin/Charity/Permissions/useSubmit.ts [hunk 14]
```diff
@@ -0,0 +1,85 @@
...
+      });
+    } catch (error) {
+      handleError(error);
+    }
+  }
+
+  return {
+    errors,
+    disabled:
+      isSubmitting || (!propMeta.isAuthorized && !watch("userDelegate")),
```
ap-justin, line 81: `propMeta.isAuthorized` is only relevant in `{update_controller_settings}` if `EndowmentController.endow_controller` is `owner_controlled` 

e.g
```ts
const disabled = ! ((owner_controlled && propMeta.isAuthorized) || delegate.address ===  wallet.address )
```

ap-justin (changes requested):

src/pages/Admin/Charity/Permissions/Permissions.tsx
```diff
@@ -0,0 +1,83 @@
...
+      "Changes to donation split parameters",
+      controller.split_to_liquid
+    ),
+    profile: createField("Changes to profile", controller.name),
+  };
+  const methods = useForm<FormValues>({
+    defaultValues: {
+      initialValues,
+      userDelegate:
+        controller.endowment_controller.delegate &&
```
ap-justin, line 57:
```suggestion
        controller.endowment_controller?.delegate?.address  === walllet?.address
```

ap-justin (changes requested):

src/pages/Admin/Charity/Permissions/Permissions.tsx
```diff
@@ -0,0 +1,83 @@
...
+    donationSplitParams: createField(
+      "Changes to donation split parameters",
+      controller.split_to_liquid
+    ),
+    profile: createField("Changes to profile", controller.name),
+  };
+  const methods = useForm<FormValues>({
+    defaultValues: {
+      initialValues,
+      userDelegate:
```
ap-justin, line 56: note that `useForm.defaultValues` will not change when `wallet` changes 

Repro: based on `useSubmit.disabled`, form should be disabled when walle is not connected - but this doesn't happen
image

ap-justin (changes requested):

src/pages/Admin/Charity/Permissions/useSubmit.ts
ap-justin: @SovereignAndrey, @misicnenad 
> since this is a problem related to all admin pages,

currently, all forms in `pages/Admin` uses the `2`nd approach 
that is: forms are submittable - then prompts user on problems

sample:
image

where problems manifest in this order
1. submitting no changes
2. wallet is not connected
3. unauthorized ( admin membership check)

now  we have 
4. permissions check for select (not all ) pages 

image


starting with the general
`{ update_controller_settings }` 

we could simply add a simple check in the submit of this page rather than disabling everything

that is , add this simple line

```
 await sendTx({
        //...
        isAuthorized: endow_controller.owner_controlled & propMeta.isAuthorized || endow_controller?.delegate?.address === wallet.address
      });
```

and on a separate PRs say
in `edit_profile`

```
 await sendTx({
        //...
        isAuthorized: name.owner_controlled & propMeta.isAuthorized || endow_controller?.delegate?.address === wallet.address
      });
```

ap-justin (changes requested):

src/pages/Admin/Charity/Permissions/Table.tsx
ap-justin, line 71: found a bug where on initial load,  `delegate` is checked but there's no initial delegate address 

image

REPRO: go to `admin/1/permissions`

ap-justin (changes requested):

src/pages/Admin/Charity/Permissions/Table.tsx
```diff
@@ -0,0 +1,129 @@
...
+                <CheckField<FormValues>
+                  name={`${fieldName}.gov_controlled`}
+                  classes={{
+                    label: "uppercase text-xs font-bold",
+                    input: "checkbox-orange",
+                  }}
+                  disabled={inputDisabled}
+                />
+              ) : null}
+              <CheckField<FormValues>
```
ap-justin, line 101: ISSUE: 
having previously set a delegate, 
image

clicking `delegate` checkbox should clear `input` before disabling

REPRO: go to `/admin/27/permissions`

ap-justin (changes requested):

src/pages/Admin/Charity/Permissions/useSubmit.ts
```diff
@@ -0,0 +1,85 @@
...
+  }: FormValues) {
+    try {
+      const diff = getPayloadDiff(initialValues, newValues);
+
+      if (isEmpty(Object.entries(diff))) {
+        return handleError("No changes detected");
+      }
+
+      const settingsController = new SettingsController(wallet);
+      const updateMsg = createUpdateEndowmentControllerMsg(id, diff);
```
ap-justin, line 40: bug: having, set all fields to `delegate` , when undelegating `account fees`, some fields are also undelegated like `beneficiary_allow_list`
image

REPRO: 
1. go to `admin/27/permissions`
2. check all fields are delegated to address
3. undelegate `account_fees`
4. verify `msg` in keplr wallet prompt 
5. notice `beneficiary_allowlist` is also undelegated

ap-justin (changes requested):

src/pages/Admin/Charity/Permissions/Permissions.tsx
misicnenad: Consider a case where `controller.endowment_controller.delegate === undefined` and the user's wallet is not connected to the webapp, i.e. `wallet?.address === undefined` -> this would make `(controller.endowment_controller?.delegate?.address  === wallet?.address) === true` thus making `userDelegate === true`, enabling unconnected users to make changes to the settings UI (the changes would still not pass due to backend validation).

ap-justin: that's right, resolving

src/pages/Admin/Charity/Permissions/Table.tsx
misicnenad: @SovereignAndrey @ap-justin we can leave the value in the input just on UI side, but not submit it with the rest of the data if `delegate` was set to `false` - this way the user is able to reenable the `delegate` checkbox without losing the address value?

src/pages/Admin/Charity/Permissions/Table.tsx
misicnenad: Unable to repro this, does the issue still occur?

src/pages/Admin/Charity/Permissions/Permissions.tsx
misicnenad: Unable to repro this, does the issue still occur?

src/pages/Admin/Charity/Permissions/Table.tsx
ap-justin: > So to determine whether a user is allowed to modify a specific row or not,

`B` and `C` should be able to modify every row (except `Endowment.endow_controller` row if it is included in this page ), since this particular page is about modifiying `EndowmentController` as a whole

the effects of other row/permission is manifested on pages of their scope 

Let's say after interacting with this page the ff. is the new permissions
```
controller (manifests in Edit Controller page) --> owner_controlled only
profile (manifests in Edit Profile page) -->  deligated only: to `A`
fees -(manifests in Fees page) -> deligated only:  to `A` 
```

it means if user `B` is trying to edit `EndowmentController` in edit controller page, he should be able to do so (and change every entries on it) since it is owner controlled 

but if user `B` tries to edit the `Profile` in edit profile page, he should not be able to do so since permission is only for `A`

..so forth for other pages scoped by a specific permission

ap-justin:
```
const userAuthorized: boolean = 
  (initDelegate && !!initDelegateAddress && initDelegateAddress !== wallet?.address) || (initOwnerControlled && propMeta.isAuthorized); // 1. or 2.
```
this logic is correct, given that `initDelegate` pertains only to `EndowController.endow_controller.delegate` since this particular permission is the one that matters to submit this edit controller page

ap-justin: sorry wrong issue, marked this as resolved a while ago

src/pages/Admin/Charity/Permissions/useSubmit.ts
misicnenad: This seems to be caused by a bug in `getPayloadDiff`, created https://github.com/AngelProtocolFinance/angelprotocol-web-app/pull/1947 to address this

src/pages/Admin/Charity/Permissions/useSubmit.ts
misicnenad: > currently, all forms in pages/Admin uses the 2nd approach
that is: forms are submittable - then prompts user on problems

@SovereignAndrey @ap-justin even though it's used on every other admin page, I would argue it is still bad UX and shouldn't be introduced here.
If a user is unable to change and submit anything on the page, that "thing" should be disabled for them to indicate so.

src/pages/Admin/Charity/Permissions/Table.tsx
misicnenad: > ```
> const userAuthorized: boolean = 
>   (initDelegate && !!initDelegateAddress && initDelegateAddress !== wallet?.address) || (initOwnerControlled && propMeta.isAuthorized); // 1. or 2.
> ```
> 
> this logic is correct, given that `initDelegate` pertains only to `EndowController.endow_controller.delegate` since this particular permission is the one that matters to submit this edit controller page

Will update the code accordingly

src/pages/Admin/Charity/Permissions/useSubmit.ts
misicnenad: Was caused by a combination of `getPayloadDiff` bug and a wrong assignment 🤦‍♂️ (see https://github.com/AngelProtocolFinance/angelprotocol-web-app/pull/1885/commits/bf9367d7c40d641ea068de391ac065a650029020)

src/pages/Admin/Charity/Permissions/Form.tsx
```diff
@@ -0,0 +1,43 @@
...
+      <Table className="max-lg:hidden" />
+      <MobileTable className="lg:hidden" />
+
+      <div className="flex flex-wrap items-center gap-3">
+        <button
+          type="reset"
+          disabled={isSubmitting}
+          className="btn-outline-filled py-2 px-8 rounded"
+        >
+          Reset
```
ap-justin, line 23: @misicnenad, `reset` isn't clear vs its wiring
say I have this starting 
image

when I click reset: 
it removes the delegated addresses but `delegate` is still checked
image

if `reset` pertains to resetting changes, then after clicking 
should just revert to original data when first going to permissions form

on the other hand if `reset` is pertains to `default`(unconfigured)
then should reset to this  state
image

we can't go halfway between since it's confusing

ap-justin (changes requested):

src/pages/Admin/Charity/Permissions/useSubmit.ts
ap-justin: @misicnenad, How can you be certain that this is bad UX ? 
>  I would argue it is still bad UX and shouldn't be introduced here.

this is to mitigate this, which was introduced in this PR
> but for someone who doesn't know that internals of this. It might be confusing to see why everything is disabled and user might try bunch of interactions just to make things enabled

by literally adding just this simple line in `submit`
```
isAuthorized: endow_controller.owner_controlled & propMeta.isAuthorized || endow_controller?.delegate?.address === wallet.address
```

src/pages/Admin/Charity/Permissions/Form.tsx
SovereignAndrey: @misicnenad `Reset` should revert the form to the last known on-chain settings that we get from the contract when the page loads.

ap-justin: @misicnenad, REPRO:
* go to `admin/27/permissions`
* click `reset`

src/pages/Admin/Charity/Permissions/useSubmit.ts
misicnenad: @ap-justin 
> How can you be certain that this is bad UX 

Quoting my [last comment](https://github.com/AngelProtocolFinance/angelprotocol-web-app/pull/1885#discussion_r1141664988):
> If a user is unable to change and submit anything on the page, that "thing" should be disabled for them to indicate as much.

> but for someone who doesn't know that internals of this. It might be confusing to see why everything is disabled and user might try bunch of interactions just to make things enabled

It would be confusing only if we do not add the top banner tooltip, read [my comment on this](https://github.com/AngelProtocolFinance/angelprotocol-web-app/pull/1885#discussion_r1136558155) and [Andrey's response](https://github.com/AngelProtocolFinance/angelprotocol-web-app/pull/1885#discussion_r1136627773): 

> > @SovereignAndrey @ap-justin might be useful for disconnected/unauthorized users to have a top banner on each page with a tooltip stating `"Your wallet needs to be connected and authorized to modify settings"` (a combination of the two messages from @ap-justin's comment, but we could have separate messages ofc). It could be similar to the one in the Withdraw page: image
> 
> That would be good. It's also a bigger issue than just on this page alone for how we ought to display forms to non-connected wallet users. Ideally, we have separate, non-form based views for when wallets are not connected in this admin area where the pages are publicly viewable and then also something like the above mentioned notice message to hint at other functionality if a wallet is connected.

We can leave things as-is for now as this is not and will not be in production yet, the tooltip banner can be created before it is pushed there @SovereignAndrey

src/pages/Admin/Charity/Permissions/Form.tsx
misicnenad: > @misicnenad `Reset` should revert the form to the last known on-chain settings that we get from the contract when the page loads.

Was expecting it to do as much, looking into this.

misicnenad: @SovereignAndrey @ap-justin the issue was using `<button type="reset" ...` and `<form onReset={...`, seems that `react-hook-form` doesn't know how to handle resetting fields when implemented this way. 
Updated to use `<button type="button" onClick={reset}...` logic (as is used usually) https://github.com/AngelProtocolFinance/angelprotocol-web-app/pull/1885/commits/a3a2c96759e889b1d5ad5b74066a3b54d8de9741

SovereignAndrey (approved):

PR #1944, opened 2023-03-17
Bug: not authorized error when executing passed poll

ap-justin:
Ticket(s):
-

* add propMeta.isAuthorized to `PollAction`

### Others
* removed unused packages

-

## UI changes for review


SovereignAndrey (approved):
Anyone can execute a proposal on-chain. Only the voting is restricted. If we have to pass the `isAuthorized` then we might just allow all passing `true`. 

We can leave as is here in this PR for now, will talk to Lambert about how he wants that set on the UI. :+1:

PR #1959, opened 2023-03-21
FIX: NPO Type changes

SovereignAndrey:
Ticket(s):
- N/A. Change requests covered here came out of product meeting discussions. 

- NPO Type marketplace filter title changes to be "NPO Type" 
- Filter display label/value uses "Charity" instead of "Non-Profit" (All NPOs are Non-Profits, but not all NPOs are Charities)
- Moves the Profile NPO Type tag into Details Column, replacing the existing tag with placeholder value
- Change Marketplace Card labels to reflect the NPO Type (since we only have non-profits now)

**NOTE:** A lambda script is now ready in AWS to change all of the `endowments` table records' `endow_designation` field from `Non-Profit` >> `Charity` for mainnet, right after merging this PR in. :+1: 

### Marketplace filters
npo-type-filters

### Marketplace Card
npo-type-marketplace-card

### Profile
npo-type-profile


src/pages/Marketplace/Cards/Card.tsx
```diff
@@ -16,7 +16,7 @@ export default function Card({
   name,
   image,
   id,
-  endow_type,
+  endow_designation,
```
misicnenad, line 19: Just noticed we have both [EndowDesignation](https://github.com/AngelProtocolFinance/angelprotocol-web-app/blob/master/src/types/aws/ap/index.ts#L90) and [EndowmentType](https://github.com/AngelProtocolFinance/angelprotocol-web-app/blob/master/src/types/contracts/common.ts#L51), what will be the difference between the two?

misicnenad (approved):

src/pages/Marketplace/Cards/Card.tsx
SovereignAndrey: > Just noticed we have both [EndowDesignation](https://github.com/AngelProtocolFinance/angelprotocol-web-app/blob/master/src/types/aws/ap/index.ts#L90) and [EndowmentType](https://github.com/AngelProtocolFinance/angelprotocol-web-app/blob/master/src/types/contracts/common.ts#L51), what will be the difference between the two?

`EndowmentType` is the Contract level endowment differentiation (`Charity` vs `Normal`), whereas `EndowDesignation` is a purely AWS DB field(string type) that further differentiates the Charity endowments.

PR #1961, opened 2023-03-21
FIX: redundant max buttons on token-field (gift-purchase)

ap-justin:
Ticket(s):
-
image
* caused by logic `step > balance ? "max" : step`

only show steps when user's balance is greater than all steps. since steps 
are computed based on tokens `min_donation_amount` (not fixed like it's overkill to show users `10, 50, 100, 200` when 
donating `ETH` for example )

prevents the need to handle edges
1. given steps `1.0, 2.0, 3.0, 4.0, 5.0` what if user's balance is `3.2` should we 
* show `max` in like `1, 2, max, 4, 5`
* limit options to `1, 2, max`

also reduces form errors, since clicking a step ` > balance` would just show "not enough balance"

-

## UI changes for review

### DIFF
--- a/src/components/TokenField/Steps.tsx
+++ b/src/components/TokenField/Steps.tsx
@@ -9,26 +9,27 @@ export default function Steps({
 }: AmountOptionsProps) {
   const base = token.min_donation_amnt || 100;
   const precision = new Decimal(token.min_donation_amnt).decimalPlaces();
+  const steps = scale.map((s) => roundDownToNum(s * base, precision));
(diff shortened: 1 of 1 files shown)


SovereignAndrey (approved):
Thanks for the quick fix. :+1:  I think in the end we ought to simply remove these buttons altogether(maybe we can do that in v1.8?), like we did for the donation form, since it simplifies the interface and removes an possible confusion over what the various numbers mean.

PR #1962, opened 2023-03-21
Updated Types for Endowment Querying Params Due To V2.0 Contract Specs

jp-mariano:
Ticket(s):
- N/A

An update in AWS DB table is made in order to accept the new lowercase values for `endow_type`. There's also a new field called `endow_tier` that will replace the `tier` once v2.0 of the smart contracts is live in production.

This PR is created to update the `params` used for Marketplace cards querying. The `tiers` param's name remains unchanged and the required check is done in the Lambda itself.

- Go to Marketplace and there should just be two endowment cards visible (endowment 11 and 19)

## UI changes for review

No major UI changes.

### DIFF
--- a/src/types/contracts/common.ts
+++ b/src/types/contracts/common.ts
@@ -49,7 +49,6 @@ export type EndowmentStatus = {
   closed: 3;
 };
 export type EndowmentType = "charity" | "normal";
-// export type CapitalizedEndowmentType = Capitalize<EndowmentType>;
 export type EndowmentStatusText = keyof EndowmentStatus;
 export type EndowmentStatusNum = EndowmentStatus[EndowmentStatusText];
 export type EndowmentStatusStrNum = `${EndowmentStatusNum}`;
--- a/src/types/contracts/common.ts
+++ b/src/types/contracts/common.ts
@@ -52,7 +52,6 @@ export type EndowmentType = "charity" | "normal";
 export type EndowmentStatusText = keyof EndowmentStatus;
 export type EndowmentStatusNum = EndowmentStatus[EndowmentStatusText];
 export type EndowmentStatusStrNum = `${EndowmentStatusNum}`;
-// export type EndowmentTier = "Level1" | "Level2" | "Level3";
 export type EndowmentTierNum = 1 | 2 | 3;
 
 export type Categories = {
--- a/src/types/contracts/common.ts
+++ b/src/types/contracts/common.ts
@@ -49,11 +49,11 @@ export type EndowmentStatus = {
   closed: 3;
 };
 export type EndowmentType = "charity" | "normal";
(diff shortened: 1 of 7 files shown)


src/types/contracts/common.ts
```diff
@@ -49,11 +49,11 @@ export type EndowmentStatus = {
   closed: 3;
 };
 export type EndowmentType = "charity" | "normal";
-export type CapitalizedEndowmentType = Capitalize<EndowmentType>;
+// export type CapitalizedEndowmentType = Capitalize<EndowmentType>;
```
misicnenad, line 52:
```suggestion
```
We can remove this since it's no longer necessary

src/types/contracts/common.ts
```diff
@@ -49,11 +49,11 @@ export type EndowmentStatus = {
...
 export type EndowmentStatusText = keyof EndowmentStatus;
 export type EndowmentStatusNum = EndowmentStatus[EndowmentStatusText];
 export type EndowmentStatusStrNum = `${EndowmentStatusNum}`;
-export type EndowmentTier = "Level1" | "Level2" | "Level3";
+// export type EndowmentTier = "Level1" | "Level2" | "Level3";
```
misicnenad, line 56:
```suggestion
```
We can remove this since it's no longer necessary.

src/types/contracts/common.ts
```diff
@@ -49,11 +49,11 @@ export type EndowmentStatus = {
   closed: 3;
 };
 export type EndowmentType = "charity" | "normal";
-export type CapitalizedEndowmentType = Capitalize<EndowmentType>;
+// export type CapitalizedEndowmentType = Capitalize<EndowmentType>;
 export type EndowmentStatusText = keyof EndowmentStatus;
 export type EndowmentStatusNum = EndowmentStatus[EndowmentStatusText];
 export type EndowmentStatusStrNum = `${EndowmentStatusNum}`;
-export type EndowmentTier = "Level1" | "Level2" | "Level3";
```
misicnenad, line 56: > There's also a new field called `endow_tier` that will replace the `tier` once v2.0 of the smart contracts is live in production... The `tiers` param's name remains unchanged and the required check is done in the Lambda itself.

@jp-mariano will this field be renamed once contracts are live?

misicnenad (changes requested):

src/types/contracts/common.ts
jp-mariano: Yes, this was necessary to update the field's type from `String` to `Integer` without breaking the live app. Right now, the check is applied to the Lambda directly so that it knows if `"Level3"` or `3` is passed then it will use the correct field name.

misicnenad (approved):

PR #1966, opened 2023-03-21
Submit ast to polygon

ap-justin:
Ticket(s):
content
* #1969 
* #1980 
* #1972 

* like `contracts/ERC20` created another interface `contracts/Account` using `@ethersproject/abi` . While it's true we could just use convenience library like `@ethersproject/Contract`. It as was opted not use it ( just yet ) for ff. reasons

#### 1. `@ethersproject/Contract` requires `@ethersproject/provider` which essentially makes us just install the `ethers` umbrella package as a whole (`+100kb`) bundle size. 

This reason is quite trivial compared to the benefits (`generated types` which lowers maintenance efforts ) we'll gain if we just use `ethers` + `hardhat-ethers`. On the other hand, manually creating types seem not that hard to do atm and got some benefits as well

* More control on how that type would look like e.g (enforce value)
e.g. in `type NewAIF`
```ts
  dao: DaoSetup; //just set to placeholder - overriden by creatDao:bool
  createDao: false; //not included in launchpad, for edit later
  proposalLink: 0;
```
* We only create `types` the web-app needs 
* The types we create closely resembles type in `solidity contracts`

```ts
//solidity
address owner; 
bool withdrawBeforeMaturity;
uint256 maturityTime; 
uint256 maturityHeight;

//created
address: string;
withdrawBeforeMaturity: boolean;
maturityTime: number;
maturityHeight: number;

//generated
 owner: PromiseOrValue<string>;
 withdrawBeforeMaturity: PromiseOrValue<boolean>;
 maturityTime: PromiseOrValue<BigNumberish>;
 maturityHeight: PromiseOrValue<BigNumberish>;
```

#### 2. Use of convenience `class Contract` doesn't fit well on our current tx flows 
We process transactions this way
```yaml
1. create `tx` content - where we include the type of tx and its arguments and then encode - then forget let this go into the wild
2. simulate `tx` from step 1- where we could get `fees` to show to users
3. submit `simulated-tx` from step 2 - in this step, we just submit with a guarantee that this tx is constructed with the correct arguments and validity checked via simulation
```
steps 1-3 is tightly coupled to `class Contract`  that is:
```
class Contract = {
   transfer //call this to send tx
   estimateGas // call this to estimate gas
   ...
}
```

so the contract needs to be visible on each steps 
```ts
const ERC20 = new Contract(ERC20abi)

//1. tx data can be extracted with
const data = ERC20.interface.encodeFunctionData("balanceOf", "address")
//2. for estimate need visibility of ERC20 contract
const gas  = ERC20.transfer(/** input arguments again */).estimateGas()
//3. for submission, need visibility of ERC20 contract
const result  = ERC20.transfer(/** input arguments again */).estimateGas()
```
It would seem like, this is just minor inconvenience but as the number of different `Contracts` to process with this flow 
the harder it gets 
```ts
//1. one of these could be the data you need to send
//polygon
const a = new A("a-abi").encodeFunctionData("a",[/** a args */])
const b = new B("b-abi").encodeFunctionData("b",[/** b args */])
const c = new C("c-abi").encodeFunctionData("c",[/** c args */])

const D  = {to from value} //send eth

//juno
const x = MsgExecuteContract //contract
const y = MsgSend

//2. estimate transaction 
if( msg ) {/** easy to estimate since just Msg in general */ } 

//there's a need to distinguish between just sending funds/ vs contract interaction
if ( contract )  {} // ??? Which of `A,B,C,` should I run the `estimateGas`, and which `method` ?

//3. submit 
if(msg) {/** juno: just general Msg  */ }

if(contract) { }
// ?? which of `A,B,C` should I run and which method 
// is it A.a() ?
// is it A.b() ?
// is it A.c() ?

```

## Disadvantages of the solution 
1. you create your own type -and update it as needed /  which what we are doing so far with `juno` 
2. the need to get abi fragments needed - also update as needed 

## Advantages of the solution
1. lightweight( just using `@ethersproject/abi` )
2. flexible and process agnostic ( just create your tx at start and it no need to worry how to process it at succeeding stages)

-

## UI changes for review


SovereignAndrey (approved):
Very very nice work on this. Despite the AWS error, I'm happy to move this PR along into the RC-v1.8 branch. We can follow up with smaller PR if any tweaks are needed for that once it gets implemented on devops side.

Issue #1928, opened 2023-03-16
Update/refactor Register Signup page layout to match Figma


SovereignAndrey:
All the `mt-8` & `my-11` can be replaced with adding `gap-8` to [parent form's class name ](https://github.com/AngelProtocolFinance/angelprotocol-web-app/blob/master/src/pages/Registration/Signup/Form.tsx#L16) and one `-mt-4` to [CheckField](https://github.com/AngelProtocolFinance/angelprotocol-web-app/blob/master/src/pages/Registration/Signup/Form.tsx#L28), as it has a smaller top margin.

All classes `mx-0 sm:mx-24` can be replaced with adding `justify-center` to parent form's class name. Header's `text-center` can then also be removed.

**Current:**
![Image](https://user-images.githubusercontent.com/19427053/223177318-052df008-33bc-423a-a29d-60f9f718e92e.png)

**Expected:**
![Image](https://user-images.githubusercontent.com/19427053/223177229-d2b3ed9a-0dba-44e6-807b-c56bb4305ad6.png)

PR #1967, opened 2023-03-21
Update/refactor Register Signup page layout to match Figma

abbasali1994:
Ticket(s):
- https://github.com/AngelProtocolFinance/angelprotocol-web-app/issues/1928

Mainly removed unnecessary classes from fields and added common class to parent tag

-

## UI changes for review
image

### DIFF
--- a/src/pages/Registration/Signup/Form.tsx
+++ b/src/pages/Registration/Signup/Form.tsx
@@ -1,6 +1,8 @@
 import { Link } from "react-router-dom";
 import { FormValues as FV } from "./types";
+import { useModalContext } from "contexts/ModalContext";
 import ExtLink from "components/ExtLink";
+import WalletModal from "components/WalletSuite/WalletSelectorOpener/WalletModal";
 import { CheckField, Field } from "components/form";
 import { Separator } from "components/registration";
 import { APP_NAME } from "constants/common";
@@ -12,7 +14,9 @@ const NEED_HELP_ARTICLE_ID = 6628120;
 
 export default function Form({ classes = "" }: { classes?: string }) {
   const { submit, isSubmitting } = useSubmit();
+  const { showModal } = useModalContext();
 
+  const handleClick = () => showModal(WalletModal, {});
   const openIntercomHelp = () => {
     const w = window as any;
     if ("Intercom" in w) {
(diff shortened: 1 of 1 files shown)


SovereignAndrey (approved):

src/pages/Registration/Signup/Form.tsx
```diff
@@ -23,7 +23,7 @@ export default function Form({ classes = "" }: { classes?: string }) {
   return (
     <form
       onSubmit={submit}
-      className={`${classes} padded-container w-full max-w-[37.5rem] grid`}
+      className={`${classes} justify-center gap-8 padded-container w-full max-w-[37.5rem] grid`}
     >
       <h3 className="text-3xl font-bold text-center">
```
misicnenad, line 28: nit: `text-center` can probably be removed (as it's already justified to center).

src/pages/Registration/Signup/Form.tsx
```diff
@@ -23,7 +23,7 @@ export default function Form({ classes = "" }: { classes?: string }) {
...
         {`Register to ${APP_NAME}`}
```
misicnenad, line 29: The "Connect wallet" section is missing.
image

Do we not need this @SovereignAndrey ? Users occasionally get confused and already registered users might find this tooltip/section useful in remembering what they need to do to "sign in".

misicnenad (changes requested):

src/pages/Registration/Signup/Form.tsx
SovereignAndrey: Good catch! Yes we need that in there.

SovereignAndrey (changes requested):
@abbasali1994 as @misicnenad noticed, we are missing the connect wallet part (see the expected design layout in the issue)

src/pages/Registration/Signup/Form.tsx
abbasali1994: @misicnenad what if user has already connected a wallet, do we display this statement then too?

src/pages/Registration/Signup/Form.tsx
abbasali1994: @misicnenad it didnt work cause justify-center center aligns the block but thetext within the block remains left aligned
text-center aligns the the text within the block of <h3> to center

src/pages/Registration/Signup/Form.tsx
SovereignAndrey: These are all great points @abbasali1994 & @misicnenad. I think, for now, we can revert this last commit, leaving the connect wallet out, as it might cause more confusion with it in there than not. It doesn't actually do **anything** at this stage (though maybe we might want to pull applications by that wallet in the future that are pending and show them a list?). I'll discuss with Lambert tomorrow and circle back. :)

SovereignAndrey (approved):

PR #1969, opened 2023-03-22
AST - polygon: create multi-tx helper 

ap-justin:
Ticket(s):
-

-

## UI changes for review


misicnenad (changes requested):

src/pages/Gift/Purchase/Submit/estimateTx.ts
```diff
@@ -1,22 +1,19 @@
...
-import { ConnectedWallet } from "@terra-money/wallet-provider";
-import { Estimate } from "./types";
+import { Estimate } from "types/tx";
 import { WalletState } from "contexts/WalletContext";
 import { SubmitStep } from "slices/gift";
 import CW20 from "contracts/CW20";
 import GiftCard from "contracts/GiftCard";
-import { extractFeeAmount, logger, scaleToStr } from "helpers";
+import { logger, scaleToStr } from "helpers";
+import { estimateTx as estimateGas } from "helpers/tx";
```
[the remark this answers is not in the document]
ap-justin: because a  function name `estimateTx` is already defined in this file `Submit/estimateTx`

src/helpers/tx/estimateTx/estimateTx.ts
```diff
@@ -0,0 +1,25 @@
...
+    switch (content.type) {
+      case "cosmos":
+        return estimateCosmosFee(wallet, content.val);
+      case "terra":
+        return estimateTerraFee(wallet, terraWallet!, content.val);
+      default:
+        return estimateEVMFee(wallet, content.val);
+    }
+  } catch (err) {
+    return null;
```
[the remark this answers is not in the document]
ap-justin: if you look at the consumers of this function. The `null` case is handled at the point where it is called giving more refined flows

```ts
const estimate = esimateTx()
if(!estimate) return "Simul Failed" 
//done at this point


```



you are proposing 
```ts
return Promise<Estimate>
```
and in the consumers 
```ts
try { 
const estimated = estimateTx()
const tx = send(estimated)

} catch (err} {
  
  // determine what is `err` and show appropriate error

}

```

to make this work we need additional type of `error` say `SimulateError`

and in catch block we do
```ts
// sendError, typeError, other type of errors exist here
if(err instance of SimulateError){
  // Simul failed
} 
```

src/pages/Launchpad/Summary/useSubmit/index.ts
```diff
@@ -27,60 +25,100 @@ export default function useSubmit(network: Network) {
       return showModal(TxPrompt, { error: "Wallet is not connected" });
```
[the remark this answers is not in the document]
ap-justin: > This whole function should be wrapped in a try/catch 
fixed

>  Optionally, all current expected errors could be thrown like throw new Error("Wallet is not connected") and let catch clause deal with it. 

if an undesirable event is known and can be handled, there's no reason to rethrow it and handle it in `catch`

src/helpers/tx/estimateTx/estimateTx.ts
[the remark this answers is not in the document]
misicnenad: > to make this work we need additional type of error say SimulateError

Might be a good idea. We already have this type of error logic throughout the webapp, see [errors.ts](https://github.com/AngelProtocolFinance/angelprotocol-web-app/blob/ast-polygon/src/errors/errors.ts). The main benefit is less intermediate checks for errors, to use your example it would mean instead of:
```ts
try { 
  const estimated = estimateTx()
  if (!estimated) {
    return // handle error somehow;
  }
  
  const tx = send(estimated) 
  if (!tx || !!tx.error) {
    return // handle error somehow;
  }
  ...
} catch (err} {
  // still have to check and handle any unanticipated errors
}
```

we would have:
```ts
try { 
  const estimated = estimateTx()
  // assume everything is fine, as otherwise `catch` clause comes into play
  const tx = send(estimated)
  // assume everything is fine, as otherwise `catch` clause comes into play
  ...
} catch (err} {
  // determine what is `err` and show appropriate error
  if(err instance of SimulateError){
    // Simul failed
  } 
  if(err instance of SimulateError2){
    // Simul failed
  } 
  ...
}
```
But we already have `ErrorContext.handleError` which does the `if(err instance of SimulateError)` checks. We could modify it or create a new `ErrorContext.handleTxError` function that uses `TxPrompt` to display errors, so we would get something even shorter:
```ts
const { handleTxError } = useErrorContext();
...
try { 
  const estimated = estimateTx()
  // assume everything is fine, as otherwise `catch` clause comes into play
  const tx = send(estimated)
  // assume everything is fine, as otherwise `catch` clause comes into play
  ...
} catch (err} {
  handleTxError(err);
}
```

src/pages/Gift/Purchase/Submit/estimateTx.ts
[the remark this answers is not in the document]
misicnenad: Makes sense.

src/helpers/tx/sendTx/sendEVMTx.ts
```diff
@@ -0,0 +1,27 @@
...
+      params: [tx],
+    });
+
+    return { hash, chainID: wallet.chain.chain_id };
+
+    //TODO:
+    //1. get transaction_receipt by polling
+    //2. get id from logs
+  } catch (err) {
+    return { error: "Error encountered while sending transaction" };
```
misicnenad, line 25: Should log `err` at least, otherwise actual error info is lost.

misicnenad: Don't see a need for this yet, but if we ever need the actual error info in calling components, an alternative approach to error creation/throwing would be to return an object `type Error = { inner: Error; message: string }` where `inner` would contain any "actual" errors that occurred (similar to how [C# Exceptions work](https://learn.microsoft.com/en-us/dotnet/api/system.exception.-ctor?view=net-7.0#system-exception-ctor(system-string-system-exception))).

src/helpers/tx/sendTx/sendTerraTx.ts
```diff
@@ -0,0 +1,43 @@
...
+        error: "Transaction submitted but failed",
+        tx: err.txhash
+          ? {
+              hash: err.txhash,
+              chainID: wallet.network.chainID,
+            }
+          : undefined,
+      };
+    }
+    return { error: "Error encountered while sending transaction" };
```
misicnenad, line 41: Should log `err`.

src/pages/Launchpad/Summary/useSubmit/index.ts
```diff
@@ -1,86 +1,133 @@
...
-        data: createEndowment.encode(toEVMAIF(completed, wallet.address)),
-      };
-      const { tx: simulated } = await estimateFee(wallet, tx);
-      const hash = await sendEVMTX(wallet, simulated);
-      return alert(hash);
-    }
+      const { chain } = wallet;
+      if (
+        network === "polygon" &&
+        !(chain.chain_id === chainIds.polygon || chain.chain_id === "1337") //polygon local
```
misicnenad, line 34: Maybe we should add local chain IDs to `const chainIds`? We handle mainnet and testnet IDs already in there, adding local IDs would help.

src/pages/Launchpad/Summary/useSubmit/index.ts
```diff
@@ -1,86 +1,133 @@
...
+      if (network === "juno" && chain.chain_id !== chainIds.juno) {
+        return showModal(TxPrompt, {
+          error: "Please connect to Juno network",
+        });
+      }
+
+      // //////////////// CONSTRUCT TX CONTENT ////////////////////
+      let content: TxContent;
+      if (
+        network === "polygon" /**TODO: && check connected wallet's chainID */
```
misicnenad, line 59: `/**TODO: && check connected wallet's chainID */`
Shouldn't this check also be included here?

src/types/tx.ts
```diff
@@ -0,0 +1,75 @@
...
+      val: CreateTxOptions;
+      wallet: ConnectedWallet /**future client/provider will be included in wallet */;
+    }
+  | { type: "evm"; val: EVMTx };
+
+export type SubmittedTx = { hash: string; chainID: string };
+
+export type TxLoading = { loading: string };
+export type TxError = { error: string; tx?: SubmittedTx };
+export type TxSuccess = SubmittedTx & {
```
misicnenad, line 28: nit: could've kept `TxSuccess` consistent with `TxError` and have it have a nested field `tx: SubmittedTx`.

misicnenad (changes requested):

src/helpers/tx/estimateTx/estimateTx.ts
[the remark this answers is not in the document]
ap-justin: one of the considerations why errors are handled upfront is that not every flow on our app falls into this 
general show `prompt` on error category

here are examples
image

image

src/helpers/tx/sendTx/sendEVMTx.ts
ap-justin: > Don't see a need for this yet, but if we ever need the actual error info in calling components, an alternative approach to error creation/throwing would be to return an object type Error = { inner: Error; message: string } 

This is already done in this PR
```ts
export async function sendEVMTx(
  wallet: WalletState,
  tx: EVMTx,
  attribute?: string //eslint-disable-line
): Promise<TxResult> {
```

where TxResult can be 
```
export type TxResult = TxError | TxSuccess;
```

src/pages/Launchpad/Summary/useSubmit/index.ts
ap-justin: "1337" will be removed once we are interfacing with testnet contracts. 
removes the need to 
```
const chainIDs = IS_TEST ? { }  : IS_LOCAL ? { } : { }
```

src/types/tx.ts
ap-justin: having them different is easier for type guards,
```
//success
{ hash, chainID }

//error
{error tx? } 
```
 no need for them to be consistent since they are two types that manifest different scenario

src/hooks/useCosmosTxSender.ts
```diff
@@ -58,54 +55,42 @@ export default function useCosmosTxSender<T extends boolean = false>(
...
+      const estimate = await estimateTx({ type: "cosmos", val: msgs }, wallet);
+      if (!estimate) {
+        return showModal(TxPrompt, {
+          error: "Simulation failed. Transaction likely to fail",
+        });
+      }
+      const { fee, tx } = estimate;
 
-      if (feeAmount > wallet.displayCoin.balance) {
+      if (fee.amount > wallet.displayCoin.balance) {
```
misicnenad, line 68: Shouldn't the below `showModal(Popup` also be `showModal(TxPrompt`?

src/helpers/tx/estimateTx/estimateTx.ts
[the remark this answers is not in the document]
misicnenad: About using `handleTxError` as part of `ErrorContext`, it was just an example of a wrapper function for `showModal(TxPrompt, ...)` which is used in different files, but it could've been understood even as a local helper function.

Point was that just throwing errors and handling them in one place might result in code that's even easier to read/understand then code with many `if/else`s throughout it.

IMO, this:
```ts
try { 
  const estimated = estimateTx()
  // assume everything is fine, as otherwise `catch` clause comes into play
  const tx = send(estimated)
  // assume everything is fine, as otherwise `catch` clause comes into play
  ...
} catch (err} {
  handleTxError(err);
}
...
function handleTxError(error: any) {
 // handle error
}
```
is a bit clearer than this:
```ts
try { 
  const estimated = estimateTx()
  if (!estimated) {
    return // handle error somehow;
  }
  
  const tx = send(estimated) 
  if (!tx || !!tx.error) {
    return // handle error somehow;
  }
  ...
} catch (err} {
  // still have to check and handle any unanticipated errors
}
```

Maybe this is a matter of personal taste.

Anyway, this was just a suggestion, let's not delve into that any further.

src/helpers/tx/sendTx/sendEVMTx.ts
misicnenad: > This is already done in this PR

I don't see any error nesting in this PR. 
And there's no apparent reason for there to be one, it's just an idea I wanted to share.

misicnenad (approved):

SovereignAndrey (approved):
Nice work!

Issue #1911, opened 2023-03-14
Admin Withdraw: Re-wired to send msgs to Polygon contracts


SovereignAndrey:
Withdraw supported networks in v1.8:
- Polygon (obviously!)
- Ethereum
- Biniance

misicnenad:
@SovereignAndrey which networks should be supported for withdrawal then, only Juno and Polygon?
What about the currently "available" (but not really) Ethereum and Binance? Should we remove them since the user is unable to withdraw to addresses on those networks (as per code in the [useWithdraw](https://github.com/AngelProtocolFinance/angelprotocol-web-app/blob/RC-v1.8/src/pages/Admin/Charity/Withdraws/Withdrawer/WithdrawForm/useWithdraw.ts#L35), used to submit the withdrawal):
```ts
...
    const isJuno = data.network === chainIds.juno;
    //if not juno, send to ap wallet (juno)
    const beneficiary = isJuno ? data.beneficiary : ap_wallets.juno_withdraw;
...
```

SovereignAndrey:
> @SovereignAndrey which networks should be supported for withdrawal then, only Juno and Polygon?
> 
> What about the currently "available" (but not really) Ethereum and Binance? Should we remove them since the user is unable to withdraw to addresses on those networks (as per code in the [useWithdraw](https://github.com/AngelProtocolFinance/angelprotocol-web-app/blob/RC-v1.8/src/pages/Admin/Charity/Withdraws/Withdrawer/WithdrawForm/useWithdraw.ts#L35), used to submit the withdrawal):
> ```ts
> ...
>     const isJuno = data.network === chainIds.juno;
>     //if not juno, send to ap wallet (juno)
>     const beneficiary = isJuno ? data.beneficiary : ap_wallets.juno_withdraw;
> ...
> ```

@misicnenad We would need to support the same withdrawal logic we have today, in that if the withdrawal is to the native chain of the accessing Endowment, either Juno OR Polygon, then take the wallet address that is passed as the destination of the withdraw. If any other chain, then we need to move the funds to the AWS withdrawals wallet (Juno or Polygon depending on the native chain of the endowment being accessed) and do the usual POST to AWS to relay the x-chain withdrawal info.

misicnenad:
@SovereignAndrey what are AP's `polygon withdraw` and `polygon deposit` addresses?
[...]

PR #1970, opened 2023-03-22
Wire withdraw polygon

misicnenad:
Ticket(s):
- https://github.com/AngelProtocolFinance/angelprotocol-web-app/issues/1911

## Note
Setting this to "Ready for Review" in case there's something I missed (and whoever reviews this catches it).

Still trying to figure out how to test this out. 
Tried creating a Charity on Polygon, but stumbled upon an issue with new endowment IDs being already used by existing charities. ~~Not sure how to handle this, trying to figure something out with @jp-mariano.~~ @jp-mariano let me know this is still WIP, so it will be hard to test out this way. If you have any ideas how to test this out, please let me know.


src/contracts/evm/AccountDepositWithdrawEndowments/abi.json
```diff
@@ -0,0 +1,4 @@
+[
+  "event UpdateEndowmentState(uint256,tuple(tuple(uint256,uint256),tuple(tuple(uint256,uint256[],address[]),tuple(uint256,uint256[],address[])),bool,tuple(tuple(uint256,address),uint8)))",
```
ap-justin, line 2: fragment `event UpdateEndowmentState` could just be added to Accounts abi as this is facet (we don't interact with this directly) of that contract

ap-justin (changes requested):

src/contracts/evm/AccountDepositWithdrawEndowments/abi.json
misicnenad: Makes sense

src/contracts/evm/index.ts
```diff
@@ -0,0 +1,3 @@
+export * as Account from "./Account";
+export * as EndowmentMultiSig from "./EndowmentMultiSig";
+export * as LockedWithdraw from "./LockedWithdraw";
```
misicnenad, line 3: @ap-justin do you you also think this is a good way to use these contract functions?
It's very easy to understand what's going on when one sees this in code:
```ts
// ok, so we import the Account contract
import { Account } from "contracts/evm";
...
// and now we use its `withdraw` function encoder
Account.withdraw.encode(...);
```

Didn't add `contracts/evm/ERC20` here, but it can be easily done.

ap-justin: @misicnenad, I realized that this construct
```
return {
  encode ()  iface.encode (...)
  decode () iface.decode( ... ) 
}
```

is just uncessary wrapper and discontinued using this in
* #2025 
for the new tx migration


going forward I opted for a more declarative

```ts
const tx = createTx("whatTx?", { ... payload_depending_on_chosen_tx } )

``` 

where the `encode/decode` call is called directly in `txs` list
```ts
export const txs: { [T in TxTypes]: (args: TxArgs<T>) => string } = {
  "accounts.create-endowment": (aif) =>
    accounts.encodeFunctionData("createEndowment", [toTuple(aif)]),
  "accounts.update-controller": (update) =>
    accounts.encodeFunctionData("updateEndowmentController", [toTuple(update)]),
  "multisig.submit-transaction": (tx) =>
    multisig.encodeFunctionData("submitTransaction", toTuple(tx)),
  "multisig.add-owner": (newOwner) =>
    multisig.encodeFunctionData("addOwner", toTuple(newOwner)),
  "erc20.transfer": (transfer) =>
    erc20.encodeFunctionData("transfer", toTuple(transfer)),
};
```

and the interface is the only thing imported from `contracts/evm`
```ts
import { erc20 } from "contracts/evm/ERC20";
import { multisig } from "contracts/evm/multisig";
import { accounts } from "../evm/Account";
```

is it possible, to point this PR to #2025 ?

misicnenad: @ap-justin 
I realized the current design was just a temporary one until we find something better. 

> is it possible, to point this PR to https://github.com/AngelProtocolFinance/angelprotocol-web-app/pull/2025 ? so with more use cases for createTx helper, we could test if it's the right structure

Sure.
Did you consider using a design similar to how we implemented Juno contract message creation, like [contracts/Account](https://github.com/AngelProtocolFinance/angelprotocol-web-app/blob/RC-v1.8/src/contracts/Account.ts)?


Also a package worth looking into is [typechain](https://github.com/dethcrypto/TypeChain), do you know about it? 
It is a tool that generates TS types for EVM contracts, might be something that we can use on the webapp, but need to investigate further. 
@SovereignAndrey do we have a task for this, I'd gladly pick it up?

ap-justin: > Did you consider using a design similar to how we implemented Juno contract message creation, like
I decided not to copy that same implementation which used as 
```
const account = new Account("", wallet)
const tx = account.createEndowTx()

```
and too many boillerplate in implementation (only difference is that inner content
```
createEmbeddedChangeEndowmentStatusMsg(payload: StatusChangePayload) {
    return this.createEmbeddedWasmMsg(Account.address, {
      update_endowment_status: payload, //<--- inner content
    });
  }

  createEmbeddedWithdrawMsg(payload: WithdrawPayload) {
    return this.createEmbeddedWasmMsg(Account.address, {
      withdraw: payload, // <--- inner content
    });
  }

  createEmbeddedInvestMsg(payload: InvestPayload) {
    return this.createEmbeddedWasmMsg(Account.address, {
      vaults_invest: payload,  // <--- inner content
    });
  }

  createEmbeddedRedeemMsg(payload: RedeemPayload) {
    return this.createEmbeddedWasmMsg(Account.address, {
      vaults_redeem: payload,  // <--- inner content
    });
```

could be refactored simply to something like
```
  "cw20.transfer": (args) => ({
    transfer: args,
  }),
  "cw20.send": ({ amount, contract, msg }) => ({
    amount,
    contract,
    msg: toBase64(msg),
  }),
```

and used as 
```ts
const msg = createMsg("cw20.transfer", { adaptive_payload})
```


> Also a package worth looking into is [typechain](https://github.com/dethcrypto/TypeChain), do you know about it?
It is a tool that generates TS types for EVM contracts, might be something that we can use on the webapp, but need to investigate further.

I decided against it in this PR
* #1966

misicnenad: > and too many boillerplate in implementation (only difference is that inner content

There is some boilerplate there, but it is very easy to understand and update - to add a new contract function, you just need to create a new `create[TargetName]Msg(...)`.
I've been looking at the contract design in https://github.com/AngelProtocolFinance/angelprotocol-web-app/pull/2025 and still trying to understand what is going on. Will give it some further thought.

misicnenad: > is it possible, to point this PR to https://github.com/AngelProtocolFinance/angelprotocol-web-app/pull/2025 ? so with more use cases for `createTx` helper, we could test if it's the right structure

I'd rather not do this now, as it's been taking me quite some time to even start to get a hang of the proposed design. It's very complex in its abstractions and its type-safety is on the weaker side.
Would rather just merge this as-is and let you update this to the new design, as you already understand how it's supposed to work.

misicnenad: My main points of contention were:
> I've been looking at the contract design in https://github.com/AngelProtocolFinance/angelprotocol-web-app/pull/2025 for 2 hours and still trying to understand what is going on

Don't want to rewrite the whole PR to use a yet unreviewed (and not yet agreed on) design, especially one which is taking time to fully comprehend. We'll rewrite the code here when we get to merging the design into the `RC-v1.8` branch, but for now it's better we keep it as-is.

> It's very complex in its abstractions

This design is admittedly very clever in the way it's set up to handle the problem of creating EVM txs, but this makes the code MUCH harder to understand. 
Just looking at [type TxOptions](https://github.com/AngelProtocolFinance/angelprotocol-web-app/blob/evm-query/src/contracts/createTx/types.ts#L36) made my head spin. Similarly, but to a lesser degree goes for [encodeTx](https://github.com/AngelProtocolFinance/angelprotocol-web-app/blob/evm-query/src/contracts/createTx/createTx.ts#L20). An even lesser thing thing is that types, data and functions for the same contracts are all defined in separate files, which may be understandable for the setup as far as I understood it, but still more complex than the design we use for Juno.

> How is this not typesafe ? it shows you exactly the txs you could create
> image
> ...

Answered the question in the PR draft itself where it's more appropriate, see https://github.com/AngelProtocolFinance/angelprotocol-web-app/pull/2025#issuecomment-1499530172.

misicnenad:
@SovereignAndrey @ap-justin 

> that's why I'm recommending to point this PR on https://github.com/AngelProtocolFinance/angelprotocol-web-app/pull/2025 (and be set as draft), since work is being done there that will enable the functionality for this

I wouldn't feel comfortable rewriting a whole PR to use a yet unreviewed (and not yet agreed on) design. We can rewrite this code if and when we agree on merging the proposed design into the `RC-v1.8` branch, but for now we can just merge it as-is.

ap-justin:
changes reflected in #2053

PR #1972, opened 2023-03-22
Submit polygon AST to AWS

ap-justin:
Ticket(s):
a component of PR
* #1966 

### In this PR
* connect metamask to `Mumbai`
* got to `/register` (make sure `REACT_APP_APP_TYPE=AST`) and complete registration
* after submitting `tx`, verify that 
correct payload is being sent to `AWS` ( aws still has some work to do to process this submission )
image

#### add `get_transactionReceipt` - this is equivalent to `terra` pollTx before ( also what is also done when waiting for cosmos tx to be finalized using `lcd`). Failure to get this receipt (timeout) means aws is not submitted to `AWS`.  A more robust approach  is to let `AWS` listen for `event EndowmentCreated` directly - though `web-app` still has to get this log to show appropriate redirect page with content like:
```
You can now manage your endowment at `admin/{retrieved_id}`
```

####  extract event log from `receipt`. This could be easily done just by using `@ethersproject/contracts` but would introduce different user flow where the current `juno` user flow is 
` .....prior steps ---> sendTx ---> extractLogs --> show success message + admin(retrieve_log) link`

to elaborate: say in a file `Account.ts`
```ts
const account = new Account(abi)
account.on("EndowmentCreated",(id) => {
   navigate(`admin/${id}`)
})
```
and in the `sendTx` function, ff user flow already happend
```
SIGN --> SUBMIT --> ?? what to show to user while `event` is not happening yet
```
At this point user can navigate away from the launchpad page and it would be dubious
that `Accounts.ts` suddenly redirect user to `/success page`

On the other hand, We could show some alerts  like
```
[  ⋙ endowment being created...]
[ ✓ endowment created -- checkout admin/(id) ] 
```
which would be different from that of existing prompts/succes pages

another reason why `@ethersproject/contracts` is not used here is that it is not used in tx flows in
* #1966 

with reason to be explained at that PR 

### Refactors
* `attribute?:string` only on tx `content type:"cosmos"`
* simplify abi to what we need, 
* ERC20, balanceOf/transfer : no need for named fragment, 
* Accounts, createEndowment: no need for named, event: no need for named since we only care for `endowId` at `idx` 0 and not 
 `newAIF`

-

## UI changes for review


src/types/tx.ts
```diff
@@ -31,9 +31,9 @@ export type TxResult = TxError | TxSuccess;
 
 // //////////// ESTIMATE TX ////////////
 export type TxContent =
-  | { type: "cosmos"; val: EncodeObject[] }
+  | { type: "cosmos"; val: EncodeObject[]; attribute?: string }
   | { type: "terra"; val: Msg[]; wallet: ConnectedWallet }
-  | { type: "evm"; val: SimulContractTx | SimulSendNativeTx };
+  | { type: "evm"; val: SimulTx; log?: LogProcessor };
```
misicnenad, line 36: Passing a "log processor" adds more flexibility than passing just an `attribute` as we do for Cosmos, as it allows us to process logs as we see fit per component. 

I know the current setup "works fine at the moment", but it forces us to:
- pass either `attribute` or `log` parameter to `TxContent` when we need some param read from the logs, the purpose of which is not immediately obvious unless you know how it will be used in the subsequent chain of functions (to return `TxSuccess.attrValue`)
- have different ways of populating `TxSuccess.attrValue` (for `cosmos` we pass `attribute` and it finds it for us, while for `evm` we pass the log filter function that the `sendTx` has to call)
- return `TxSuccess.attrValue`, the purpose of which is not immediately obvious (unless you know how it was populated)
- might not be a need for this atm, but this setup would stop us from being able to read multiple attributes from the logs

What do you think about just returning `TxSuccess.logs` instead of returning `TxSuccess.attrValue` and processing those logs as we see fit in specific functions/components? `.

src/types/tx.ts
```diff
@@ -31,9 +31,9 @@ export type TxResult = TxError | TxSuccess;
 
 // //////////// ESTIMATE TX ////////////
 export type TxContent =
-  | { type: "cosmos"; val: EncodeObject[] }
+  | { type: "cosmos"; val: EncodeObject[]; attribute?: string }
```
misicnenad, line 34: I know the current setup "works fine at the moment", but it forces us to:
- pass either `attribute` or `log` parameter to `TxContent` when we need some param read from the logs, the purpose of which is not immediately obvious unless you know how it will be used in the subsequent chain of functions (to return `TxSuccess.attrValue`)
- have different ways of populating `TxSuccess.attrValue` (for `cosmos` we pass `attribute` and it finds it for us, while for `evm` we pass the log filter function that the `sendTx` has to call)
- return `TxSuccess.attrValue`, the purpose of which is not immediately obvious (unless you know how it was populated)
- might not be a need for this atm, but this setup would stop us from being able to read multiple attributes from the logs

What do you think about just returning `TxSuccess.logs` instead of returning `TxSuccess.attrValue` and processing those logs as we see fit in specific functions/components? `.

misicnenad (approved):

src/types/tx.ts
ap-justin: > pass either attribute or log parameter to TxContent when we need to read some param from the logs,

You are right it's confusing. I'm still looking for ways to make it more comprehensible. The reason `attribute` and `fn log` is passed in `TxContent` rather than `sendTx` is that in `tx creation` step, this is where we distinguish a tx type from another along with how those differences manifest. E.g. for cosmos (`attribute`) and for polygon (`log`)


> have different ways of populating TxSuccess.attrValue (for cosmos we pass attribute and it finds it for us, while for evm we pass the log filter function that the sendTx has to call)

the reason they're different was due to the difference on how those logs can be parsed. For `cosmos` it's a simple finding `value` for a `key` (`attribute` which is human readable). which is typically just primitive `string | number`. For polygon or `evm` in general. `keys` are encoded based on `Event signature` and values are also encoded and needs to be `decoded` depending on the `Event signature`. Moreover, the `encoded` value isn't just some `string | number` an could be complex depending on `Event signature` 
sample
```
"event SwapToken(uint256,uint8,uint256,address,address,uint256)",
"event EndowmentSettingUpdated(uint256,string)",
```

SovereignAndrey (approved):

Issue #1973, opened 2023-03-23
urgent marketplace editing issues.


SovereignAndrey:
CU #: 862jg6t1j - https://app.clickup.com/t/862jg6t1j
Created on: Mar 23 2023 
Original Creator: Tim Stirrup

## Description
<p>Please see this message for details <a href="https://discord.com/channels/834976399228010527/857111094019883010/1088147435945197710" target="_blank">https:&#x2F;&#x2F;discord.com&#x2F;channels&#x2F;834976399228010527&#x2F;857111094019883010&#x2F;1088147435945197710</a><br/><br/>and this loom which illustrates both <br/><br/>Unmanaged custom blot!<br/><br/>and here&#x27;s another issue - not being able to change the HQ details - <br/><br/>Unmanaged custom blot!</p>

misicnenad:
@SovereignAndrey just a note, the "Minified React error #321..." is something that started occurring due to some newly added/updated pkg (not sure which), but seems to not be affecting the webapp.

misicnenad:
@SovereignAndrey it is possible to fix the TikTok URL issue, but this is not the first time we'd had to deal with a URL validation edge case. See [this comment on the related CU task](https://app.clickup.com/t/862jg6t1j?comment=90040006961312) for more context and details.

misicnenad:
@SovereignAndrey about the HQ update bug, seems to be some data format mismatch between webapp and AWS, currently waiting for more info from @jp-mariano once he investigates on AWS side if something might be wrong there or if the data format should be updated on the webapp.

jp-mariano:
> @SovereignAndrey about the HQ update bug, seems to be some data format mismatch between webapp and AWS, currently waiting for more info from @jp-mariano once he investigates on AWS side if something might be wrong there or if the data format should be updated on the webapp.

Regarding the HQ update bug, the updatooor Lambda is still expecting the old format of an `hq` object instead of the new `hq_country` field. A hot fix is already implemented and this should be ok now. 👌🏾

PR #1979, opened 2023-03-23
Fix img size bug

misicnenad:
Ticket(s):
- https://github.com/AngelProtocolFinance/angelprotocol-web-app/issues/1973

- the size of the uploaded file gets automatically increased once cropped due to the way `cropperjs` works, see official docs section ["Known Issues"](https://github.com/fengyuanchen/cropperjs#known-issues)
- additionally, all image types were being converted to PNG, no matter the original type
- to address this, all that was necessary was to set the cropped file's type to be the same as the original's (see ["Known Issues"](https://github.com/fengyuanchen/cropperjs#known-issues))
- the things is that this results in a cropped img size smaller from the original, which is to be expected
- problem is that occasionally, especially for PNGs, if the image was previously compressed, the canvas API used by `cropperjs` actually *removes this compression*, thus increasing the image size (see [this answer](https://github.com/fengyuanchen/cropperjs/issues/538#issuecomment-508450556) and [this answer](https://github.com/fengyuanchen/cropperjs/issues/804) on related GH issues)
- even when the image size is reduced as expected, the user might still be confused to find out that their image with a size larger than 1 MB was uploaded with no problems
- to address both of these issues, updated the existing tooltip from:
`Valid types are: PDF, JPG, PNG and WEBP. File should be less than 1MB.`
to:
`Valid types are: PDF, JPG, PNG and WEBP. File should be less than 1MB. Cropped file size can be different from the original's. Current cropped file size: ${file.size / BYTES_IN_MB}.`

image

- connect Charity 1 wallet on Juno
- go to http://localhost:4200/admin/1/edit-profile
- scroll to "Banner image of your organization"
- click "Upload" btn
- upload the image attached in this Discord message https://discord.com/channels/834976399228010527/857111094019883010/1088147435945197710
- move the cropping section to left- and right-most edges, like in image below:
image
- crop the image
- verify the cropped image size is no longer ~8 times larger (difference should be ~20Kb)
- verify the cropped image type is same as the original (PNG)
- repeat the same process with the version of that same image which was updated to be JPG ([link](https://user-images.githubusercontent.com/19427053/227178020-1b69909c-0286-4cf1-9534-88ba7b2917ae.jpg))
- verify the cropped image size is no longer ~8 times larger (should actually be smaller than original)
- verify the cropped image type is same as the original (JPG)

## UI changes for review

New tooltip (note: updated "file" to read "image"):
image

With error message (forced by temporarily reducing valid img size) (note: updated "file" to read "image"):
image

### DIFF
--- a/src/components/ImgEditor/ImgEditor.tsx
+++ b/src/components/ImgEditor/ImgEditor.tsx
@@ -4,6 +4,7 @@ import { useDropzone } from "react-dropzone";
 import { FieldValues, useFormContext } from "react-hook-form";
 import { ImgLink, Props } from "./types";
 import Icon from "components/Icon";
+import { humanize } from "helpers";
 import useImgEditor from "./useImgEditor";
 
 type Key = keyof ImgLink;
@@ -112,8 +113,13 @@ export default function ImgEditor<T extends FieldValues, K extends keyof T>(
       <p className="text-xs text-gray-d1 dark:text-gray mt-2">
         <span>
           Valid types are: PDF, JPG, PNG and WEBP. File should be less than 1MB.
-          Cropped file size can be different from the original's.{" "}
-          {file ? `Current cropped file size: ${file.size / BYTES_IN_MB}.` : ""}
+          Cropped file size can be different from the original's.
+          <br />
+          {file
+            ? `Current cropped file size: ${humanize(
+                file.size / BYTES_IN_MB
+              )}MB.`
+            : ""}
         </span>{" "}
         <ErrorMessage
           errors={errors}
--- a/src/components/ImgEditor/ImgEditor.tsx
+++ b/src/components/ImgEditor/ImgEditor.tsx
@@ -116,7 +116,7 @@ export default function ImgEditor<T extends FieldValues, K extends keyof T>(
           Cropped file size can be different from the original's.
           <br />
           {file
-            ? `Current cropped file size: ${humanize(
+            ? `Current cropped image size: ${humanize(
                 file.size / BYTES_IN_MB
               )}MB.`
             : ""}
--- a/src/components/ImgEditor/ImgEditor.tsx
+++ b/src/components/ImgEditor/ImgEditor.tsx
@@ -112,11 +112,10 @@ export default function ImgEditor<T extends FieldValues, K extends keyof T>(
       </div>
       <p className="text-xs text-gray-d1 dark:text-gray mt-2">
         <span>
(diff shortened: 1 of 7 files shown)


src/components/ImgEditor/ImgEditor.tsx
```diff
@@ -4,11 +4,14 @@ import { useDropzone } from "react-dropzone";
 import { FieldValues, useFormContext } from "react-hook-form";
 import { ImgLink, Props } from "./types";
 import Icon from "components/Icon";
+import { humanize } from "helpers";
 import useImgEditor from "./useImgEditor";
 
 type Key = keyof ImgLink;
 const fileKey: Key = "file";
 
+const BYTES_IN_MB = 1e6;
```
misicnenad, line 13: since this conversion is unlikely to change in our lifetimes, no need to have it in shared constants.

abbasali1994 (approved):

ap-justin (dismissed):

src/components/ImgEditor/ImgEditor.tsx
```diff
@@ -109,6 +113,13 @@ export default function ImgEditor<T extends FieldValues, K extends keyof T>(
       <p className="text-xs text-gray-d1 dark:text-gray mt-2">
         <span>
           Valid types are: PDF, JPG, PNG and WEBP. File should be less than 1MB.
```
SovereignAndrey, line 115: Remove PDF from acceptable file types. Slight tweaks to size wording. 
```suggestion
          Valid types are: JPG, JPEG, PNG and WEBP. Original uploaded image should be less than 1MB in size.
```

src/components/ImgEditor/ImgEditor.tsx
```diff
@@ -109,6 +113,13 @@ export default function ImgEditor<T extends FieldValues, K extends keyof T>(
...
+          Cropped file size can be different from the original's.
```
SovereignAndrey, line 116: Not really needed.
```suggestion
```

src/components/ImgEditor/ImgEditor.tsx
```diff
@@ -109,6 +113,13 @@ export default function ImgEditor<T extends FieldValues, K extends keyof T>(
...
+          <br />
+          {file
+            ? `Current cropped image size: ${humanize(
```
SovereignAndrey, line 119:
```suggestion
            ? `Current (cropped) image size: ${humanize(
```

SovereignAndrey (approved):
Left some minor tweaks/suggestions. Will just batch commit those and full send. Nice work!

PR #1981, opened 2023-03-23
WIP: Contract query reworks

ap-justin:
Ticket(s):

* move individual `junoApi.endpoints` into single `useQueryContract` hook
```ts
//definition
export function useQueryContract<T extends QT>(
  type: T,
  options: QueryOptions<T>,
  config?: SWRConfiguration
): SWRResponse<ReturnType<Q[T]["transform"]>> {
  return useSWR([type, options], (args) => queryContract(...args), config);
  
 //sample usage
  const { data: fundList = [] } = useQueryContract("index-fund.funds", {});
  
```
* no need to input contract address when contract is hardcoded. This possible because of query id convention `{contract}.{query}`
```ts
//before
const endowment = await queryContract(
          "accEndowment",
          contracts.accounts,
          { id: numId }
        );
        
 //now
  const endowment = await queryContract("accounts.endowment", {
        id: numId,
      });
```

-

## UI changes for review


ap-justin:
will try to implement these abstraction with rtk-query as it is not cumbersome to do so now that only polygon queries has to be supported

PR #1988, opened 2023-03-24
Verify only uploaded (pre-crop) image size

misicnenad:
Ticket(s):
- https://github.com/AngelProtocolFinance/angelprotocol-web-app/issues/1973

The idea is to only validate the original image that the user uploaded to `ImgEditor` and not the cropped/processed image (for explanation see [this comment](https://app.clickup.com/t/862jg6t1j?comment=90040006955728&threadedComment=90040006971841)).
To do that, we store the original (precropped) image and only focus on its validation (type, size etc.). If original image is valid, the cropped image will be valid for sure.

- connect Charity 1 wallet
- go to Edit Profile page http://localhost:4200/admin/1/edit-profile
- verify new tooltip appears under image upload components
- verify image upload works as before
- verify user is only ever able to upload images that have a size of `<= 1MB` (no matter the cropped image size)

### DIFF
--- a/src/components/ImgEditor/ImgEditor.tsx
+++ b/src/components/ImgEditor/ImgEditor.tsx
@@ -4,19 +4,18 @@ import { useDropzone } from "react-dropzone";
 import { FieldValues, useFormContext } from "react-hook-form";
 import { ImgLink, Props } from "./types";
 import Icon from "components/Icon";
-import { humanize } from "helpers";
 import useImgEditor from "./useImgEditor";
 
 type Key = keyof ImgLink;
 const fileKey: Key = "file";
-
-const BYTES_IN_MB = 1e6;
+const precropSizeKey: Key = "precropSize";
 
 export default function ImgEditor<T extends FieldValues, K extends keyof T>(
   props: Props<T, K>
 ) {
   const { name, classes } = props;
   const filePath: any = `${String(name)}.${fileKey}`;
+  const precropSizePath: any = `${String(name)}.${precropSizeKey}`;
 
   const {
     formState: { errors, isSubmitting },
@@ -25,7 +24,6 @@ export default function ImgEditor<T extends FieldValues, K extends keyof T>(
   const {
     onDrop,
     handleOpenCropper,
-    file,
     isInitial,
     noneUploaded,
     handleReset,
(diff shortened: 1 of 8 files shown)


src/components/ImgEditor/ImgEditor.tsx
```diff
@@ -112,18 +112,22 @@ export default function ImgEditor<T extends FieldValues, K extends keyof T>(
...
-          Valid types are: JPG, JPEG, PNG and WEBP. Original uploaded image
-          should be less than 1MB in size.
-          <br />
-          {file
-            ? `Current (cropped) image size: ${humanize(
-                file.size / BYTES_IN_MB
-              )}MB.`
-            : ""}
+          Valid types are: JPG, JPEG, PNG and WEBP.{" "}
+          {maxSize ? (
```
misicnenad, line 116: No need to show the "max size" tooltip if there's no maximum size set.

src/components/ImgEditor/ImgEditor.tsx
```diff
@@ -112,18 +112,22 @@ export default function ImgEditor<T extends FieldValues, K extends keyof T>(
...
+            <>
+              Image should be less than {maxSize / BYTES_IN_MB}MB in size.
+              <br />
+              {imgSize
```
misicnenad, line 120: Current image size is only relevant if there is a maximum image size. 
@SovereignAndrey we can show this nevertheless of course, this is just a proposal.

src/components/ImgEditor/types.ts
```diff
@@ -13,4 +17,8 @@ export type Props<T extends FieldValues, K extends keyof T> = {
   accept: string[];
   classes?: Classes;
   aspect: [number, number];
+  /**
+   * Maximum image size in bytes
```
misicnenad, line 21: Adding JSDoc comments can be very helpful during development as it provided useful info in code editors (see image for on-hover info in VS Code).
image

We should make it a habit to add them, at least for fields/params whose purpose is not immediately obvious.

src/components/ImgEditor/types.ts
```diff
@@ -13,4 +17,8 @@ export type Props<T extends FieldValues, K extends keyof T> = {
...
+   */
+  maxSize?: number;
```
misicnenad, line 23: Adding `maxSize` prop makes `ImgEditor` tooltip at the bottom reflect the actual max image size (currently was hardcoded)

src/components/ImgEditor/ImgEditor.tsx
SovereignAndrey: I can't imaging a world where we don't have _some form_  of upper limit on file size uploads. Even if large, it'd still need to exist.

SovereignAndrey (approved):

ap-justin (approved):

Issue #1991, opened 2023-03-27
religious organization SDG issue


SovereignAndrey:
CU #: 863gaaxce - https://app.clickup.com/t/863gaaxce
Created on: Mar 23 2023 
Original Creator: Tim Stirrup

## Description
<p>PLease see the Loom. <br/><br/>Religious organizations have to add an SDG at the moment, and they probably shouldn&#x27;t. So there needs to be a &#x27;zero&#x27; option for those. <br/><br/>also, weren&#x27;t we going to restrict SDGs allowed? as ticking all means you can game the search&#x2F;filter system bu always appearing. <br/><br/>all on here - <br/>Unmanaged custom blot!</p>

## COS
- [ ] Restrict number of SDGs to ~~4~~ (8) on both the Registration portal form & on the Edit Profile form
- [ ] Remove `select-all` button

PR #1997, opened 2023-03-27
Fix: sdg limits and select-all + UI fixes

ap-justin:
Ticket(s): 
* #1991 

1. disable `select-all` as default as most of our `multiple` selectors doesn't make sense for all options to be selected
* Activity countries
* EditProfile/Registration - SDGs
* KYC - US states

image

on the other hand allow `select-all` for widget currencies

2. limit sdgs (8) - already implemented

-

## UI changes for review

### DIFF
--- a/src/components/Selector.tsx
+++ b/src/components/Selector.tsx
@@ -32,6 +32,7 @@ interface Props<
   placeholder?: string;
   options: OptionType<V>[];
   disabled?: true;
+  allSelectable?: M extends true ? boolean : never;
   classes?: Classes;
   children?: (selected: VarOption<M, V>) => ReactNode;
 }
@@ -52,6 +53,7 @@ export function Selector<
   children,
   classes,
   multiple,
+  allSelectable = false as any,
 }: Props<T, K, ValueType, Multiple>) {
   const { container = "", button = "" } = classes || {};
   const {
@@ -112,14 +114,18 @@ export function Selector<
         </Listbox.Button>
         <Listbox.Options className="rounded-sm text-sm border border-prim absolute top-full mt-2 z-10 bg-gray-l5 dark:bg-blue-d6 w-full max-h-[10rem] overflow-y-auto scroller">
           {multiple && (
-            <div className="flex justify-between p-4">
+            <div className="flex p-4">
               {isAllSelected ? (
                 <Action onClick={() => onChange([])}>Deselect All</Action>
               ) : (
-                <Action onClick={() => onChange(options)}>Select All</Action>
+                allSelectable && (
+                  <Action onClick={() => onChange(options)}>Select All</Action>
+                )
               )}
 
-              <Action onClick={() => resetField(name)}>Reset</Action>
+              <Action onClick={() => resetField(name)} classes="ml-auto">
+                Reset
+              </Action>
             </div>
           )}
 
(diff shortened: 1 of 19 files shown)


SovereignAndrey:
@ap-justin We cannot have a charity endowment with `None` for sdgs. They all must have at least one. We'll need to either default all religious organisations to a single SDG(which I don't think is the right way to handle it) or for now we force them to pick at least one they feel they match with.

SovereignAndrey:
@ap-justin Tim confirmed the following: 
- [ ] Limit the selected SDG options to 8 items, just like we have in place for the display limit on the Marketplace Cards
- [ ] Remove the select all for the SDG dropdown as it doesn't make sense to have in context of limits

src/pages/Registration/Steps/Documentation/Form/index.tsx
```diff
@@ -77,7 +78,7 @@ export default function Form() {
         fieldName="hqCountry"
         placeholder="Select a country"
         classes={{
-          container: "px-4 bg-orange-l6 dark:bg-blue-d7",
```
ap-justin, line 80: fixes bug (inconsistent bgg)
image

src/pages/Admin/Charity/EditProfile/Form.tsx
```diff
@@ -127,6 +128,7 @@ export default function Form() {
             label: option.label,
             value: option.value,
           }))}
+          classes={{ button: "field-input-admin" }}
```
ap-justin, line 131: fixes bug (endow designation selector inconsistent)

image

src/pages/Admin/Charity/EditProfile/Form.tsx
```diff
@@ -144,7 +146,6 @@ export default function Form() {
         <ActivityCountries<FV, "active_in_countries">
           name="active_in_countries"
           classes={{
-            container: "bg-white dark:bg-blue-d6 px-4 py-3",
```
ap-justin, line 147: fixes
image

src/components/Selector.tsx
```diff
@@ -52,6 +53,7 @@ export function Selector<
   children,
   classes,
   multiple,
+  allSelectable = false as any,
```
misicnenad, line 56: nit: no need for default value assignment, will be `undefined` by default which will make the "Select All" action hidden.

src/types/aws/ap/registration.ts
```diff
@@ -78,7 +79,7 @@ export type TDocumentation = {
   KycDonorsOnly: boolean;
   CashEligible: boolean;
   HqCountry: string;
-  EndowDesignation: string;
+  EndowDesignation: EndowDesignation | "";
```
misicnenad, line 82: nit: there are a couple of places where either `EndowDesignation` or an empty string `""` is allowed as a value. It could make sense to add the empty string `""` to `EndowDesignation` and treat it as some kind of a "default" value.

misicnenad (approved):

src/types/aws/ap/registration.ts
SovereignAndrey: > nit: there are a couple of places where either `EndowDesignation` or an empty string `""` is allowed as a value. It could make sense to add the empty string `""` to `EndowDesignation` and treat it as some kind of a "default" value.
> 
> @SovereignAndrey can there be endowments for whom the `endowment designation` is empty?

@misicnenad @ap-justin All Charity Endowments would need to have some form of Endowment Designation. The AST Endowments would not have this today, though I would suspect they could get their own set of sub-designation at some point in the future.

SovereignAndrey (approved):

Issue #1733, opened 2023-02-09
Fiat donation flow designs match w/ Figma


SovereignAndrey:
**CU Ticket:** https://app.clickup.com/t/865bm5pu2
**Assigned To:** 
**Version:** v1.8

abbasali1994:
Checked kado docs on 9 March and no updates have been shared on release of "post order status" feature as of now
Since this feature is a necessary element for our fiat flow design keeping this ticket in backlog for now,
also will communicate with kado team via mail to inquire on when they plan to release this feature

PR #1999, opened 2023-03-27
Meld FIat Donation integration

abbasali1994:
Ticket(s):
- [Fiat donation flow designs match w/ Figma#1733](https://github.com/AngelProtocolFinance/angelprotocol-web-app/issues/1733)

-

## UI changes for review


SovereignAndrey (changes requested):
As mentioned in Discord, for starters, please:
- [x] point to RC-v1.8
- [x] move the API key into the `.env` and accessed by `process.env.REACT_APP_MELD_ACCESS_TOKEN` vs storing in the constants

I've setup an AWS app for testing Meld flows: https://meld-integration.dpspevs7tj1ov.amplifyapp.com/
It has the `REACT_APP_MELD_ACCESS_TOKEN` set in it's env vars and build file.

SovereignAndrey (changes requested):
A bunch of things needed still. Marking this PR as draft. Please mark as ready once these items are done. Thanks!
- [ ] There is no new fiat donation thank you page for the redirect. You still need to add that. 
- [ ] We'll need to properly send the country code as well. Noticed that it's hard-coded to "US" in your code. 
- [ ] Please coordinate with @jp-mariano about changes needed with AWS handling the crypto widget submission and will get the response from AWS now. 
    - [ ] no need to do uuid on web app end, since AWS can handle (remove that package)
    - [ ] only pass the amount, currency, country code, donation Endow ID and split details to AWS (ask Paolo for exact list needed)
    - [ ] Remove the env var and related MELD API items. No longer needed, as this will live in AWS
    - [ ] Send the user to the returned crypto widget. No need for modal. The redirect button will point them back to the Fiat Donation Thank you page on our dapp.

SovereignAndrey (changes requested):
@abbasali1994 I've pushed some commits up to start moving this PR branch towards a ready state, but there a lot that needs to be done. 

### Things I added:
- Moved the FiatDonateThanks component into a page
- Added some AST copy toggling for donate vs contribute where I noticed it
- Added a Link to the Wallet connection needed notice in Steps, so that user's without wallets would know they can still donate in fiat. Points to a placeholder DonateFiat endpoint route.

### Things you need to do (before leaving on Firday):
- [x] Fix the fiat calculations. It seems it is not accounting for the fiat to be donated (only shows gas amount in total)
costs_wrong
- [x] Donation without crypto wallet isn't possible due to the wallet check error thrown. This is not acceptable UX since the point of fiat donations is that users don't need a web3 wallet. Please create the `DonateFiat` endpoint (placeholder atm), so that it's basically the existing `Donate` endpoint without wallet checks and such (ie. Donation without wallet & only Fiat currency options). Feel free to Copy-Paste Donate and strip out what you don't need for now as a starting point. Existing Donate endpoint can keep the Fiat options it has.

src/components/donationFiat/Steps/FiatSubmit/Form/index.tsx
```diff
@@ -0,0 +1,150 @@
...
+  ) : (
+    <form
+      onSubmit={handleSubmit(submit)}
+      className="grid gap-4 rounded-md w-full"
+      autoComplete="off"
+    >
+      <h4 className="font-bold text-sm">Enter your payment details:</h4>
+      <div className="flex-col">
+        <Label className="mb-2" htmlFor="country">
+          Country of Origin *
```
SovereignAndrey, line 69: nit: make this more understandable/relatable for users.
```suggestion
          Country of Residence *
```

src/pages/DonateFiatThanks/index.tsx
```diff
@@ -0,0 +1,35 @@
+import { Link } from "react-router-dom";
+import Icon from "components/Icon";
+import { IS_AST } from "constants/env";
+import { appRoutes } from "constants/routes";
+
+export default function DonateFiatThanks() {
+  return (
+    <div className="justify-self-center display-block m-auto max-w-[35rem] py-8 sm:py-20 scroll-mt-6">
+      <Icon type="CheckCircle" size={96} className="text-green mb-4 ml-10" />
```
SovereignAndrey, line 9: Can we center this icon over the text please? I think this may have been my bad from when I moved it into a page and reduced some of the wrappers of divs around the contents. Everything else on the page looks OK. 

image

SovereignAndrey (changes requested):
Some little fixes left and I think we're good. :)

src/pages/DonateFiatThanks/index.tsx
abbasali1994: Done

src/components/donationFiat/Steps/FiatSubmit/Form/index.tsx
abbasali1994: Done

SovereignAndrey (approved):

Issue #1927, opened 2023-03-16
Replace `img` uses with `/components/Image` component


SovereignAndrey:
Continuing work started in https://github.com/AngelProtocolFinance/angelprotocol-web-app/pull/1849

All `<img>` uses should be replaced with the newly created `components/Image` component, as it supports loading logic+UI and placeholder image displays.

PR #2013, opened 2023-03-28
Replace `img` uses with `/components/Image` component

misicnenad:
Ticket(s):
- https://github.com/AngelProtocolFinance/angelprotocol-web-app/issues/1927

- updated `Image` to fix some issues that became apparent during testing. Notable examples of these are:
   - showing `ImagePlaceholder` instead of `ContentLoader` when `props.src` was still being loaded (as indicated by `props.isSrcLoading`).
   - reduced thickness of profile logo border to align with Figma (border was applied twice in `/components/Image`, but is now applied only once)
    before:
    image
    after:
    image
- replaced `ImageWrapper` with `Image`.
- flattened `ImageProps` to make it as similar as possible as plain `img` props.

- verify all images (or as many as you can check) appear as before


src/components/Image/Image.tsx
```diff
@@ -1,49 +1,73 @@
...
+const Image = React.forwardRef<HTMLImageElement, ImageProps>(
+  ({ className, alt = "", ...props }, forwardRef) => {
+    const ref = useRef<HTMLImageElement>(null);
+    const [isLoading, setLoading] = useState(!!props.src || props.isSrcLoading);
+    const [isError, setError] = useState(false);
 
-export default function Image({ img, className }: Props) {
-  const ref = useRef<HTMLImageElement>(null);
-  const [isLoading, setLoading] = useState(!!img?.src || img?.isSrcLoading);
+    useImperativeHandle<HTMLImageElement | null, HTMLImageElement | null>(
```
misicnenad, line 23: One of the rarely used hooks in React, but very useful in this case - `ImgCropper` needs to [forward a ref to the nested `img` component inside `Image`](https://github.com/AngelProtocolFinance/angelprotocol-web-app/blob/img-to-image/src/components/ImgEditor/ImgCropper.tsx#L60), but `Image` itself needs to have a `ref` to that same component. The solution is to tell any parent component to use this internal `ref` for any and all purposes, which is accomplished with this hook.

src/components/Image/Image.tsx
```diff
@@ -60,6 +84,8 @@ function WithLink({
       {children}
     </ExtLink>
   ) : (
-    <div className={className}>{children}</div>
```
misicnenad, line 63: Wrapping in `div` was causing `ImgCropper` to display a very small image on screen when using this component. Rendering `img` directly like this enables it (and by extension `ImgCropper`) to render a bigger image (up to [defined max size](https://github.com/AngelProtocolFinance/angelprotocol-web-app/blob/img-to-image/src/components/ImgEditor/ImgCropper.tsx#L50)).

src/components/Image/Image.tsx
```diff
@@ -1,49 +1,73 @@
...
+           */
+          <WithLink
+            className={`${className} ${shouldLoad ? "hidden" : ""}`}
+            href={props.href}
+            title={props.title}
+          >
+            <img
+              ref={ref}
+              src={props.src}
+              className={`object-contain ${
```
misicnenad, line 56: Most of the components that call `Image` use `object-contain` as one of their classes, so we can put it as one of the default classes. All components that need to use some other [object-fit](https://tailwindcss.com/docs/object-fit) style need just pass it and it will overwrite this class name.

src/components/Image/Image.tsx
```diff
@@ -1,49 +1,73 @@
...
+                shouldLoad ? "hidden" : ""
+              } ${className}`}
+              alt={props.alt || ""}
+              loading={props.loading}
+              onLoad={() => setLoading(false)}
+              onError={(e) =>
+                props.onError ? props.onError(e) : setError(true)
```
misicnenad, line 62: @SovereignAndrey @ap-justin this logic makes little sense, but it has a reasoning behind it.
The usual way would be to treat `props.onError` as a hook into the error event that is run in addition to any internal error handling, like:
```ts
onError={(e) => {
  if (props.onError) props.onError(e);
  // other internal error logic
  // ...
}}
```
But because I wanted to maintain existing behavior in just one component - `Card`, had to implement `onError` here to be used as a direct replacement for any internal error handling (either use `props.onError` OR use internal error logic).

The existing behavior in question is that `Card` wants to just [append a blue background when an error happens with no placeholder image](https://github.com/AngelProtocolFinance/angelprotocol-web-app/blob/img-to-image/src/pages/Marketplace/Cards/Card.tsx#L39). If we could make it so that the `onError` behavior here is updated to just show the [ImagePlaceholder](https://github.com/AngelProtocolFinance/angelprotocol-web-app/blob/img-to-image/src/components/Image/ImagePlaceholder.tsx) on error, i.e. use the [default `Image.onError` behavior](https://github.com/AngelProtocolFinance/angelprotocol-web-app/blob/img-to-image/src/components/Image/Image.tsx#L27), it would enable us to update `props.onError` to be treated as a hook into the error event (described above).

This would also enable us to remove `onError` in other places where it's used and update them to use some a fallback image. Currently kept their logic as-is as it's less changes to make (see [Leaderboard > Table > Row](https://github.com/AngelProtocolFinance/angelprotocol-web-app/blob/img-to-image/src/pages/Leaderboard/Table/Row/index.tsx#L20) and [Admin > Proposal > DiffTable](https://github.com/AngelProtocolFinance/angelprotocol-web-app/blob/img-to-image/src/pages/Admin/Proposal/Content/Preview/DiffTable.tsx#L54)).

src/components/Image/Image.tsx
```diff
@@ -1,49 +1,73 @@
...
+  useRef,
+  useState,
+} from "react";
 import ContentLoader from "components/ContentLoader";
 import ExtLink from "components/ExtLink";
 import ImagePlaceholder from "./ImagePlaceholder";
 
-export type ImageProps = {
-  src?: string;
+export type ImageProps = React.ImgHTMLAttributes<HTMLImageElement> & {
```
misicnenad, line 11: Merged `ImageProps` and `Props` as it makes using `Image` component much more similar to using plain `img`, making it more intuitive.

src/components/Image/Image.tsx
```diff
@@ -1,49 +1,87 @@
...
-            className={`${className} object-contain w-full h-full`}
-            alt=""
-            onLoad={() => setLoading(false)}
+            src={props.src}
+            className={`object-contain ${
+              isLoading ? "hidden" : ""
+            } ${className}`}
+            alt={props.alt || ""}
+            loading={props.loading}
+            onError={(e) => (props.onError ? props.onError(e) : setError(true))}
```
misicnenad, line 78: @SovereignAndrey @ap-justin this logic makes little sense, but it has a reasoning behind it. The usual way would be to treat `props.onError` as a hook into the error event that is run in addition to any internal error handling, like:

```ts
onError={(e) => {
  if (props.onError) props.onError(e);
  // other internal error logic
  // ...
}}
```

But because I wanted to maintain existing behavior in just one component - `Card`, had to implement `onError` here to be used as a direct replacement for any internal error handling (either use `props.onError` OR use internal error logic).

The existing behavior in question is that `Card` wants to just [append a blue background when an error happens with no placeholder image](https://github.com/AngelProtocolFinance/angelprotocol-web-app/blob/img-to-image/src/pages/Marketplace/Cards/Card.tsx#L39). If we could make it so that the `onError` behavior here is updated to just show the [ImagePlaceholder](https://github.com/AngelProtocolFinance/angelprotocol-web-app/blob/img-to-image/src/components/Image/ImagePlaceholder.tsx) on error, i.e. use the [default `Image.onError` behavior](https://github.com/AngelProtocolFinance/angelprotocol-web-app/blob/img-to-image/src/components/Image/Image.tsx#L27), it would enable us to update `props.onError` to be treated as a hook into the error event (described above).

This would also enable us to remove `onError` in other places where it's used and update them to use some a fallback image. Currently kept their logic as-is as it's less changes to make (see [Leaderboard > Table > Row](https://github.com/AngelProtocolFinance/angelprotocol-web-app/blob/img-to-image/src/pages/Leaderboard/Table/Row/index.tsx#L20) and [Admin > Proposal > DiffTable](https://github.com/AngelProtocolFinance/angelprotocol-web-app/blob/img-to-image/src/pages/Admin/Proposal/Content/Preview/DiffTable.tsx#L54)).

ap-justin (approved):

src/components/Image/Image.tsx
misicnenad: @ap-justin 
> I have to assume you took it as a negative thing.

No no, just wanted to clarification on what from that comment you "thumbed up". I guess you obviously agreed to the whole comment. Will create a task to update `Card > Image.onError` behaviour.

Issue #1929, opened 2023-03-16
Add red border to fields that have an error attached to them


SovereignAndrey:
It is good UI to add red borders to input fields that have errors attached to them (see image below).

![Image](https://user-images.githubusercontent.com/19427053/223148564-4ea6ac81-c564-4bd0-a59a-945ff44494cd.png)

Use the below code snippet as template for adding this style to the field border (the example assumes the component is using `react-hook-form`, but the styles can used regardless of the pkg/logic used in the condition part).
```ts
type FormValues = { field: string };
...
const { formState: { errors } } = useFormContext<FormValues>();
...
${errors?.field?.message ? "border-2 border-red dark:border-red-l2" : "border border-prim" }
```

PR #2020, opened 2023-03-29
Field error border addition

abbasali1994:
Ticket(s):
- [Add red border to fields that have an error attached to them#1929](https://github.com/AngelProtocolFinance/angelprotocol-web-app/issues/1929)

-

## UI changes for review


abbasali1994:
Please mention any field components that i might have missed which require these borders to be included

src/components/CountrySelector/index.tsx
```diff
@@ -68,7 +69,7 @@ export default function CountrySelector<
       as="div"
       className={`relative items-center grid grid-cols-[auto_auto_1fr] w-full field-container ${
         props.classes?.container || ""
-      }`}
+      } ${getErrorClasses(errors, `${props.fieldName}.name`)}`}
```
ap-justin, line 72: @abbasali1994
Another way to attach these error classes would be to add aria attribute

```ts
const isError = !!errors[name /** usually provided*/]?.message
<Combobox aria-invalid={isError}/>
```
and in `index.css` could put classes in `field-container` and `field-input` utilities
```
@apply ... aria-invalid:border-red dark:aria-invalid:border-red

```
this way no need to generalize gettng error info as the nesting depends on every component as shown in this function

```
export const resolvePath = (object: any, path: string) =>
  path.split(".").reduce((o, p) => (o ? o[p] : {}), object);
```

ap-justin (changes requested):

src/components/form/Field.tsx
```diff
@@ -53,7 +55,7 @@ export function Field<T extends FieldValues, K extends InputType = "text">({
         ...(type === textarea ? {} : { type }),
         id,
         disabled: isSubmitting || disabled,
-        className: input,
+        className: `${input} ${getErrorClasses(errors, name)}`,
```
ap-justin, line 58: @abbasali1994, @SovereignAndrey

I think these red borders should also be applied to other form UIs for consistency 
image

SovereignAndrey: > I think these red borders should also be applied to other form UIs (rich text, selectors, dropzones) for consistency image

@ap-justin, agreed that's best for consistency of UI. 
@abbasali1994, please add that into this PR. Thanks! :pray:

abbasali1994: okay

abbasali1994: Added error borders for multiselects and dropzones

abbasali1994: Implemented logic using aria-invalid like suggested by @ap-justin

src/components/CountrySelector/index.tsx
abbasali1994: Done

src/helpers/resolveJsonPath.ts
```diff
@@ -0,0 +1,8 @@
+export const resolvePath = (object: any, path: string) =>
```
ap-justin, line 1: @abbasali1994, no need for this helper found in `react-hook-form/ErrorMessage` you could use [`get`](https://github.com/react-hook-form/error-message/blob/master/src/ErrorMessage.tsx#L21) util

package.json
```diff
@@ -120,6 +120,7 @@
     "react-test-renderer": "^18.2.0",
     "source-map-explorer": "^2.5.2",
     "tailwindcss": "^3.2.7",
+    "tailwindcss-aria-attributes": "^1.3.0",
```
ap-justin, line 123: @abbasali1994, `aria-*` states are already included in `3.2.7`. I think there's no need for plugin for this - As other classes like `aria-disabled:*` or `aria-selected:*` is working

ap-justin (changes requested):

src/helpers/resolveJsonPath.ts
abbasali1994: DOne

package.json
abbasali1994: aria-invalid isnt working as part of class def without the use of dependency (gives undefined error), but when defined seperately it works for tailwindcss, so modified index.css to remove the additional dependency

yarn.lock
```diff
@@ -6,13 +6,13 @@ __metadata:
   cacheKey: 8
 
 "@adobe/css-tools@npm:^4.0.1":
-  version: 4.0.1
-  resolution: "@adobe/css-tools@npm:4.0.1"
-  checksum: 80226e2229024c21da9ffa6b5cd4a34b931f071e06f45aba4777ade071d7a6c94605cf73b13718b0c4b34e8b124c65c607b82eaa53a326d3eb73d9682a04a593
+  version: 4.1.0
```
ap-justin, line 9: @abbasali1994, why is `yarn.lock` changed whilst no changes in `package.json` - did you by any change deleted this file and reinstalled deps again ?

SovereignAndrey:
@abbasali1994 please double check all changes as I cannot get past installing now to run the branch and check thing. Thanks!

yarn.lock
abbasali1994: i will check , i just ran yarn install instead of npm install , and removed the additional dependency like you requested

abbasali1994: yep i checked , if i revert it back i am getting lint errors, running yarn install brings all these changes to the file

yarn.lock
```diff
@@ -6,13 +6,13 @@ __metadata:
   cacheKey: 8
 
 "@adobe/css-tools@npm:^4.0.1":
-  version: 4.0.1
```
ap-justin, line 9: should not have changes in `yarn.lock` without changes in dependency `package.json`
see #2034

ap-justin (changes requested):

yarn.lock
abbasali1994: justin I am not contradicting that it won't work, it's not gonna work for me cause of the other PR I ended up with a lot of build issues and so I had to delete yarn.lock, and eventually clean clone everything, if your PR is an exact copy of mine lets just use that and get this code merged, cause no matter what i do on my local unless i delete the lock file and do a yarn install again my code doesn't work

yarn.lock
abbasali1994: ok pls set this PR as ready for review and lets get that reviewed and merged

ap-justin (approved):

Issue #1909, opened 2023-03-14
Admin - Charity Applications Multisig: Rewired to approve/reject Applications on Polygon


SovereignAndrey:
The Admin area for the AP Review Applications Multisig will need to be rewired to vote on and execute the application proposals on Polygon. May not actually need much more work than adding the Polygon wiring to approve/reject standard proposals, as there's no special message involved other than voting and execute the already packaged message.

PR #2023, opened 2023-03-31
Admin - Charity Applications Multisig: Rewired to approve/reject Applications on Polygon

misicnenad:
Ticket(s):
- https://github.com/AngelProtocolFinance/angelprotocol-web-app/issues/1909


src/constants/contracts.ts
```diff
@@ -69,3 +69,63 @@ const local: Contracts = {
 
 export const contracts: Contracts =
   NETWORK === "LOCAL" ? local : NETWORK === "TESTNET" ? testnet : mainnet;
+
+export const polygonContracts = {
```
misicnenad, line 73: @SovereignAndrey @ap-justin added this here to avoid having to hardcode polygon contracts in particular files all the time, but this is still a WIP, or described through a clever use of a famous meme:
image

### An idea on how to avoid having to copy/paste contract addresses as were doing with cosmwasm contracts

We can use the fact that TypeScript allows for importing JSON files and that it automatically adds types to the imported JSON objects ([VS Code supports types JSON objects](https://code.visualstudio.com/docs/languages/json), but even if you're using some other editor, TS supports typing JSON out-of-the-box as well, see [official docs](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-9.html#new---resolvejsonmodule)). This means that instead of copy-pasting particular contract addresses into the `contracts.ts` file, we can copy-paste the **entire file containing the contract addresses**.
Our code structure would look something like:
```
src
- constants
  - contracts-local.json
  - contracts-testnet.json
  - contracts-mainnet.json
  - contracts.ts // --> the file we currently use for contract addresses
```
And in `contracts.ts` we'd have:
```ts
import local from "./contracts-local.json";
import testnet from "./contracts-testnet.json";
import mainnet from "./contracts-mainnet.json";
import { NETWORK } from "./env";
...
export const contracts: Contracts =
  NETWORK === "LOCAL" ? local : NETWORK === "TESTNET" ? testnet : mainnet;
```

Files `contracts-local.json`, `contracts-testnet.json` and `contracts-mainnet.json` contents would just be copy/pasted contents of the current `contract-address.json` in EVM project. We could update the EVM deploy scripts to create separate `contract-address.json` files for mainnet, testnet and local and copy those in respective files in the webapp.

Additionally, if we were to get real fancy, we could create a script or update an existing one to automatically copy/paste these `contracts-[X].json` files into the webapp and auto-create a PR to desired webapp branch (using [GitHub CLI](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/creating-a-pull-request?tool=cli)). This way we would never have to manually copy/paste contracts again, they would always be up-to-date on both EVM project and webapp project.

src/contracts/evm/APTeamMultiSig/index.ts
```diff
@@ -0,0 +1,47 @@
...
+  processLogs(logs: TxLog[]) {
+    const execTopic = iface.getEventTopic("Execution");
+    const execLog = logs.find((log) => log.topics.includes(execTopic));
+    if (!!execLog) {
+      return "execution";
+    }
+    const execFailTopic = iface.getEventTopic("ExecutionFailure");
+    const execFailLog = logs.find((log) => log.topics.includes(execFailTopic));
+    if (!!execFailLog) {
+      throw new Error("ExecutionFailure");
```
misicnenad, line 36: Since the current `sendTx` logic [assumes the return of a single `string | undefined` attribute](https://github.com/AngelProtocolFinance/angelprotocol-web-app/blob/RC-v1.8/src/helpers/tx/sendTx/sendEVMTx.ts#L32) from the [LogProcessor](https://github.com/AngelProtocolFinance/angelprotocol-web-app/blob/RC-v1.8/src/types/evm.ts#L23) (even though it can return `any` type), it really makes it hard to let the user know that an error occurred.
Since there's no other way currently to check what was the result of the transaction, implemented this to throw an error if the execution was not successful to indicate that to the user (would be caught by the calling `try/catch`).

We should definitely look into returning something else instead of [attrValue](https://github.com/AngelProtocolFinance/angelprotocol-web-app/blob/RC-v1.8/src/helpers/tx/sendTx/sendEVMTx.ts#L32) for EVM transactions, as it is not always useful in its current form. Will create a proposal draft PR when I get some idea.

This is related to the discussion we had before https://github.com/AngelProtocolFinance/angelprotocol-web-app/pull/1972#discussion_r1147281550

src/contracts/evm/APTeamMultiSig/index.ts
```diff
@@ -0,0 +1,53 @@
+import { Interface } from "@ethersproject/abi";
+import { BigNumber } from "@ethersproject/bignumber";
+import { TxLog } from "types/evm";
+import abi from "./abi.json";
+
+const iface = new Interface(abi);
```
ap-justin, line 6: @misicnenad, `AP`, `REVIEW`, and EndowmentMultisig should have the same interface as they all come from `MultisigGeneric`. we'll separate them once difference on their interfaces appear

ap-justin (changes requested):

src/contracts/evm/APTeamMultiSig/index.ts
ap-justin: > logic assumes the return of a single string | undefined attribute

`string | undefined` is just the current types expected to be put in place of it ( may come from `LogProcessor` or `getWasmAttribute`. We could always modify this to say `unknown`

> it really makes it hard to let the user know that an error occurred.

we only need to check the `attributeValue`  as what is currently done
```ts
const { attrValue: endowId, ...okTx } = result;
        if (!endowId) {
          return showModal(TxPrompt, {
            error: "Endowment was created but failed to save to AWS",
            tx: okTx,
          });
        }
 //or if uknown
 
if(typeof endowId !== "string") {
 //... means this data is not retrieved successfully 
 //hence show prompt to user
}

``` 

> Since there's no other way currently to check what was the result of the transaction,

there's a way to check it, and should be checked on the scope where `sendTx` was called. 


> We should definitely look into returning something else instead of attrValue for EVM transactions,

sure we could return just `EVMLog[]` but it would be in conjunction to what other senders will return say `cosmosLog[]`

and the result would need to be checked either way
```
if( cosmosLog ) { ... run cosmos parsing }
if (evmLog ) { ... run evem parsing } 
```

I think coupling the processing of this data to where it came from and just returning a general `attrValue` | `returnValue` makes more sense as we just need to check that value against our expected type for that process and show appropriate error if that expectation is not met

ap-justin (changes requested):

ap-justin (dismissed):

src/constants/contracts.ts
misicnenad: Here's an example script that can be used to update contracts in the webapp:
https://github.com/AngelProtocolFinance/angelprotocol-web-app/pull/2027/files

I tested  this during the weekend on one of my dummy GH projects and it worked flawlessly.

The script is assumed to be a part of the EVM project (created here just for easier showcasing), it can be used as standalone (run as some `yarn` script) or as part of some other script and is of course subject to change as we see fit.

src/contracts/evm/APTeamMultiSig/index.ts
misicnenad: @ap-justin to get the attribute from the logs, a custom log processing function needs to be created/passed from the component calling the `sendTx(type, val, log)` func. So we need to process the logs no matter how we organize this. 

IMO it would be good to make it so that if the calling component has to extract some data from the logs, they should do it as part of the `onSuccess` function/hook, where they would have access to either `EVMLog[]` (if an EVM tx) or to `CosmosLog[]` (if Cosmos tx).
Makes it easier to reason about what is going on in the code. Also allows for better type-safety (see below).

Arguments for returning logs instead of `attrValue`:
- having to pass either `attribute` or `log` parameter to `TxContent` makes the purpose of those parameters unclear unless you know how they will be used in the subsequent chain of functions (to return `TxSuccess.attrValue`)
- for someone looking at `type TxSuccess` the purpose of `TxSuccess.attrValue` is not immediately obvious and they'd have to investigate how it was populated to understand it, especially when sending "evm" Txs (one needs to pass a function named `log` into a chain of functions and in `onSuccess` they somehow get a field named `attrValue`, no apparent connection between the two)
- might not need this atm, but current setup would stop us from being able to read multiple attributes from the logs in the future. This could be addressed by returning `unknown` as you suggested, but then we lose type-safety. Can be addressed by returning logs themselves, which are typed, which the calling component can process however they see fit.

src/contracts/evm/ApplicationsMultiSig/index.ts
```diff
@@ -0,0 +1,53 @@
...
+        "Execution",
+        execLog.data,
+        execLog.topics
+      );
+      return (id as BigNumber).toString();
+    }
+    const execFailTopic = iface.getEventTopic("ExecutionFailure");
+    const execFailLog = logs.find((log) => log.topics.includes(execFailTopic));
+    if (!!execFailLog) {
+      throw new Error("ExecutionFailure");
```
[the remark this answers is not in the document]
misicnenad: > Currently, `log / processLog` just parses logs and return something or nothin.

No reason for this function to remain unchanged if we find good reason for it. 
It would be more useful to throw different errors when they do appear than to return just "something or nothing". 

Why do you think it's better to have just these two results?

This also ties into the discussion about returning `TxSuccess.attrValue` VS returning transaction logs (see https://github.com/AngelProtocolFinance/angelprotocol-web-app/pull/2023#discussion_r1155542344). If we were to return logs in `TxSuccess`, it would make it possible to handle these different events/errors that get emitted.

Throwing errors which are handled by the calling component would allow for richer error messages both in our (console) logs and for the user.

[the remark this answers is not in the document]
misicnenad: Let's say the called contract can emit `NotEnoughConfirmations` event (or error). We could do something like:
```
if `Execution` -> showMessageToUser("Transaction successful!");
else if `NotEnoughConfirmations` -> showMessageToUser("Not enough confirmations!");
else -> showMessageToUser("Transaction failed!");
```

Without these errors, there would be no way to know what exactly happened. We could still log them for dev purposes, sure, but it would be bad UX for the user to always show them "Transaction failed" when a more precise message could have been displayed, don't you agree @ap-justin @SovereignAndrey ?

src/constants/contracts.ts
misicnenad: @ap-justin can you expand upon the benefit of this approach?
```ts
  export type Contract =
  | "registrar"
  | "index-fund"
  | "multisig/ap"
  | "multisig/review"
  | "accounts"
  | "halo"
  | "gift-card"
  | "gov"
  | "airdrop";
  
  type Contracts = { [key in Contract]: string };
  ...
  const ap = contracts["multisig/ap"]
```

src/contracts/evm/ApplicationsMultiSig/index.ts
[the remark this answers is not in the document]
ap-justin: @misicnenad,
> It would be more useful to throw different errors when they do appear than to return just "something or nothing".

what were the different errors? To give an example in `endowmentCreation` context

1. I just sent and endowmentCreation msg and expecting `id` back (evm | polygon) doesn't matter - I already included in tx additional info that will retrieve that `id` (`string | undefined/null`)
2. I got back `attrValue`
3. just check if it's defined, `id ? "Success" : "Tx submitted but we didn't get id"

> Let's say the called contract can emit NotEnoughConfirmations event (or error). We could do something like:
if `Execution` -> showMessageToUser("Transaction successful!");
else if `NotEnoughConfirmations` -> showMessageToUser("Not enough confirmations!");
else -> showMessageToUser("Transaction failed!");

`Let's say the called contract can emit NotEnoughConfirmations event` - let's say this is actually the case and not just hypotheical scenario. If you are expecting this in your transaction you should just include it in the implementation of your log processor

1. get logs
2. in these logs, find an entry with event topic `NotEnoughConfirmations event`
3. if encountered, set your `attribute` value to say `"error"` else set to other value you expect from your logs `"something"`
4. then after tx check your `attrValue` against your expectation
```ts
attrValue === "error" ? "Hey something went wrong" : "Success" 
```

realistically though, ` can emit NotEnoughConfirmations event` low confirmation can be handled in `sendEVMTx/getReceipt` recursion

src/constants/contracts.ts
misicnenad: > A middleground probably would be to retain the static list but it's value is directly from json

This can be done, would allow us to at least not have to constantly copy/paste new contract addresses into the webapp, but have it done semi-automatically (still have to approve a PR).

src/pages/Admin/Proposal/Voter/useVote.ts
```diff
@@ -1,40 +1,45 @@
...
+      type === "application"
+        ? polygonContracts.multiSig.ApplicationsMultiSigProxy
+        : polygonContracts.multiSig.APTeamMultiSigProxy;
+
+    const voteMsg: SimulContractTx = {
+      from: wallet.address,
+      to: toAddress,
+      data:
+        vote === "no"
+          ? MultiSigGeneric.revokeTransaction.encode(transactionId)
```
ap-justin, line 37: @misicnenad, It's not part of current UX to allow users to revoke already confirmed transactions, or should we allow them to @SovereignAndrey  ?

misicnenad: @SovereignAndrey @ap-justin this is actually a good question. Since our UI allows for voting "no" on transactions (in Juno called "proposals"), but in our contracts a "no" vote **does not exist** (abstaining is considered a "no" vote), we could make it so that:
1. initially only voting "yes" or abstaining is possible.
2. if the user already voted on a transaction (with a "yes" vote) and the transaction can still be voted on, they are able to change their vote to "no" in order to revoke their vote.

src/contracts/evm/ApplicationsMultiSig/index.ts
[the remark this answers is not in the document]
misicnenad: In order not to hold the PR "hostage" to this discussion and treating this change as something to be discussed more, will just return `null` when the events weren't found.

src/pages/Admin/Proposal/Voter/useVote.ts
misicnenad: @SovereignAndrey on the other hand, even if wanted to do this, it should be treated as a separate task/PR, as it includes changing the actual UI/UX logic. 
Will remove the ability to vote "no" and revoke transactions for now, but we can look into updating the flow as described in [my previous comment](https://github.com/AngelProtocolFinance/angelprotocol-web-app/pull/2023/files#r1160564271).

src/contracts/evm/APTeamMultiSig/index.ts
misicnenad: In order not to hold the PR "hostage" to this discussion, we can treat this as something to be discussed further, so updated the `log` funcs to return `null` when the events weren't found.

src/pages/Admin/Proposal/Voter/useVote.ts
misicnenad: Terminology of "yes" and "no" makes sense in the context of voting on proposals. On the other hand, in the context of "submitting a signature" to pass a multisig transaction, a terminology that would make a little more sense would probably be "submit signature" and "revoke signature".

SovereignAndrey: @misicnenad & @ap-justin: 

> @misicnenad, It's not part of current UX to allow users to revoke already confirmed transactions, or should we allow them to @SovereignAndrey ?

Yes, we have the ability to revoke a previously confirmation, so we should think about how that would look. The Decision Center UI _might_ have something to point us in the right direction. 

> @SovereignAndrey on the other hand, even if wanted to do this, it should be treated as a separate task/PR,

Yes it should be covered in a separate PR since it's an bigger enhancement with wider UI changes that will need to come with it. 

> Terminology of "yes" and "no" makes sense in the context of voting on proposals. On the other hand, in the context of "submitting a signature" to pass a multisig transaction, a terminology that would make a little more sense would probably be "submit signature" and "revoke signature".

"Confirm" and "Revoke Confirmation" makes more sense than "yes"/"no". We'd also only want to show "Confirm" option first and then only show "Revoke Confirmation" once a confirmation was signed to a TX. Having both options shown at once does not make sense.

misicnenad:
Converted this PR to draft as it will be reworked as part of general polygon-related updates, for more details see https://discord.com/channels/@me/877860812117647441/1094904098295402496.

ap-justin:
changes reflected in #2065

PR #2024, opened 2023-03-31
Web3 auth integration

abbasali1994:
Ticket(s):
-

-

## UI changes for review


package.json
```diff
@@ -76,9 +78,10 @@
     "react-helmet": "^6.1.0",
     "react-hook-form": "^7.43.1",
     "react-icons": "^4.8.0",
-    "react-redux": "^8.0.2",
+    "react-redux": "^8.0.5",
     "react-router-dom": "^6.2.0",
     "react-scripts": "5.0.1",
+    "web3": "^1.9.0",
```
ap-justin, line 84: @abbasali1994, kindly add requirements in PR description - we might be able to suggest alternatives to adding `web3`

SovereignAndrey: It is a heavy package, but web3auth delivers a critical UX feature requested by many charities to help simplify the web3 experience. We'll take a heavier, feature rich app over lighter with less features that causes friction points for users. We can always work on refactoring to a lighter package that does the same things if a good setup can be arrived at. :+1:

SovereignAndrey (approved):

package.json
ap-justin: @SovereignAndrey, @abbasali1994 
I was not directly vetting against the use `web3js` package and just clarifying the requirements so I can start reviewing this PR properly and even suggest a different integration approach if there's one

e.g 
1. instead of installing `"@web3auth/modal"`, is it possible to just use `@web3auth/no-modal` as we already have wallet selection modal ?

2. just looking now i saw that `web3js` is used just to call `getAccounts` and `getChainIDs` which could be also be implemented by just using existing `EIPMethods` 

3. is there a need for `@web3auth/metamask-adapter` while we have native support for metamask ? 


> We can always work on refactoring to a lighter package that does the same things if a good setup can be arrived at. 👍

I agree with this, but the problem is we don't usually have the capacity to do refactors  as we have higher priority task to do - and is doomed to work with the burden of these new packages (slower dev server) until such time we have bandwidth to deal with this. 

Anyhow, even merged, I'll leave comments to this PR as a reference on areas needed to be updated

PR #2025, opened 2023-03-31
Tx and query migration to polygon

ap-justin:
Ticket(s):

merging `tx` and `query` updates in this PR as they are interrelated
All pending works listed in 
* #2072 

merged:
* #2040 
* #2057 
* #2053 
* #2065 
* #2074 

### Migrate juno queries to polygon counterpart
- [x] "registrar.config" 
- [x] "registrar.config-extension" (included in `registrar.config`) **[A/N: Not needed as it's in registrar config now and we'll respect that setup vs Cosmos]**
- [x] "index-fund.funds"
- [x] "index-fund.alliance-members" 
- [x] index-fund.config" 
- [ ] gift-card.balance" **[A/N: Gift Card contract should be setup now]**
- [x] cw20.balance (erc20)
- [x] cw4.members (multisig `getOwners`)
- [x] cw4.member (multisig owner has no more details (e.g weight)
- [x] cw3.voter (same as cw4.member)
- [x] cw3.voters same as `cw4.members`
- [x] cw3.config (constructed from `threshold` and `requireExecution` query )
- [x] cw3.proposals (array of transaction/proposal id)
- [x] cw3.proposal  
- [x] cw3.votes (array of owners who signed: multisg `getConfirmations`) 
- [x] accounts.endowment
- [x] accounts.state
- [x] accounts.token-balance (prev includeded in account.state)
- [x] accounts/settings.controller (included in `accounts.endowment`)

### Migrate juno tx to polygon
- [x] cw3.config (`threshold` and `requireExecution` in own tranasaction )
- [x] cw3.execute-tx - (migrated) for testing
- [x] cw3.add/submit-tx - (migrated)  for testing
- [x] cw3.vote - (migrated) for testing
- [x] cw3/review.config - (migrated) same as generic multisig - for testing
- [x] cw3/review.vote-application - in separate ISSUE
- [x] cw3/charity.withdraw - in separate ISSUE
- [ ] accounts.change-status 
- [x] accounts.withdraw (for testing) 
- [ ] accounts.update-strategy (new implementation - not implemented in juno)
- [x] accounts.endowment-settings (new implementation - not implemented in juno)
- [x] accounts.deposit ( not used in juno, but handy to add for adding funds to endowment)
- [x] cw4.update members  - (migrated multisig.add-owner/remove-owner) for testing
- [x] cw20.transfer 
- [x] (new) erc20.approve
- [x] cw20.send (future along with gov)
- [ ] giftcard.deposit **[A/N: Gift Card contract should be setup now]**
- [ ] giftcard.spend **[A/N: Gift Card contract should be setup now]**
- [x] index-fund.config - migrated:for testing
- [x] index-fund.update-owner 
- [x] index-fund.create-fund (error)
- [x] index-fund.remove-fund 
- [x] index-fund.remove-member-from-funds (new: not implemented in juno) 
- [x] index-fund.update-fund-members 
- [x] index-fund.update-alliance-members  
- [ ] index-fund.update-alliance-member (not implemented)
- [ ] registrar.config
- [ ] registrar.owner

### Migrate Juno queries to polygon counterpart: 
- gov.staker" (future)
- gov.state" (future)
- gov.config" (future)
- gov.polls (future)
- cw20.info (future)

### Migrate Juno TX to Polygon
- airdrop.claim (future)
- gov.stake via `cw20.send` (future)
- gov.create-poll (future)
- gov.unstake (future)
- gov.claim (future)
- gov.end-poll (future)
- gov.vote (future)
- lp.buy (future)
- lp.sell (future)

-

## UI changes for review


misicnenad:
Checked out the proposed design and can say after hours of looking at it I got enough of a grip around it to be able to create a review.

### TLDR review
The design seems to be a product of dissatisfaction with the amount of boilerplate in current Juno contract logic and attempts to address it (see https://github.com/AngelProtocolFinance/angelprotocol-web-app/pull/2035#discussion_r1158022115).
It addresses it in a very clever use of TS type system and a lot of "convention" in design to circumvent some of the TS type system shortcomings, as well as to handle some edge cases that necessarily appear with this design (handling erc20 txs and different multisig addresses that have the same implementation, i.e. AP and Review multisigs).

The moderate use of convention in this design can be ignored, as it can be framed as just a coding style disagreement, but what needs to be addressed it that this design introduces a problem that's very hard to ignore:

**It uses "clever" and terse code to such an extent that it drastically reduces readability and comprehension.**
**Readability and comprehension are extremely important for long-term health of the project and especially when working in teams, no matter how small they may be.**

### Detailed review
I want to address the fact that the code is very hard to read and understand. Will start by noting where the so-called "convention" was used and what I think were reasons for it and than turn the discussion to why the code as it is set up causes comprehension problems.

 The review will be formed as an answer to a [question](https://github.com/AngelProtocolFinance/angelprotocol-web-app/pull/1970#discussion_r1156860193) posted in PR #1970, quoted below:

> How is this not typesafe ? it shows you exactly the txs you could create
> image

I wasn't talking about the `createTx` function, it seems to have type-safety set up. Problems can arise because typings in the new design seem to depend more on convention that on strict types. This may be a coding style preference, but we should minimize convention whenever possible as it's inherently not type-safe. 

Will just point out the parts where this "expected convention" exists which may or may not be problematic and where applicable, will try and create a workaround or redesign recommendation, so we can discuss.

For example:
**1. There is a missing "link" between [type Txs](https://github.com/AngelProtocolFinance/angelprotocol-web-app/blob/evm-query/src/contracts/createTx/types.ts#L19) and [type Contract](https://github.com/AngelProtocolFinance/angelprotocol-web-app/blob/evm-query/src/types/lists.ts#L35), in the sense that nothing is stopping us from mistyping a contract name in `Txs` or just adding a non-existing one. For example it is possible to update `multisig.add-owner` into `multisgi.add-owner` (notice the last 2 letters of "multisig" are mistyped) and the compiler wouldn't complain that the contract `multisgi` does not exist.**

This was probably intentional because you tried to both have as much type-safety as possible, but also accommodate the fact that AP and Review multisig contracts share the same underlying functions, but have different addresses. This is evident by the fact that `type Contract` has:
```ts
| "multisig/ap"
| "multisig/review"
```
as possible contracts, but `type Txs` assumes the existence of only `multisig.{func_name}`. Which ties into the next point.
**2. [type TxOptions](https://github.com/AngelProtocolFinance/angelprotocol-web-app/blob/evm-query/src/contracts/createTx/types.ts#L36) depends on the convention that `type Txs` values are always in format `${contract_name}.${func_name}`, but it also have a check just in case this is not the case, which makes this actually fine. And then it has a check for whether the contract_name part extends [type Contract](https://github.com/AngelProtocolFinance/angelprotocol-web-app/blob/evm-query/src/types/lists.ts#L35) and if it doesn't, it adds a new field with the name of contract_name. This field is processed in [encodeTx](https://github.com/AngelProtocolFinance/angelprotocol-web-app/blob/evm-query/src/contracts/createTx/createTx.ts#L20) again by convention (assuming `type: T` will be in format `${contract_name}.${func_name}`). This was obviously done to be able to pass specific AP/Review multisig addresses and ERC20 addresses and we can see as much in places where those contracts are targeted (e.g. [useUpdateMembers](https://github.com/AngelProtocolFinance/angelprotocol-web-app/blob/evm-query/src/pages/Admin/templates/cw4/Members/useUpdateMembers.ts#L59)).**

**This is admittedly a very clever way to set up code to process these cases, but it makes the code much _much_ harder to understand, which is very important when working in teams. And as Google results would show when searching ["clever vs readable code"](https://www.google.com/search?q=clever+vs+readable+code), it's much better to have understandable (readable) code than the one that is clever or terse.**

To get back to the comprehension problem at hand, it is not immediately obvious why it's necessary to check whether pre-dot `Txs` values are part of `Contract` type, why `encodeTx` returns some random field as destination if it's not a contract and it takes hours of looking at the code and how it's used and going back and forth between all those files to start getting a hang of it.

In case you feel like judging me for taking hours to understand something that may seem obvious to you, bare in mind:
1. This just shows the disadvantage of "clever code" when working in a team, as opposed to "readable code"
2. You were the one who spent even more hours (if not days) thinking about and creating this design, so it's logical that at this point you'd at least understand what you were building.

Finally, I tried to see if it would be possible to make the code rely more on types than on convention, but my attempts have been largely unsuccessful. They mostly revolved around trying to make `type Contract` into `enum Contract` and rewriting `type Txs` into some kind of structure that would resemble this:
```ts
export type Txs = {
  [Contract.accounts]: {
    createEndowment: Tx<NewAIF>;
    updateController: Tx<SettingsControllerUpdate>;
  };
  erc20: {
    transfer: Tx<Transfer>;
  };
  [Contract.multisigAp]: {
    submitTransaction: Tx<NewTransaction>;
    addOwner: Tx<{ address: string }>;
    removeOwner: Tx<{ address: string }>;
    confirmTx: Tx<{ id: number }>;
    executeTx: Tx<{ id: number }>;
  };
  [Contract.multisigReview]: {
    submitTransaction: Tx<NewTransaction>;
    addOwner: Tx<{ address: string }>;
    removeOwner: Tx<{ address: string }>;
    confirmTx: Tx<{ id: number }>;
    executeTx: Tx<{ id: number }>;
  };
};
```

**But even if this rewrite was to succeed, it wouldn't change the fact that this design is very difficult to understand.** It's a rewrite of the logic we use for Juno that drastically reduces the code size (see relevant review of alternative design pointing this out [here](https://github.com/AngelProtocolFinance/angelprotocol-web-app/pull/2035#discussion_r1158022115)). But still very difficult to understand.

ap-justin:
@misicnenad, @SovereignAndrey 

> The design seems to be a product of dissatisfaction with the amount of boilerplate in current Juno contract logic and attempts to address it

Not solely about "boilerplate" (nor negative aspects of previous structure). I also provided the advantages of this new approach, which for some reason you don't reference here

>  a lot of "convention" in design to circumvent some of the TS type system shortcomings,

you mean this convention `${Contract}.${string}` ? (which is just a [template type](https://www.typescriptlang.org/docs/handbook/2/template-literal-types.html)) that just uses this another typescript type ? 

```ts
export type Contract =
  | "registrar"
  | "index-fund"
  | "multisig/ap"
  | "multisig/review"
  | "accounts"
  | "gift-card";
```

> The moderate use of convention in this design can be ignored, as it can be framed as just a coding style disagreement,

I don't see why using typescript constructs is considered `convention` or `coding-style` would you please elaborate? 

> I wasn't talking about the createTx function, it seems to have type-safety set up. Problems can arise because typings in the new design seem to depend more on convention that on strict types. 

Again, need to elaborate why a valid typescript type `${Contract}.${string}` is considered "convention". And when you say `New design seem to depend more on convention that on strict types`

are you referring to this? (a signature strictly typed)
```ts
export async function queryContract<T extends QT>(
  type: T,
  options: QueryOptions<T>
): Promise<ReturnType<Q[T]["transform"]>>
```
where `QueryOptions` is also strictly typed
```ts
export type QueryOptions<T extends ContractQueryTypes> =
  T extends `${infer C}.${string}`
    ? C extends Contract
      ? //if args is also null, options is empty
        QueryArgs<T> extends null
        ? Empty
        : QueryArgs<T>
      : //if contract address is not hardcoded, need to supply
        Record<C, string> & (QueryArgs<T> extends null ? {} : QueryArgs<T>)
    : Empty;
```

there's no convention here, this is all strictly typed and no implicit knowledge is required to interact with this signature. 


A convention is something like this. I hope you would see the difference
```ts
createTx(
"sender", 
"     ", //Hey there's no typings for this, but please just provide a string with this format  xxx.xxxx
{ /** ..arsg */ },  //there's also no typing for this, so go to definitions file and see the appropriate arguments for this
)
```


> here is a missing "link" between type Txs and type Contract, in the sense that nothing is stopping us from mistyping a contract name in Txs or just adding a non-existing one

Sure you could type an incorrect transaction, but since it is `strictly` typed. You will get an error
//image here


>  type TxOptions depends on the convention that type Txs values are always in format ${contract_name}.${func_name}, but it also have a check just in case this is not the case, which makes this actually fine. And then it has a check for whether the contract_name part extends type Contract and if it doesn't, it adds a new field with the name of contract_name. This field is processed in encodeTx again by convention (assuming type: T will be in format ${contract_name}.${func_name}). This was obviously done to be able to pass specific AP/Review multisig addresses and ERC20 addresses and we can see as much in places where those contracts are targeted (e.g. useUpdateMembers).

I don't know if you are referring to this as a problem or just explaining how the API works.
`it adds a new field with the name of contract_name`. Some contracts like `ERC20` or `Multisig` is generic and could work whatever `address` you provide that's why it's dynamic.

If however there's a specific multisig (for example) tx you want to add, it could be added as 
```ts
"multisig/ap.tx_name": ...tx_info
```
this way, `multisig` address won't be required, as it's clear that this tx is unique for `multisig/ap`


> This is admittedly a very clever way to set up code to process these cases, but it makes the code much much harder to understand, which is very important when working in teams. And as Google results would show when searching "clever vs readable code", it's much better to have understandable (readable) code than the one that is clever or terse.

I don't even consider this code clever/hacky, as the signature is well-typed and the implementation is fairly straight forward.

```ts
//encode tx
const [contract_key] = type.split(".");
const { [contract_key]: c, ...args } = options
const contract =
    contract_key in contracts ? contracts[contract_key as Contract] : c;

return [txs[type](args), contract];
```

If, however, you find that implementation has some casting, there's a tradeoff you need to consider between making your signature precise, vs implementation where mentioned in this book [effective typescript](https://effectivetypescript.com/) that it's always better to make the signature precise as this is the user facing code 


> But even if this rewrite was to succeed, it wouldn't change the fact that this design is very difficult to understand. It's a rewrite of the logic we use for Juno that drastically reduces the code size (see relevant review of alternative design pointing this out here). But still very difficult to understand.

The main theme of this whole comment is `this is hard to understand`. Are we sure we wan't to forego these benefits 
* declarative function that's fully typed with autocompletion
* ~95% less code when adding new contracts ( since we don't need to create another class ) - see overall change in demo PR
* ~50% less code when adding new transaction
- see single template wiring refactor in demo PR
* no boilerplate


because of this reason? which:
*  doesn't have informative explanation as to why, other than mentioning vague ideas like `convention` ,`type disconnect` which clearly isn't the case 


@SovereignAndrey, @misicnenad
As a summary, I didn't choose this approach just because I feel like a change is good to have. This is after considering how the previous structure fared with juno and how would this fare to all of our existing (migrated to polygon) and future (governance) transactions and also queries (`type Contract` is also used in queries, so that when querying a hardcoded contract, there's no need to input the contract address in the query call)

Despite the reasons I've provided, let's assume we go with #2025 , the effect would be
1. All migration progress in this PR would be have to be re-written/re-configured 
2. As that design is based only on `~2` transactions? (didn't considered query calls, future transaction formats(gov), processing of decoded types). Using this new design means re-refining for the edge cases as it is encountered on the rewriting/reconfiguration process

> Note: code in this PR is not in its final form, many checks and optimizations were omitted for simplicity. If the team agrees to approve these changes (in this or modified form), those checks/optimizations would be added.


> Any and all suggestions to make this even better will be greatly appreciated!
 
And let's say @misicnenad, still continues this parallel work, I would have to spend time 
looking for suggestions which may already be implemented on the very PR I'm workin (which seems like I'm just duplicating my work). Or point out shortcomings that may also be already addressed in #2025

ap-justin:
@SovereignAndrey, @misicnenad 
Working on #2025, and also commenting/arguing/discussing on #2036 that's just a duplicate of what I've been doing here is a significant slowdown on my part. Not to mention it's just a structure based on `2` transactions that's slowing considerable amount of progress in #2025. I hope we could convert #2036 to a ticket/discussion which we could discuss from time to time as it is not @misicnenad's task in the first place. 

If,  however, @misicnenad, still insist on that duplicate work - I suggest it is pointed in this draft and applied to every `tx` and `query` already migrated here so we could clearly see its significance and not just discuss theoreticals

misicnenad:
@ap-justin 

> Not solely about "boilerplate" (nor negative aspects of previous structure). I also provided the advantages of this new approach, which for some reason you don't reference here

I don't think I ignored any of the arguments, I just boiled them down to "the main advantage is less code/less boilerplate". Please list any other arguments I may have ignored, I will address them.

> you mean this convention `${Contract}.${string}`
> ...
> Again, need to elaborate why a valid typescript type `${Contract}.${string}` is considered "convention". And when you say New design seem to depend more on convention that on strict types

Yes,`${Contract}.${string}` is a convention, but let me expand on that. 
This is referring to [type Txs](https://github.com/AngelProtocolFinance/angelprotocol-web-app/blob/evm-query/src/contracts/createTx/types.ts#L19) which is not a fixed template type as this comment may imply. A more precise "definition" for it would be `${Contract OR "erc20" OR "multisig"}.${string}`. Even more precise would be `${string}.${string}`, as there's nothing stopping us from inputting any string in the pre-dot part. But even this is not it's true type definition. 
The type also allows for itself to be extended with a value that doesn't contain any dots, like:
```ts
type Txs = {
  "accounts.create-endowment": Tx<NewAIF>;
  "accounts.update-controller": Tx<SettingsControllerUpdate>;
  ...
  "just_some_single_value_without_a_dot": Tx<SomeType>;
}
```
which makes `Txs` true format actually just `string`. The fact that the code "just works" is that we use the "dot separator" convention for the type's field names.

You can see that all types/functions that use `Txs` (like `TxOptions`) depend on the **convention** (or "assumption") that `Txs` has the format of `${Contract}.${string}`. If it does not, the design falls back to using basically a generic `object` type (the `Empty` type).
  
> ```ts
> export type QueryOptions<T extends ContractQueryTypes> =
>   T extends `${infer C}.${string}`
>     ? C extends Contract
>       ? //if args is also null, options is empty
>         QueryArgs<T> extends null
>         ? Empty
>         : QueryArgs<T>
>       : //if contract address is not hardcoded, need to supply
>         Record<C, string> & (QueryArgs<T> extends null ? {} : QueryArgs<T>)
>     : Empty;
> ```

This type definition is even more difficult to understand than `TxOptions`.

>  convention is something like this. I hope you would see the difference

A convention is saying "please always use `${Contract}.${string}` format when extending `Txs` with new contracts/functions, otherwise the function won't know what arguments it needs to accept".

> I don't know if you are referring to this as a problem or just explaining how the API works

Wanted to assure any reader that I actually attempted to understand the code by explaining what it does and why it seems to do it. By putting the apparent rationale behind the code into writing, I attempted to show how difficult it is to even understand the writing itself, not to mention the code - which ties into my main point of code being very hard to understand.

**NOTE: when I say the code is "hard to understand", I'm not referring to the way it's supposed to be used with `createTx` or `encodeTx`, those are simple enough to use. I'm of course referring to how difficult it is to understand what is going on underneath those functions, what makes them work at all.**

 > If however there's a specific multisig (for example) tx you want to add, it could be added as
> ```ts
> "multisig/ap.tx_name": ...tx_info
> ```
> this way, multisig address won't be required, as it's clear that this tx is unique for multisig/ap

The design does allow for that, yes. I don't think I made any reference to this not being possible.

> If, however, you find that implementation has some casting, there's a tradeoff you need to consider between making your signature precise, vs implementation where mentioned in this book [effective typescript](https://effectivetypescript.com/) that it's always better to make the signature precise as this is the user facing code

I didn't make any reference to "casting" as being a problem, but this piece of code `const [contract_key] = type.split(".")` again shows a dependence on "convention" (`type` being in format `${string}.${string}`).

> The main theme of this whole comment is this is hard to understand. Are we sure we wan't to forego these benefits

All the benefits listed can be narrowed down to:
1. it uses _declarative_ code
2. it uses _less_ code

Number 2 is definitely an objective benefit.
Number 1 on the other hand is just a coding paradigm change, which admittedly ties into number 2, but has no other inherent advantages as opposed to using imperative code.

> doesn't have informative explanation as to why, other than mentioning vague ideas like `convention` ,`type disconnect` which clearly isn't the case

I think I made a detailed explanation for my argument, what parts of my argument did you find lacking in information, I'd be happy to expand on them?

> This is after considering how the previous structure fared with juno and how would this fare to all of our existing (migrated to polygon) and future (governance) transactions and also queries 

Please provide more details on the following:
1. How did previous structure fare poorly with juno in your opinion that warrants an update?
2. How would the previous structure not be adequate for the future (governance) transactions?
3. How would the previous structure not be adequate for the queries?

> ...(type Contract is also used in queries, so that when querying a hardcoded contract, there's no need to input the contract address in the query call)

This again only falls under the "previous structure uses boilerplate" argument (which is a fair argument, but must be taken in the context of making code easy to read).

> Despite the reasons I've provided, let's assume we go with https://github.com/AngelProtocolFinance/angelprotocol-web-app/pull/2025

Presumably you meant to say "let's assume we go with https://github.com/AngelProtocolFinance/angelprotocol-web-app/pull/2035".

> ...the effect would be
> 
>   1. All migration progress in this PR would be have to be re-written/re-configured

This wouldn't have been a problem had the design proposal been implemented as a PR dedicated to just that redesign, instead of containing many other unrelated changes. But now this is the situation we find ourselves in.

>   2. As that design is based only on `~2` transactions? (didn't considered query calls, future transaction formats(gov), processing of decoded types). Using this new design means re-refining for the edge cases as it is encountered on the rewriting/reconfiguration process

"Using this new design" - #2035 is not using a "new" design, it is using the "current" design. 
Query calls can be inserted in #2035, I can create an example.
Can you expand on how the #2035 design would be inappropriate for "future transaction formats(gov), processing of decoded types"?

> Using this new design means re-refining for the edge cases as it is encountered

Well to be fair, this is the case for any possible design, even the one proposed in this PR.

misicnenad:
> @SovereignAndrey, @misicnenad Working on #2025, and also commenting/arguing/discussing on #2036 that's just a duplicate of what I've been doing here is a significant slowdown on my part. Not to mention it's just a structure based on `2` transactions that's slowing considerable amount of progress in #2025. I hope we could convert #2036 to a ticket/discussion which we could discuss from time to time as it is not @misicnenad's task in the first place.
> 
> If, however, @misicnenad, still insist on that duplicate work - I suggest it is pointed in this draft and applied to every `tx` and `query` already migrated here so we could clearly see its significance and not just discuss theoreticals

@ap-justin you are creating a completely new design for how EVM transactions are created, which would be used by the whole team for the foreseeable future. Therefore you must expect and appreciate feedback, criticism, suggestions and even alternative solutions. It is worth spending time investigating how to do this correctly and in a way that's beneficial for all team members.

This is something that will greatly affect how we write future code throughout the whole app, as it affects how blockchain Txs are created. You cannot take it upon yourself to decide how this will be done without taking any input from other people from the project who will be using that same code to write their own, so **this IS my task as much as it is yours** (talking of course just about the actual tx creation redesign, not the other wiring stuff).

ap-justin:
@misicnenad, @SovereignAndrey 
 >  Therefore you must expect and appreciate feedback, criticism, suggestions and even alternative solutions. It is worth spending time investigating how to do this correctly and in a way that's beneficial for all team members.
 
Have I not considered your suggestions ? I have

> Did you consider using a design similar to how we implemented Juno contract message creation, like [contracts/Account](https://github.com/AngelProtocolFinance/angelprotocol-web-app/blob/RC-v1.8/src/contracts/Account.ts)?

and have replied accordingly [see full reply](https://github.com/AngelProtocolFinance/angelprotocol-web-app/pull/1970#discussion_r1155815135) providing, initial reasons why the old approach isn't optimal for the new requirements

> Did you consider using a design similar to how we implemented Juno contract message creation, like

> I decided not to copy that same implementation which used as

> const account = new Account("", wallet)
> const tx = account.createEndowTx({...})
> and too many boillerplate in implementation (only difference is that inner content

> createEmbeddedChangeEndowmentStatusMsg(payload: StatusChangePayload) {
    return this.createEmbeddedWasmMsg(Account.address, {
      update_endowment_status: payload, //<--- inner content
    });
  }
  
  
You obviously didn't like agree with the reasons I provided, so instead of providing feedback in this PR, you started creating a PR of your own  #2035 in parallel to this task that I am working #2025 (which could have been more comprehensible and comparable if those changes are pointed here) 

This led me to take significant amount of time providing reasons why #2035 is suboptimal and shortsighted ( design based on 2 transactions only), again considering everything
* existing transactions
* future transactions
* queries 
* cache invalidation

But despite all that, you still insist to go with your premature design which effect would be 
* reconfiguring all the progress done in this PR, see list 
* handling again different edge cases, optimizations where you mentioned 


> code in this PR is not in its final form, many checks and optimizations were omitted for simplicity. If the team agrees to approve these changes (in this or modified form), those checks/optimizations would be added.

All of these causes further slowdown, duplicated effort, which will make us not meet our upcoming demo this week.

SovereignAndrey (approved):
Really great work! :clap:

PR #2027, opened 2023-04-03
Create draft script to update contracts

misicnenad:


ap-justin:
@misicnenad, @SovereignAndrey, there might be no need for this as most contract addresses are `proxy`
image

Issue #2021, opened 2023-03-30
Hide Locked Withdraw for Normal/AST members in Admin


SovereignAndrey:
Not possible for early locked withdraws on AST/Normal Endowments so there's no need to show that. They should only see their Liquid balance and interact with a liquid withdraw form.

PR #2028, opened 2023-04-03
Remove 'locked' tab for normal/AST endows

misicnenad:
Ticket(s):
- https://github.com/AngelProtocolFinance/angelprotocol-web-app/issues/2021

Since `AccountTabs` component is used only in [Withdrawer](https://github.com/AngelProtocolFinance/angelprotocol-web-app/blob/RC-v1.8/src/pages/Admin/Charity/Withdraws/Withdrawer/index.tsx), moved its contents inside it - other than placing it in the only placed where it's used, it also allows for simpler hiding of "Locked" tab/form from AST endowments.

- connect Charity 1 wallet
- go to the related AST endowment's admin withdrawal page http://localhost:4200/admin/27/withdraws
- verify neither "Locked" tab nor the corresponding form appear

## UI changes for review

AST:
- desktop:
image
- mobile:
image

Charity:
- desktop:
image
- mobile:
image


src/pages/Admin/Charity/Withdraws/Withdrawer/index.tsx
```diff
@@ -17,17 +23,30 @@ export default function Withdrawer() {
       }}
     >
       {({ tokens_on_hand }) => (
-        <AccountTabs
```
ap-justin, line 20: @misicnenad, should not revert this abstraction, as this is used in strategies before which is now auto-invest once reimplemented. You could however move this to `/Withdrawer` as it is only used there atm

ap-justin (changes requested):

src/pages/Admin/Charity/Withdraws/Withdrawer/index.tsx
misicnenad: @ap-justin it's unclear what you meant to say, do you mean that `AccountTabs` will be used again?

ap-justin: currently removed as wiring is intended for `auto-invest` ,but would be brought back once we could query strategies from somewhere
image

misicnenad: > currently removed as wiring is intended for auto-invest

@ap-justin still unclear what you meant to say, **do you mean that AccountTabs will be used again?**

misicnenad: @ap-justin anyway, this component seemed to be a wrapper around `headlessui > Tab`, but other than an additional level of abstraction it added nothing new - no custom styles nor some extra logic.

To use your way of thinking, let's use the `Tab` component directly for now and wrap it again when there is a need for that.

ap-justin: @misicnenad  
 > nothing new other than an additional level of abstraction - no custom styles nor some extra logic.
 
 you mean for when I add back strategies tabs I need to add these boilerplate again? 
 ```ts
 const types: { [key in EndowmentType]: AccountType[] } = {
  charity: ["liquid", "locked"],
  normal: ["liquid"],
};
```
```ts
 {types[endow_type].map((type) => (
              <Tab.Panel key={`tab-panel-${type}`}>
```
or code back `<AccountTabs/>` component to bring these to pages the way there were before? 
```tsx
<AccounTabs>
  <Withdraw liquid />
  <Withdraw locked />
</AccountTabs>
```

```tsx
<AccounTabs>
  <Strategies liquid />
  <Strategies locked />
</AccountTabs>
```

Don't remove this component as there's a usecase for this which I just showed you. 

for this task you could even just simple return a single form 
```tsx
if(endow = normal)
 return <Withdraw type="liquid"/> 
 
 else return 
 <AccounTabs>
  <Withdraw liquid />
  <Withdraw locked />
</AccountTabs>

```

src/pages/Admin/Charity/common/AccountTabs.tsx
```diff
@@ -1,35 +0,0 @@
-import { Tab } from "@headlessui/react";
-import { Children, ReactElement } from "react";
-import { AccountType } from "types/contracts";
-
-const types: AccountType[] = ["liquid", "locked"];
-export default function AccountTabs(props: {
```
ap-justin, line 6: no need to delete this component as this is to be used 
rather 
```
if(endow == normal)
return <Withdraw  form = "liquid />

else return
<AccountTabs>
  <Withdraw liquid/>
  <Withdraw locked/>
<AccountTabs/>
```

ap-justin (changes requested):

src/pages/Admin/Charity/Withdraws/Withdrawer/index.tsx
SovereignAndrey: @ap-justin The auto-invest feature was removed a while back from the contracts. We're not planning on adding that anymore since the cross-chain gas fees (both getting accurate estimates to pay & handling failed transactions) complicates things.

ap-justin (approved):

PR #2035, opened 2023-04-04
EVM Contract Logic: Proposal to reuse the logic used to create transaction messages for Terra and later Juno contracts

misicnenad:
The logic used to create transaction messages to send to Juno contracts (previously Terra too) is based around creating a TS class for each contract we needed to interact with.
For any of the contract's functions we'd create a class method that created an appropriate transaction message which we then pass to the `sendTx` function (passed to the underlying `sendCosmosTx`, see [source](https://github.com/AngelProtocolFinance/angelprotocol-web-app/blob/RC-v1.8/src/helpers/tx/sendTx/sendCosmosTx.ts)).

For more details about Juno contract classes used in the webapp check out the link below:
https://github.com/AngelProtocolFinance/angelprotocol-web-app/tree/RC-v1.8/src/contracts

Main advantages of this approach:
1. Tried and tested logic, served us well until now.
2. All team members already have a good understanding of the design.
3. Types, data and logic (functions) kept in the same file.
4. Easy to understand, reason about, maintain and extend.

With these benefits in mind, created a proposal implementation of this same pattern for our current EVM contract code (`Account` and `ERC20`) to showcase that it can be used for them as well.


src/contracts/evm/Account/Account.ts
```diff
@@ -0,0 +1,62 @@
...
+  },
+} as const;
+
+type Events = typeof eventTemplates;
+
+/**
+ * Representation of the `Account` contract that extends and wraps the generic `Contract` functionality.
+ * Ensures that `Contract` methods are called with the Account ABI.
+ */
+export default class Account extends Contract {
```
ap-justin, line 28: @SovereignAndrey , @misicnenad - I have mentioned  in [pr comment](https://github.com/AngelProtocolFinance/angelprotocol-web-app/pull/1970#discussion_r1155815135) why this old approach is not to be copied moving forward to polygon, but since you still insist that this structure is to be retained and done initial works in parallel to #2025.  I would take the time here to explain in detail why I believe this approach is suboptimal considering how it fared throughout the app (juno setup),

long thread coming up.

ap-justin: @misicnenad, @SovereignAndrey 

* ### Class structure (inheritance) works well with the nature of cosmos msgs, but not with the new EVM setup.

To encode a cosmos `msg` `MsgExecuteContract | MsgSend` or any other type of msg, the contract doesn't need to know what contract that msg came from.

that is
```ts
//all MsgExecuteContract
const msg1:MsgExecuteContract = account.createMsg(..args)
const msg2:MsgExecuteContract = multisig.createMsg(..args)
const msg3:MsgExecuteContract = registrar.createMsg(..args)

```
that's why currently in `contracts/Contract` there's an abstracted `createExecuteContractMsg`

```ts
createExecuteContractMsg(
    to: string,
    msg: object,
    funds: Coin[] = []
): MsgExecuteContractEncodeObject {
  return {
    typeUrl: "/cosmwasm.wasm.v1.MsgExecuteContract",
    value: {
      contract: to,
      sender: this.walletAddress,
      msg: toUtf8(JSON.stringify(msg)),
      funds,
    },
  };
}
```
which classes inheriting from `class Contracts` can use and just need to pass tx content.
```ts
//accounts.ts
createDepositMsg(payload: DepositPayload, funds: Coin[]) {
  return this.createExecuteContractMsg(
    Account.address,
    {
      deposit: payload,
    },
    funds
  );
}

```
this eventually will be encoded using `MsgExecuteContract.encode` method.


This is not the case with EVM contracts. The information to encode/decode transaction data is 
bound to the contract interface. Making a generic `Contract` and then extending it seems pointless

```ts
//Contract.ts
 protected constructor(
    abi: string[],
    contractAddress: string,
    wallet?: WalletState
  ) 

//Account.ts
constructor(contractAddress: string, wallet: WalletState) {
  super(abi, contractAddress, wallet);
}
```

`Account` has its own identity (`abi`), and has the capability to `encode/decode` transactions base on that identity. Why does it need to extend a generic `Contract`, pass that distinct identity and extend it later on?

You could just instantiate an account interface and use it directly in your tx and query definitions
```ts
//account/index.ts
const account = new Interface(account_abi)

//query.ts
({ id }) => accounts.encodeFunctionData("queryEndowmentDetails", [id])

//tx.ts
 "accounts.create-endowment": (aif) =>
    accounts.encodeFunctionData("createEndowment", [toTuple(aif)]),

```
no additional encode/decode wrappers


in case of events, just construct your event topic always on the context of the transaction where this topic is needed

```ts
const ev = accounts.getEvent("EndowmentCreated");
const topic = accounts.getEventTopic(ev);
/** 
 * NOTE use iface.encodeFilterTopics, to narrow down
 * topic (include specific arg values)
 * 
*/
```

src/contracts/evm/Contract.ts
```diff
@@ -0,0 +1,121 @@
...
+  Tupleable,
+  TxLog,
+} from "types/evm";
+import { WalletState } from "contexts/WalletContext";
+import { toTuple } from "helpers";
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
+import Account from "./Account/Account";
+
+/** Represents generic contract functionality. Needs to be extended by actual contracts like {@link Account}. */
+export default class Contract {
```
ap-justin, line 14: @misicnenad, @SovereignAndrey, 
If this structure is the right way to go ( which I believe is not with reasons mentioned in other comments ) we could just install `@ethersproject/contract` along with `@ethersproject/provider` or for this matter just install `ethers` as a whole (that is using this packages somehow doesn't manifest inefficiencies outlined in #1966)

There's no need for another PR parallel to #2025 just to re-create these functionalities.

ap-justin:
@misicnenad 
> Tried and tested logic, served us well until now.

just using this old structure to bring `polygon` related wiring to the app won't make these wirings/logic work automatically. Each piece of wiring whether within some old or new structure would need to be tested/ensured to be working


> Easy to understand, reason about, maintain and extend.

As someone who worked(add/edit/remove) on most of these transactions, I would say it's hard to maintain that's why previously (long before this polygon migration), I explored ways to simplify this clickup: (https://app.clickup.com/t/3w4nd1j)
draft pr: https://github.com/AngelProtocolFinance/angelprotocol-web-app/blob/txs/src/slices/transaction/transactors/prepareTx/genTxResource.ts

misicnenad:
@ap-justin @SovereignAndrey 

In short, the main criticism of this design proposal to reuse Juno contract logic is that it introduces boilerplate. This is not bad in and off itself.

I left a response on the alternative proposal #2025 for why creating "smaller" code might not be so good in the long term. To check it out see https://github.com/AngelProtocolFinance/angelprotocol-web-app/pull/2025#issuecomment-1499530172. 
It also outlines the main benefit of Juno contract approach - readability and comprehension.

Another big benefit of this approach was already mentioned, but is worth mentioning again - **everyone already understands it and is comfortable with it, there's no need to introduce yet a new design that needs to be learned and tested by time.**

misicnenad:
Closing in favor of https://github.com/AngelProtocolFinance/angelprotocol-web-app/pull/2049.

PR #2039, opened 2023-04-06
Fix: web3-auth bugs and trim deps

ap-justin:
Ticket(s): N/A
### Fix  Bugs:
 * #### user disconnecting internally from wallet is not reflected in app
   REPRO in RC-v1.8: 
     1. connect wallet via web3-auth
     2. open connected torus modal and logout
     3. wallet is still connected in dapp
  
 * #### user changing network internally from wallet is not reflected in app
   REPRO in RC-v1.8
     1. connect wallet via web3-auth
     2. open connected torus modal 
     3. switch to chain e.g `polygon-mainnet ` ( should be detected, and error since `mainnet` is not supported in testnet )

 * #### in charity registration, `web3-auth` is the only option wallet but user could still connect keplr without any error
     REPRO in RC-v1.8
     1.  go to registration with reference`fa6cb288-91b4-4227-a37a-2207415c0ed7`
     2. go to wallet submission setp
     3. connect keplr 
 
## Improvements
1. compressed `chainID`, `address` and `provider`  into `type Connected` state
2. only initialize Web3Auth if there's `cachedAdapter` (a user previously connected and didn't explicitly disconnect)
3. remove  `console.log` errors and replace with execption so error handler can catch it and show appropriate UI

## Deps changes ( ~500kb build reduction )
1. removed `web3` js as it is only used to for getting `accounts` and `chainId` which can also be done with  `request` method of `SafeEventEmitterProvider`
2. replaced `@web3auth/modal ` with `@web3auth/no-modal` + `@torus-evm-adapter`. 
3. removed all peer packages required by `@web3auth/modal`

-

## UI changes for review


src/contexts/WalletContext/useWeb3Auth/index.ts
```diff
@@ -1,112 +1,140 @@
...
-  }, [provider, chainId]);
-
-  const login = async () => {
-    setIsLoading(true);
-    const web3Auth = await init();
-    if (!web3Auth) {
-      console.log("web3auth not initialized yet");
-      return;
+    } catch (err) {
+      logger.error(err);
```
misicnenad, line 104: What kind of errors can appear here? Should they some error modal be shown to the user?

src/pages/Registration/Steps/WalletRegistration/WalletSubmission.tsx
```diff
@@ -18,13 +18,7 @@ export default function WalletSubmission({
...
   const { data } = useRegState<3>();
 
-  if (
-    !(
-      providerId === "keplr" ||
-      providerId === "keplr-wc" ||
-      providerId === "web3auth-metamask"
-    )
-  ) {
+  if (!(providerId === "web3auth-torus")) {
```
misicnenad, line 21: We have a missing `try/catch` around `disconnect` call in this component (line 38), which might cause the app to bug out on the user.
Can you add it as part of this PR?

misicnenad (changes requested):

src/contexts/WalletContext/useWeb3Auth/index.ts
```diff
@@ -1,112 +1,147 @@
...
     try {
-      await web3auth.logout();
-    } catch (e) {
-      console.log(e);
+      await web3Auth.logout({ cleanup: true });
+      web3Auth.provider?.off("chainChanged", handleChainChange);
+      web3Auth.provider?.off("accountsChanged", handleAccountsChange);
+      setState({ status: "disconnected" });
+    } catch (err) {
+      throw new Error("Failed to disconnect from wallet");
```
misicnenad, line 119:
```suggestion
      logger.error(err);
      throw new Error("Failed to disconnect from wallet");
```

misicnenad: Throwing a generic error here is OK, but we still need to handle it in [WalletSubmission > line 38](https://github.com/AngelProtocolFinance/angelprotocol-web-app/blob/trim-web3auth/src/pages/Registration/Steps/WalletRegistration/WalletSubmission.tsx#L17).
The "handling" should probably be logging + displaying the error using the `ErrorContext.handleError`.

misicnenad, line 119: We are also missing error handling logic (`try/catch` + `ErrorContext.handleError`) in [WalletSuite > DisconnectBtn](https://github.com/AngelProtocolFinance/angelprotocol-web-app/blob/trim-web3auth/src/components/WalletSuite/ConnectedWallet/Details/index.tsx#L76) :open_mouth:

misicnenad (changes requested):

src/contexts/WalletContext/useWeb3Auth/index.ts
ap-justin: you already created PR for this so we could resolve this

misicnenad (approved):

PR #2052, opened 2023-04-11
catch file upload API failures to avoid update profile crash

abbasali1994:
Ticket(s):
- [Bug: Content clears on edit profile if the update fails (should not happen)#2031](https://github.com/AngelProtocolFinance/angelprotocol-web-app/issues/2031)

The upload API failure leads to abrupt crash of the update function
Adding a catch clause avoids that crash (The issue is intermittent and only occurs cause of network issues )
 
-

## UI changes for review

### DIFF
--- a/src/helpers/uploadFiles.ts
+++ b/src/helpers/uploadFiles.ts
@@ -26,7 +26,7 @@ export async function uploadFiles(
           fileName: `${timeStamp}-${f.name.replace(SPACES, "_")}`,
         }),
         headers: { authorization },
-      })
+      }).catch((e: any) => console.log(e))
     )
   );
 


src/helpers/uploadFiles.ts
```diff
@@ -26,7 +26,7 @@ export async function uploadFiles(
           fileName: `${timeStamp}-${f.name.replace(SPACES, "_")}`,
         }),
         headers: { authorization },
-      })
+      }).catch((e: any) => console.log(e))
```
ap-justin, line 29: >  The upload API failure leads to abrupt crash of the update function

@abbasali1994, I tried to reproduce this error by just throwing error inside `uploadFiles.ts`

```ts
export async function uploadFiles(
  files: File[],
  bucket: Bucket
): Promise<string | null> {
  throw new Error("throw error thrown from upload files");
```

and error is catched by caller as expected ( no crash )
image

also edit profile content is not cleared as stated in issue #2031

abbasali1994: Like I mentioned earlier the issue seems intermittent , even I was able to reproduce it only once after multiple tries, this was just a safeguard I thought would be best to avoid that, but based on your input the catch clause does seem redundant, I will put the PR in draft for now

PR #2070, opened 2023-04-18
Admin fixes

ap-justin:
Ticket(s): n/a
1. AP admin text is hidden on lightmode
2. After voting user vote is not reflected ( was able to repro in testnet, but due to cache invalidation only - vote is still reflected on page refresh )

1. AP admin nav light text color
2. invalidate `proposal` tag on vote

-

## UI changes for review

### DIFF
--- a/src/pages/Admin/Core/Nav.tsx
+++ b/src/pages/Admin/Core/Nav.tsx
@@ -13,6 +13,6 @@ export default function Nav() {
 }
 
 const styler = createNavLinkStyler(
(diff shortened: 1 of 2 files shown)


SovereignAndrey (approved):

SovereignAndrey:
@ap-justin Thanks for the quick fix around the "missing" admin bar & the cache busting that was found whilst investigating the original errors with regards to gas used vs wanted gas.

PR #2088, opened 2023-04-24
AST Admin deposits

SovereignAndrey:
Ticket(s):
- https://app.clickup.com/t/865c5xmja

- Creates an Admin deposits endpoint for Charity & AST endowment members to easily contribute to their endowments directly from within the Admin area: `/admin/{id}/deposits`
- This page uses the existing Donations page to keep the code DRY 
- We also hide breadcrumbs for AST deposits for less visual clutter

## UI changes for review
image

### DIFF
--- a/src/pages/Admin/Charity/Deposits/index.tsx
+++ b/src/pages/Admin/Charity/Deposits/index.tsx
@@ -1,0 +1,16 @@
+import Content from "../../../Donate/Content";
+import { useAdminResources } from "../../Guard";
+
+export default function Deposits() {
+  const { id } = useAdminResources<"charity">();
+  const container = "dark:bg-blue-d6 border border-prim rounded max-w-lg  p-8";
+
+  return (
+    <div className="grid gap-8 justify-items-center">
+      <h2 className="text-center font-bold text-3xl -mb-2">Deposit</h2>
+      <div className={container}>
+        <Content name={"your AST"} id={id} isKYCRequired={false} />
+      </div>
+    </div>
+  );
+}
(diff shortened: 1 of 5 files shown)


src/pages/Admin/Charity/Deposits/index.tsx
```diff
@@ -0,0 +1,16 @@
+import Content from "../../../Donate/Content";
```
ap-justin, line 1: should move `Donate/Content` to `components/donation` as it is used now by two pages/features 

then classes that is more related to its external be also factored out and let bounding container provide it 


e.g
in `/donate`
```jsx 
<Content classes="justify-self-center padded-container scroll-mt-6 max-w-[35rem]"/> 
```

and in `admin/deposit` like
```jsx
<Content classes="justify-self-center dark:bg-blue-d6 max-w-lg"/> 
```

also breakpoints be updated from window based  e.g `sm:` to container counterpart `@sm:` as  `donate` is full page while `admin/deposit` is part of the admin layout 

These changes could be done on dedicated PR as right now `donate/{id}` doesn't show up and can't be visually inspected 
image

ap-justin (approved):

src/pages/Admin/Charity/Deposits/index.tsx
SovereignAndrey: Thanks @ap-justin! Will start making some of these changes as part of the in-progress work to show fiat-only flow, for when no wallet is connected. You can see the `/donate/{id}` example for ASTs with a published profile. 
A working example: https://rc-v1-8.d2f5hbetruke97.amplifyapp.com/donate/21

PR #2105, opened 2023-04-25
Donate refactor v1.8

SovereignAndrey:
Ticket(s):
- https://app.clickup.com/t/865c6g1ku
- https://app.clickup.com/t/865c6t0p8

- DRY up some of the code that is repeated for FiatDonate vs Donate
- Shows the Fiat ONLY Donation flow when a user lands on the page without a wallet connected

Profile > Donate: `/donate/21`
Admin > Donate: `/admin/21/donate`

## UI changes for review
None


src/components/donation/Steps/CurrentStep.tsx
```diff
@@ -23,6 +21,30 @@ export default function CurrentStep(props: ConfigParams) {
     !wallet && dispatch(resetDetails());
   }, [wallet, dispatch]);
 
+  // If there is no wallet connected take user the fiat-only path
+  if (!wallet) {
+    switch (state.step) {
+      case 3: {
+        return <FiatSubmit {...state} />;
```
ap-justin, line 28: @SovereignAndrey, I tried to also remove duplicate markups like `<FiatSubmit/>` , `<FiatForm/>`
in 
* #2107 

the difference is `country` field is below `amount` when `token === "fiat"`

image

removed from summary

SovereignAndrey: @ap-justin 
> the difference is country field is below amount when token === "fiat"
> removed from summary

What was the reason it could not stay on the summary page and use the payment type to toggle if it's visible or not? 

Judging by the screenshots above, seems like if a user switches to Fiat payment, then that field pops in and would disappear if they decided to choose a crypto payment? :thinking:

ap-justin: > What was the reason it could not stay on the summary page and use the payment type to toggle if it's visible or not?

I noticed that summary page has been duplicated just to add `country` field
image


also, putting this field below `token` makes it somehow connected to it 
like if `usd` provide `country`

image

> Judging by the screenshots above, seems like if a user switches to Fiat payment, then that field pops in and would disappear if they decided to choose a crypto payment

Yup, if crypto payment (wallet is connected), the `tokens` to select is `cryptokens + fiats`
image

if fiat payment (no connected walelt), the `tokens` to select is `fiats` only
image

SovereignAndrey: @ap-justin Okay then. We'll go with that for now. Hopefully, product team is OK with those visual changes. :crossed_fingers:

ap-justin: 🤞 will open draft PR then

ap-justin: 🤞 will open draft PR then

ap-justin: 🤞 will open draft PR then

PR #2113, opened 2023-04-26
Donate: Skip donation KYC step completely if required

misicnenad:
Ticket(s):
- https://app.clickup.com/t/865c6tp04

- connect wallet
- start donation process for some AST that has KYC step set as skippable
- verify KYC step is skipped completely
- verify for same AST (with KYC to be skipped) that clicking "Back" on "Submit" step (no. 3) navigates the user to step 1 (skips KYC as well)

### DIFF
--- a/src/components/donation/Steps/Progress.tsx
+++ b/src/components/donation/Steps/Progress.tsx
@@ -2,20 +2,24 @@ import { useGetter } from "store/accessors";
 import { PAYMENT_WORDS, titleCase } from "constants/env";
 
 export default function Progress({ classes = "" }: { classes?: string }) {
-  const step = useGetter((state) => state.donation.step);
+  const donationState = useGetter((state) => state.donation);
+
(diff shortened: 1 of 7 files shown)


SovereignAndrey (changes requested):
@misicnenad Looks like it skips the KYC step for an AST on their profile > donation page, but NOT for their Admin > donate  page?
image

ap-justin (dismissed):

SovereignAndrey:
>  Saw it contained only some dummy data (`name={"your AST"}` etc.), so set to have KYC show there by default, but can change it if that's the component in question.

Yes, that's it. The dummy data bit is actually just my attempt to shorten things and simplify the UI a bit, since a user viewing that is already in the Admin area, where we show the AST name in the top of the submenu, there's really no need to repeat it in the donate window view.

SovereignAndrey (approved):

PR #2125, opened 2023-04-28
Add connect wallet page in Launchpad

abbasali1994:
Ticket(s):
- https://app.clickup.com/t/865c61dnb

Since the step is more of a check step to ask user to connect to wallet before creating AST, there is no input from user getting stored on that page
The page comes just before the Summary page

-

## UI changes for review

### DIFF
--- a/src/pages/Launchpad/ConnectWallet/ChooseWallet/index.tsx
+++ b/src/pages/Launchpad/ConnectWallet/ChooseWallet/index.tsx
@@ -1,0 +1,25 @@
+import { Link } from "react-router-dom";
+import { steps } from "pages/Launchpad/constants";
+import WalletConnector from "./WalletConnector";
+
+export default function ChooseWallet() {
+  return (
+    <div className="w-full grid">
+      <h3 className="text-lg">Choose a wallet</h3>
+      <p className="mb-8 text-sm text-gray-d1 dark:text-gray mt-2">
+        We recommend using a new wallet.
+      </p>
+      <WalletConnector name="Keplr" label="Socal login or email" />
+      <WalletConnector name="Web3 Auth" label="Socal login or email" />
+      <WalletConnector name="Metamask" label="Metamask" />
+      <div className="grid grid-cols-2 md:flex gap-3 items-center mt-8">
+        <Link
+          to={`../${steps[6]}`}
+          className="min-w-[8rem] btn-outline-filled btn-reg"
+        >
+          Back
+        </Link>
+      </div>
+    </div>
+  );
+}
--- a/src/pages/Launchpad/ConnectWallet/ChooseWallet/index.tsx
+++ b/src/pages/Launchpad/ConnectWallet/ChooseWallet/index.tsx
@@ -9,7 +9,7 @@ export default function ChooseWallet() {
       <p className="mb-8 text-sm text-gray-d1 dark:text-gray mt-2">
         We recommend using a new wallet.
       </p>
-      <WalletConnector name="Web3 Auth" label="Socal login or email" />
+      <WalletConnector name="Web3 Auth" label="Social login or email" />
       <WalletConnector name="Metamask" label="Metamask" />
       <div className="grid grid-cols-2 md:flex gap-3 items-center mt-8">
         <Link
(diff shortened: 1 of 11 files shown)


SovereignAndrey (changes requested):
@abbasali1994 AST launchpad can't use the keplr wallet for connecting to Polygon. Please remove that as an option. Thanks! 
image

ap-justin:
> Since the step is more of a check step to ask user to connect to wallet before creating AST, t

If this is the case: then after finishing this step and now in summary page. If I disconnect my wallet, I should be reverted back to this step 

( i'm still on summary page, after disconnecting wallet ) 
image

currently, even without this step, the user still needs to connect a wallet (and be in correct network ) to submit AST
image
image



> there is no input from user getting stored on that page

@SovereignAndrey, I'm not sure why we have `connect wallet` step but we don't use any `data` (e.g. `wallet address`) from that step
  * would that `wallet address` be used as default member of multisig? 
  * be included by default in any whitelists ?
  * what if user connects `walletA` and in summary disconnects and use another `walletB`
 
 no changes have been made to make use of this new step, `sender wallet` is still used to populate most fields
 
 
 @SovereignAndrey, @abbasali1994, IMO,  connect wallet step makes sense in `charity-registration` as we're submitting via `aws`- but in AST, user interacts with the contract directly and is just natural to connect wallet at the end to submit the transaction. If many users, necessitates this step - then we should resolve some ux inconsistencies mentioned above

SovereignAndrey:
> IMO, connect wallet step makes sense in `charity-registration` as we're submitting via `aws`- but in AST, user interacts with the contract directly and is just natural to connect wallet at the end to submit the transaction. If many users, necessitates this step - then we should resolve some ux inconsistencies mentioned above

@ap-justin The point of the wallet connect step is a courtesy to new AST users that might not be familiar with web3 paradigms like needing to have a wallet connected. If they are a savvy web3 user and already have a wallet connected then it will show on that step as connected. If they don't have one, then most importantly we can offer a smoother flow towards them setting up a web3auth social-based wallet (ie. less friction).

> If this is the case: then after finishing this step and now in summary page. If I disconnect my wallet, I should be reverted back to this step

That's a bit of an edge case, buuut in said scenario we might better have:
The final "SUBMIT" button on summary step changes into a "CONNECT WALLET" button when a wallet is not connected, which should point to the previous step connect wallet page. Less jarring than yanking the user to a page just because they (possibly legitimately) disconnect their wallet. 
@abbasali1994 Please implement this change mentioned. Thanks!

> @SovereignAndrey, I'm not sure why we have `connect wallet` step but we don't use any `data` (e.g. `wallet address`) from that step
> [sic]
> no changes have been made to make use of this new step, `sender wallet` is still used to populate most fields

There's no real need to populate anything based on the wallet connect step, so long as we're using correctly using the `sender wallet` which would be the wallet connected/created in the connect step page at the time of submission.

ap-justin:
@SovereignAndrey 
> There's no real need to populate anything based on the wallet connect step,

got this, clarifies wallet-connect-step doesn't affect previous steps in any way


> That's a bit of an edge case, buuut in said scenario we might better have:
The final "SUBMIT" button on summary step changes into a "CONNECT WALLET" button when a wallet is not connected, which should point to the previous step connect wallet page. 

this is good idea 👍

SovereignAndrey (dismissed):

src/pages/Launchpad/ConnectWallet/ChooseWallet/index.tsx
```diff
@@ -0,0 +1,24 @@
...
+import WalletConnector from "./WalletConnector";
+
+export default function ChooseWallet() {
+  return (
+    <div className="w-full grid">
+      <h3 className="text-lg">Choose a wallet</h3>
+      <p className="mb-8 text-sm text-gray-d1 dark:text-gray mt-2">
+        We recommend using a new wallet.
+      </p>
+      <WalletConnector name="Web3 Auth" label="Socal login or email" />
```
SovereignAndrey, line 12:
```suggestion
      <WalletConnector name="Web3 Auth" label="Social login or email" />
```

SovereignAndrey (approved):

ap-justin (approved):

PR #2127, opened 2023-05-01
Hardcoded chainid fix for AWS queries

abbasali1994:
Ticket(s):
-

-

## UI changes for review

### DIFF
--- a/src/pages/Donations/useDonations.ts
+++ b/src/pages/Donations/useDonations.ts
@@ -8,6 +8,7 @@ import {
 } from "services/apes";
 import { useSetter } from "store/accessors";
 import useDebouncer from "hooks/useDebouncer";
+import { chainIds } from "constants/chainIds";
 
 export default function useDonations() {
   const { address } = useParams<{ address: string }>();
@@ -19,7 +20,7 @@ export default function useDonations() {
 
   const [params, setParams] = useState<DonationsQueryParams>({
     id: address || "",
-    chain_id: "80001",
+    chain_id: chainIds.polygon,
   });
 
   const queryState = useDonationsQuery(params, {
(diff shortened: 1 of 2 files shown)


src/pages/Donations/useDonations.ts
```diff
@@ -19,7 +20,7 @@ export default function useDonations() {
 
   const [params, setParams] = useState<DonationsQueryParams>({
     id: address || "",
-    chain_id: "80001",
+    chain_id: chainIds.polygon,
```
SovereignAndrey, line 23: We need to detect the chain that the user's wallet is connected to here, since they (unlike the Endowments who only receive on Polygon) can send donations from other chains.

SovereignAndrey (changes requested):

src/pages/Donations/useDonations.ts
abbasali1994: understood, thanks

SovereignAndrey: @abbasali1994 What's the status on this? Thanks!

abbasali1994: I have debugged this and found the following:
this param chain_id when set to polygon returns all the donations even if not of polygon. the filter tag within the donation page is responsible for filtering of donations as per selected chain in filter dropdown , the following is not connected with wallet in any sense (The address also is picked from query param and not from connected wallet)

If we try to make it such that it only returns donations of the connected network on first render it might become confusing for end user as the filters section also provides a network select option which would conflict with the network we select in wallet section.

Modifying "80001" -> chainIds.polygon should take care of test and prod deployment queries

SovereignAndrey (approved):

PR #2169, opened 2023-05-09
Migrate giftcard

ap-justin:
Ticket(s):
-

## UX Changes
 - when purchasing giftcard via `erc20` there's a need to add allowance for the giftcard contract for spending of such `erc20`
 
image

- gift card balances appear as standalone tokens (for reasons below) 

image

- in `juno`, we can send `BankSendMsg` (to send `juno`) and `ExecuteContractMsg` (to send `spendMsg` to giftcard) at the same time which matches UI in  
  * #1600 
  
this is not the case with `polygon (evm)` so gift token is standalone (to send `spendMsg` only)
image

NOTE: 
  * spending `native` gift-token doesn't work due to this issue 
    * https://github.com/pedalsup/Angel-protocol/issues/63
  * purchasing gift-card without `recipient`  is still WIP and waiting for 
    * https://github.com/AngelProtocolFinance/devops/issues/232
     

-

## UI changes for review


src/services/juno/queryContract/queryObjects.ts
```diff
@@ -21,15 +22,10 @@ import { multisig } from "contracts/evm/multisig";
 import { registrar } from "contracts/evm/registrar";
 import { toTuple } from "helpers";
 
-type MigrationState =
```
ap-justin, line 24: all queries are now migrated to polygon so no need for this migration state

PR #2181, opened 2023-05-16
Meta for nested objects

ap-justin:
Ticket(s):
determining diffs for nested objects is already handled by `getPayloadDiff` 
```ts
type Differ = (prev:T, next:T) => Partial<T>
```
however, showing to user what changed on this nested object brings a lot of problem
* UI is coupled to the nested object type which means for each nested object type corresponding flat type should be created
```ts
type Payload  = {
  settings1: {
    address: string;
    isTrue: boolean;
     }
}

/** to render to a table ( flat `key | prev | next ) ,  need to create flat type */
type FlatPaylaod = {
  settings1NewAddress: "new address"
  settings1isTrue: true
}
```

this approach could be problematic the deeper the nesting of a given payload

* flatten object before diffing
* since payload diff  is only `Partial<typeof prev | next>` (may be nested) and need to be transformed to a more table friendly format `[Key, Prev, Next][]` via `getDiffEntries`, bundled `formatting` to `diffing`

that is instead of 
```ts
type Differ = (prev:T, next:T) => Partial<T>
```
now
```ts
type Diff = [Key, Prev, Next]
type Differ = (prev:T, next:T) => Diff[]
```

-

## UI changes for review
any complex payloads can now be represented using table 

image


ap-justin:
included in #2186

PR #2224, opened 2023-06-14
Permissions table refactors

ap-justin:
Ticket(s):
-

* remove duplicate markup with separate `<Table/>` and `<MobileTable/>` 
* after successful update changes are immediately reflected in the form
* when unchecking `delegate`, delegate address gets deleted
* when checking `delegate`, delegate address input gets focused
* delegate address validation now contained in row error appear
image
* locking a permission - if delegate is `checked` and no address nor invalid address is provided - lock operation would focus the delegate address instead 

-

## UI changes for review


src/pages/Admin/Charity/Permissions/constants.ts
```diff
@@ -0,0 +1,10 @@
+import { TPermission } from "./types";
+
+export const keys: { [K in keyof TPermission]: K } = {
+  addr: "addr",
+  isActive: "isActive",
+  locked: "locked",
+  modifiable: "modifiable",
+  ownerControlled: "ownerControlled",
```
SovereignAndrey, line 8: This looks to be based around older code for the permissions. We should consider updating the TPermission type. Newer Solidity struct for `SettingsPermission` looks like: 
```solidity
struct SettingsPermission {
    bool locked;
    Delegate delegate;
}
  
struct Delegate {
    address addr;
    uint256 expires; // datetime int of delegation expiry
}
```
  
If we flatten those into something similar to the orig `TPermission`,  then maybe we get something like the following psudocode:
```typescript
type TPermission {
    boolean locked;
    address delegate_addr;
    integer delegate_expires;
}
```

SovereignAndrey (changes requested):
Overall, this looks good! Please see my comment regarding the TPermission type and possible improvements we can make to bring it in line with current Solidity structs.

src/pages/Admin/Charity/Permissions/constants.ts
ap-justin: @SovereignAndrey, `TPermission` is only a convenience type is intented for form use only 

after getting `SettingsPermission (types/contract)` from contract
it is first converted to `TPermission`

and when form submits - converted back to `SettingsPermission`

Most of the time, types from contracts are not directly used in forms as their structure doesn't always cater to the required UI 

sample initialization of `SettingsPermission` to `TPermission` 
```ts
export const formPermission = (setting: SettingsPermission): TPermission => {
  const delegate = setting.delegate;
  const isDelegated = delegate.addr !== ADDRESS_ZERO;
  return {
    isActive: isDelegated,
    addr: isDelegated ? delegate.addr : "",
    //what user toggles (lock / unlock / undo), before submitting
    locked: setting.locked,

    /** meta */
    // persisted locked state of the permission
    modifiable: !setting.locked,
    
    // always admin controlled
    ownerControlled: true,
    // future: delegate.addr === endowment.govAddress
    govControlled: false,
  };
};
```

sample conversion back to `SettingsPermission`
```ts
(permission: keyof TPermissions): SettingsPermission => {
    const val = permissions[permission];
    return {
      locked: val.locked,
      delegate: {
        addr: val.isActive ? val.addr : ADDRESS_ZERO,
        expires: 0, //in design: no expiry for delegation,
      },
    };
  };
```

SovereignAndrey: Got it! Thanks for walking me through that. :+1: Will look to the rest of the code with that conversion sort of setup in mind for other bits.

SovereignAndrey (approved):

PR #2238, opened 2023-06-20
Pre-submit UI restrictions

### LINKED ISSUE #2176: Admin Area: Disable submit buttons & have read only input fields for non-connected wallet users
## COS
- [x] Hide the SWITCH button on sidebar menu when: no wallet is connected OR there is less then 2 ASTs associated with the connected wallet  
- [x] When a user is on the Admin area of the site, without a wallet connected, the following UI should be seen:
    - [x] all "Submit" buttons or other "action" buttons/links should be hidden
    - [x] all input fields, where possible, should be disabled
- [x] When a user has a wallet connected, everything should be active (is it exists today)

NOTES: @ap-justin 
- [ ] ~~lock forever buttons on individual settings page~~ design not final

Image

ap-justin:
Ticket(s):
* closes #2176 

* checks now as derived state instead of read-only during `submit`

```ts
//before
function submit(){
 //read_premissions()
}

//now
const [permissionState] = usePermissionsState()

function submit(){
 //use permission state
}

//UI 
<>{use permission state on UI and reflect restrictions}</>

```
* apply UI restrictions on every multisig forms`

>  When a user is on the Admin area of the site, without a wallet connected, the following UI should be seen:
 all "Submit" buttons or other "action" buttons/links should be hidden
 all input fields, where possible, should be disabled
 
 * though issue only outlines what happens on disconnected wallet, other checks must also done 
   - wallet not connected in polygon
   - wallet not member of admin
   - for charity, wallet not member of admin, but is a delegate
   - for withdraws, wallet not member of admin, but whitelisted (beneficiary/ maturity )
   - setting is locked forever
also, adding tooltip is helpful so user would know why some action UI is missing 

### Others
* reuse new `AdminWallet` in admin/AP and admin/Review pages

-

## UI changes for review
sample form: wallet not connected
image

sample form: incorrect network 
image

not admin or delegate (also considers withdraw where bene/maturity list is checked) 
image

for maturity related forms 
image

locked forever
image


SovereignAndrey (approved):
Awesome work on this!

PR #2272, opened 2023-07-21
Programs form: milestones

ap-justin:
Ticket(s):
-

## In this PR
* going to `/admin/5/programs`, creating or editing a program directs to create/edit program form
* could click `add milestone` and new milestones is appended
image

* expanding a particular milestone show `MilestoneForm`
Screenshot 2023-07-26 at 5 16 11 PM

* user is free to expand and edit/delete any newly created or existing milestone in the list

### Works for another PR
- [ ]  render milestones in `/profile`, only rendered in program-editor
- [ ] `milestone_banner` can be video or embedded, now (image only)
image

- [ ] draggable_handles to change milestone order, @jp-mariano, I guess , rn, we could just return sorted `Milestone[]` according to `milestone_date` for the `/profile`, `/ast` and `profile/program` endpoints ? 

image

-

## UI changes for review


SovereignAndrey:
> return sorted Milestone[] according to milestone_date
@ap-justin @jp-mariano We should def return milestones by their date for viewing in the UIs and not some custom sorting.

@ap-justin We can drop the draggable UI part too, as I don't think it even really makes sense to have draggable milestones in the admin UI since they are date-driven entities. Only alternative that makes sense (without some admin side sort checks on save events) is just no dragging, no auto sort and we just add milestones in the Admin and worry about sorting for the rendering of Profiles. I actually like that more as a first-pass approach. Wdyt?

ap-justin:
@SovereignAndrey,
got these ser, 
If ser @jp-mariano would return sorted milestones, then no extra client side sorting needed

user could just add milestones in any order and when they submit - it would be displayed in order after refetch and re-render


I would then add the tooltip in the milestones section👍

ap-justin:
@SovereignAndrey, added milestones tooltip
image

SovereignAndrey (approved):

PR #2321, opened 2023-08-28
ap-681: Withdraw UI

ap-justin:
Ticket(s):
- 

* reflect contract changes
* update test coin to axelar usdc 
* update withdraw wiring to include endowments as beneficiary
* update fee calculations wiring to include endowments as beneficiary

### Others
* update `<CloseEndowment/>` template to test withdraw flow on closed endowments 
* convert  `<Radio/>` to uncontroller component and reuse in kyc, admin/other-settings and `registration`
* update KYC form to use uncontrolled `<Radio/>`
* reuse `<Radio/>` in withdraw beneficiary selection

-

## UI changes for review
* endowment to withdraw from is closed
image

* withdrawing from DAF
image

* withdrawing from `AST | Charity`. 
Added beneficiary type tabs
Network is moved to bottom, as it being rendered depends on beneficiary type 
image

* when beneficiary type is endowment, don't show network and breakddown ( no fees applies )
image


SovereignAndrey:
@ap-justin This looks great overall. Just a few notes on logic to check and some other things to improve the UI:
1. When an Endowment is closed now, only the final beneficary wallet/endowment will be able to pull the funds to their wallet/endowment. It's the pull vs push setup for moving funds. Therefore we need an auth check, when the endowment is closed, to see if the connected wallet is either the beneficiary wallet or is a member of the beneficiary endowment set.
  - Ex: Endowment 1 closes, setting Wallet `0x1` as the final beneficiary. Moving forward only `0x1` can withdraw the funds of Endowment 1 to that wallet. This is handled by a standard withdraw transaction sent to the Accounts contract by beneficiary wallet. 
  - Ex: Endowment 1 closes, setting Endowment 2 as the final beneficiary. Moving forward only Endowment 2 can withdraw the funds of Endowment 1 to it's own Endowment. This would be handled as a withdraw proposal message created for the Endowment 2 multisig members to approve, to withdraw from Endowment 1's account (ie. `id` = 1, beneficiary endowment id = 2).
2. For the Charity/AST Endowments tab, can we make this a list of endowment options (val: endowID, display: Name)? Similar to how the Widget endowment selection dropdown functions. Would be 100% better UI/UX than expecting the user to know the ID ahead of time. We'd need to use the AWS API though to pull the list. This can be a separate ticket/PR, as it's an overall improvement and not a blocker (see: https://linear.app/angel-protocol/issue/AP-754/withdraw-endowment-dropdown-list).

ap-justin:
@SovereignAndrey 
 > Ex: Endowment 1 closes, setting Wallet 0x1 as the final beneficiary. Moving forward only 0x1 can withdraw the funds of Endowment 1 to that wallet. This is handled by a standard withdraw transaction sent to the Accounts contract by beneficiary wallet.

this check is [implemented here](https://github.com/AngelProtocolFinance/angelprotocol-web-app/pull/2321/files#diff-fe395ba7c21f5c9eef7a15058e98389881f1fa0b595af36aee7e54c96baa2d77R161)

> Ex: Endowment 1 closes, setting Endowment 2 as the final beneficiary. Moving forward only Endowment 2 can withdraw the funds of Endowment 1 to it's own Endowment. This would be handled as a withdraw proposal message created for the Endowment 2 multisig members to approve, to withdraw from Endowment 1's account (ie. id = 1, beneficiary endowment id = 2).

the UI implementation of this could be: 

scenario: 
`admin-2-member` is in `admin/2/withdraws`
image

in the withdraw form, `admin-2-member` must have the ability to set `withdraw from endowment: 1` (`endow 2 by default` ). This would mean that `admin-2-member` must be have access to a list `AccountToWithdrawFrom[]`

where this list conists of
1. endow: 2, - own account must be included in this list
2. closed account where beneficiary is endow:2

perhaps we could create subgraph endpoint for this - utilizing `EndowmentClosed` event. 

> For the Charity/AST Endowments tab, can we make this a list of endowment options (val: endowID, display: Name)?

yes ser, this could be done in another PR

src/pages/Admin/Context.tsx
```diff
@@ -157,8 +158,16 @@ export const useAdminContext = <T extends AdminType = any>(
     operations.every((op) => {
       switch (op) {
         case "withdraw-liquid":
+          // if beneficiary is endowment, sender must be multisig
+          if (closed && closingBeneficiary.type === "wallet") {
+            return closingBeneficiary.value === sender;
+          }
           return allowlistedBeneficiaries.includes(sender);
```
SovereignAndrey, line 165: @ap-justin If the endowment is `closed` the allowlists do not apply anymore. Only the final Beneficiary applies at that point.

1. IF not closed: check send is included in appropriate allowlist (based on maturity status)
2. ELSE (ie. is `closed`): 
    - If beneficiary is a Wallet: check final beneficiary wallet == sender
    - If beneficiary is an Endowment: check sender is member of final Beneficiary Endowment multisig

SovereignAndrey:
> scenario: `admin-2-member` is in `admin/2/withdraws`
> in the withdraw form, `admin-2-member` must have the ability to set `withdraw from endowment: 1` (`endow 2 by default` ). This would mean that `admin-2-member` must be have access to a list `AccountToWithdrawFrom[]`
> 
> where this list consists of:
>     1. endow: 2, - own account must be included in this list
>     2. closed account where beneficiary is endow:2

@ap-justin That could work, though I'll need to think about the UI/UX a bit more. Don't want to overly complicate an form that's already getting pretty complex. Maybe we can only show this "withdraw source" dropdown when there's a closing endowment set for an Endowment? Assume it's the usual self-sourced withdraw otherwise?

> perhaps we could create subgraph endpoint for this - utilizing `EndowmentClosed` event.

Could work! We can also update the subgraph Endowment entity with some `state` struct details (like Closing Beneficiary), so that it is queriable in the sense of: "Find all Endowments where `Endowment.closingBeneficiary.endowmentId = <endowID>`"?

src/pages/Admin/Context.tsx
ap-justin: > If the endowment is closed the allowlists do not apply anymore. Only the final Beneficiary applies at that point.

I believe this is already the case, when determining `isDelegated` ( that is the `sender` sends `msg` directly to accounts contract )
```ts
if (closed && closingBeneficiary.type === "wallet") {
     return closingBeneficiary.value === sender;
 }
 
```
and if not closed 
```ts
return allowlistedBeneficiaries.includes(sender);
```

The flow is like, a random `user-x` goes to `admin/2` it is closed, and user is `closingBeneficiary wallet` so could use this withdraw form to send `withdraw()` directly to accounts contract. This is assuming that `user-x`knows beforehand that `endow:2` is closed and `user-x` is beneficiary. 

Another flow would be in wallet dropdown, a particular `user-x` has access to some  list `ClosedAccount[]`
where `user-x` is the beneficiary. 
* `endow/13`
* `endow/5`
* `endow/6`

if `user-ex` visits one of these endowments (e.g `endow/13`), `isDelegated` and therefore can send `withdraw()` directly to accounts contract. 


> If beneficiary is an Endowment: check sender is member of final Beneficiary Endowment multisig

for when `closingBeneficiary === endowment`, this isn't the place to check it as this whole admin `<Context/>`  is scoped to a particular admin. e.g. if this is admin `2`, all checks must all just pertain to whether `user-x` is member of `admin-2` or is delegated to perform transactions particular to `endow-2`. 

there's even no need to check depending on how we implement the UI. e.g if I am `admin-2-member` and I am on `/admin/2`, any multisig transaction I send is `authorized` (e.g `withdraw()` ) - given that a list of `CloseEndowmentWhereAdmin2MustBeSender[]` readily available. 

as  `admin-2-member`, I could send 
```
submitTransaction 
withdraw {
  beneficiaryEndow: 13
}
//or 
submitTransaction 
withdraw {
  beneficiaryEndow: 5
}
//or 
submitTransaction 
withdraw {
  beneficiaryEndow: 6
 }
```
as it is already checked that `multisig-2` should be sender of such withdraw msgs

SovereignAndrey: > there's even no need to check depending on how we implement the UI. e.g if I am admin-2-member and I am on /admin/2, any multisig transaction I send is authorized (e.g withdraw() ) - given that a list of CloseEndowmentWhereAdmin2MustBeSender[] readily available.

You're right! Thanks for that. This was the part I was missing while walking over the logic. Alright, I'll merge this in now and the UI implementation of this can be handled in a smaller/separate PR. I think if we start with "withdraw source" only shown when needed then it's a decent POC and we'll refine from there with Operational team feedback. Will try to get the subgraph stuff updated no later than EOD or early tomorrow.

SovereignAndrey: PS. Yes! This is 100% valid use case!

> Another flow would be in wallet dropdown, a particular user-x has access to some list ClosedAccount[]
where user-x is the beneficiary.
> If user-ex visits one of these endowments (e.g endow/13), isDelegated and therefore can send withdraw() directly to accounts contract.

PR #2331, opened 2023-09-04
AP-756: Donate fees

ap-justin:
Ticket(s):
-

* add new `coingecko_denom` field on `type Token`
* refactor `estimateDonation` helper to include usd values of items
* refactor `<Breakdown/>` to make use of new format estimate format

image

### Others
* hide `admin/deposit` for non-ast endowments per https://github.com/AngelProtocolFinance/angelprotocol-web-app/pull/2088

-

## UI changes for review


src/components/donation/Steps/Submit/index.tsx
```diff
@@ -137,22 +132,25 @@ function TxTotal({
...
+                {item.cryptoAmount ? (
+                  <div className="grid justify-items-end">
+                    <span>
+                      {humanize(item.cryptoAmount.value, 4)}{" "}
+                      {item.cryptoAmount.symbol}
+                    </span>
+                    <span className="text-xs"> {item.prettyFiatAmount}</span>
+                  </div>
+                ) : (
+                  `~${item.prettyFiatAmount}`
```
SovereignAndrey, line 150:
```suggestion
                  `${item.prettyFiatAmount}`
```

SovereignAndrey (dismissed):
@ap-justin Looks superb! :+1: Let's just drop the tilde (`~`) to give it a cleaner feel. We say "Estimated" for proceeds, so it's implied it's all an estimation already.

src/components/donation/Steps/Submit/estimateDonation.ts
```diff
@@ -113,7 +137,65 @@ export async function estimateDonation({
...
+        value: txEstimate.fee.amount.toString(),
+        symbol: txEstimate.fee.symbol,
+      },
+      fiatAmount: feeUSDValueAmount.toNumber(),
+      prettyFiatAmount: `$${humanize(feeUSDValueAmount, 4)}`,
+    };
+
+    const { isFiscalSponsored, endowType } = recipient;
+    const angelGivingFeeRatePCT =
+      endowType === "charity" ? (isFiscalSponsored ? 5.8 : 2.9) : 0;
```
SovereignAndrey, line 175: Just noticed that there's a bug here wherein if the endowment is not a Charity then the the estimate will be wrong (0% for fee). This is my fault for not explaining the fee structure better in the ticket. 

The base fee (contract-level) of 1.5% is always applied. The crypto fee of 1.4% is applied for non-fiat donations to all types. Only the FS Fee of 2.9% is relevant to Charity types only. Here's what the fee total logic would roughly look like building up from a base fee:

```javascript
let baseFee = 1.5;
if (!fiat) {
  baseFee += 1.4;
}
if (isFiscalSponsored) { 
  baseFee += 2.9;
}
```

SovereignAndrey (changes requested):

src/components/donation/Steps/Submit/estimateDonation.ts
ap-justin: @SovereignAndrey, instead of `AngelGiving Fee` what would the fees be called for ASTs ?

SovereignAndrey: > @SovereignAndrey, instead of `AngelGiving Fee` what would the fees be called for ASTs ?

@ap-justin I'd say `AngelImpact Fee`.

> update: also for fiat donations where currency is `EUR` | `GBP` | `AUD`, is it needed to convert first to `USD` before applying `1.5%` base fee ?

Yes, good catch! CoinGecko quotes are all in USD I think, so it would reason we'd need to either:
1. Convert to their non-USD fiat value to USD OR 
2. Have an FX rate to turn the token prices into the rough EUR/GBP/etc equivalent 

I would say Option 2 is better UX since they can see everything in their native currency still.

Fiat donation: 100 EUR
FX Rate (EUR/USD): 0.98
Token price: 1.00 USD
Token price (~EUR): 0.98 EUR
Fees can show in EUR as they are % based

This can all come in a separate PR though once we add Stripe in. :)

PR #2386, opened 2023-09-19
FIX: applications table 

ap-justin:
Ticket(s):
-

* remove unused registration status in context of applications
* remove `all` as this gets data from `registration/list` which some records are not complete (on-going registration)
* move query lower down tree so that `no found` wont hide the table including the `status selector`
image

* show check on selected status instead of hiding selected option
image

-

## UI changes for review


jp-mariano:
The data for `under-review` is ideal in this situation because the data is complete. For `active` and `inactive`, this will list all past registrations that have those statuses. Unlike in `under-review` where we can pick only `mainnet` applications due to the existence of `chain_id` field, `active` and `inactive` does not have a `chain_id` field. For `approved` and `pending`, this can be removed since it's not being used as a status in AWS.

The question here is if we really want to list `inactive` applications (because it can be spammed now that we don't check the email) and `active` applications (since they're already an endowment, the best source of data here is the `endowments_v2` table)?

SovereignAndrey (approved):
Thank you for fixing this! :heart:

PR #2388, opened 2023-09-20
temp legacy notice banner

SovereignAndrey:
Ticket(s):
- N/A

Add a temporary banner letting users know of the old legacy site URL & the sunset deadline. 

-

### Light theme
scrnsht_banner

### Dark theme
scrnsht_banner_drk

### DIFF
--- a/src/App/Header/index.tsx
+++ b/src/App/Header/index.tsx
@@ -1,7 +1,6 @@
 import { useEffect, useRef, useState } from "react";
 import { Location, matchRoutes, useLocation } from "react-router-dom";
 import { Link } from "../types";
-import ExtLink from "components/ExtLink";
 import Image, { ImageProps } from "components/Image";
 import WalletSuite from "components/WalletSuite";
 import { appRoutes } from "constants/routes";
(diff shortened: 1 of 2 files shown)


src/App/Header/index.tsx
```diff
@@ -35,27 +36,42 @@ export default function Header({ classes, links, logo }: Props) {
   const mb = hasBanner(location) ? "-mb-[6.5rem]" : "mb-0";
 
   return (
-    <header
-      className={`${classes} ${isScrolled ? "shadow-lg" : ""} ${bg} ${mb}
+    <>
```
ap-justin, line 39: @SovereignAndrey, the overall page layout is 
```
grid auto_1fr_auto
header
-
-
-
page
-
-
-
footer
```
with the header now being two elements enclosed by `<></>

the layout becomes

```
grid auto_1fr_auto
notice
header
-
-
-
body
-
-
-
footer
```
causing other pages to break
image


Solution would be to change `src/App/Layout.tsx` grid layout to `auto_auto_1fr_auto` to account for the notice in inside the `<Header/>` 

this change however would make such layout not obvious
as it is expected that the child is `4` elements, to make it more evident
it's probably best to just put `Notice` in `src/App/Layout.tsx`
`i.e.`

```
div grid auto_auto_1fr_auto
  --Seo (not rendered)
  --{...insert notice here..} (auto)
  --Header (auto)
  --Body (1fr)
  Footer (auto)
 div
```

ap-justin (changes requested):

src/App/Header/index.tsx
SovereignAndrey: @ap-justin I hate to make big layout changes for a hotfix that will only exist for 3 weeks before being removed. :thinking: What if we only render that on the marketplace page? `if route == marketplace ? show : return nothing`

src/App/Layout.tsx
```diff
@@ -14,13 +15,22 @@ const { HEADER_LINKS, GROUPS_DATA, SOCIAL_MEDIA_LINKS } = IS_AST
...
   return (
-    <div className="grid grid-rows-[auto_1fr_auto]">
+    <div className="grid grid-rows-[auto_auto_1fr_auto]">
       <Seo /> {/* Load all defaults for SEO meta tags */}
-      <Header
-        classes="sticky top-0 z-20"
-        links={HEADER_LINKS}
-        logo={LOGO_DARK}
-      />
+      <div className="top-0 text-center dark:bg-blue bg-blue-d3">
```
ap-justin, line 20:
```suggestion
      <div className="text-center dark:bg-blue bg-blue-d3">
```

src/App/Layout.tsx
```diff
@@ -14,13 +15,22 @@ const { HEADER_LINKS, GROUPS_DATA, SOCIAL_MEDIA_LINKS } = IS_AST
...
+            key="banner-link-legacy"
+            className="transition ease-in-out duration-300 hover:text-orange-l1"
+            href="https://legacy.angelgiving.io"
+          >
+            https://legacy.angelgiving.io
+          </ExtLink>{" "}
+          and will be sunset on Oct 13, 2023. Migrate to Polygon today!
+        </span>
+      </div>
+      <Header classes="sticky z-20" links={HEADER_LINKS} logo={LOGO_DARK} />
```
ap-justin, line 33:
```suggestion
      <Header classes="top-0 sticky z-20" links={HEADER_LINKS} logo={LOGO_DARK} />
```

`top-0` should stay in `Header` so that it remains to stick on top of the page
image

this would also mean, `Notice` is hidden when user scrolls down

ap-justin (changes requested):

PR #2407, opened 2023-10-02
apes endpint

ap-justin:
Ticket(s):
-

-

## UI changes for review


SovereignAndrey (commented):
@ap-justin Is there no way to have the "Donate with card" button at the top as an option even if a crypto wallet is connected? I imagine the stripe button could just override the crypto flow path and send a donor out to pay on stripe anyways (once a split is picked, but cancelling the split pick could take you back to the screen with both options)?

ap-justin:
> @ap-justin Is there no way to have the "Donate with card" button at the top as an option even if a crypto wallet is connected? I imagine the stripe button could just override the crypto flow path and send a donor out to pay on stripe anyways (once a split is picked, but cancelling the split pick could take you back to the screen with both options)?

@SovereignAndrey, I have to further clarify overall expected interaction
 with this given form: 
image

1. location of some `<DonateToCardBtn/>`, do you mean just below the text `Connect the wallet of your choice to donate crypto.` ? 
> at the top as an option even if a crypto wallet is connected? 

2. if this `<DonateToCardBtn/>` is clicked, should the `amount field` be `disabled`  or be `replaced` by a tooltip that says `Donation amount to be specified in checkout page`

3. I would assume that the `Please send me a tax receipt` checkbox be also disabled/hidden

SovereignAndrey:
If we can re-work the setup so that something like this is possible (ignoring the color scheme. I was fiddling with HTML from GlobalGiving.org to get the UI look/feel), where Fiat is the default first option and the crypto option can be shown when clicked
donation_options

ap-justin:
> @ap-justin Maybe we can re-work the setup so that something like this is possible (ignoring the color scheme and exact copy at the bottom part, as I was fiddling with HTML from GlobalGiving.org to get the UI look/feel), where Fiat is the default first option and the crypto option can be shown when clicked donation_options

done implementing this changes ser ☝️

SovereignAndrey (approved):
Looks great!

PR #2437, opened 2023-10-19
AP-830: remove envs

ap-justin:
Ticket(s):
-
* other env variables to be removed on other PR as involves work AWS side

-

## UI changes for review


src/constants/common.ts
```diff
@@ -1,30 +1,3 @@
-import { EndowDesignation } from "types/aws";
-import { IconType } from "components/Icon";
-import { ImageProps } from "components/Image";
```
ap-justin, line 3: as much as possible, `/constants` should be top-level, just imports from `./types` or third party libraries. This would be helpful if we somehow decide to package `/constants` folder

SovereignAndrey: Agreed, unless there's a good reason to split up a logical grouping of a larger number of constants. :+1:

SovereignAndrey (approved):

PR #2439, opened 2023-10-20
Towards AP-834: Admin: Banking page posts bank details to AWS endpoint and can show `bank_verified` status on that page

misicnenad:
1. This PR is just the "functional" part of the page, so the UI is not pleasing to the eye. The UI will be part of a separate PR.
2. This PR uses a temporary placeholder logic for fetching form fields from Wise (see `# Important` section below)

Reasoning: putting all the necessary changes into one PR would make it tough to review effectively, so I split the changes into *functional* and *UI* parts (PR in progress). 

## Important
**We're experiencing a blocker where Wise API token doesn't really work, so all Wise requests fail with `error: "invalid_token"`.
We'll be able to test the real logic out only when this is solved.**

- register if necessary
- sign in
- go to http://localhost:4200/admin/1/banking
- verify no bugs are present

## UI changes for review

Prior to loading fields:
image

Valid form:
image

Errors:
image

### EDIT: 2023-10-30
Added "Check requirements" button when requirement refresh is needed, see https://github.com/AngelProtocolFinance/angelprotocol-web-app/pull/2439#discussion_r1376041702:
image

### EDIT: 2023-10-31
Added Bank Verification status component:
- "Not Submitted":
image

- "Under Review":
image

- "Approved":
image

- "Rejected":
image

- No bank verification present (shouldn't happen):
image


src/components/Selector/Selector.tsx
```diff
@@ -1,16 +1,18 @@
...
-import { FieldValues, Path, get, useController } from "react-hook-form";
+import { FieldValues, get, useController } from "react-hook-form";
 import { OptionType, Props, ValKey } from "./types";
 import { DrawerIcon } from "components/Icon";
 import FocusableInput from "./FocusableInput";
 import { styles, valueKey } from "./constants";
 
-export function Selector<
-  T extends FieldValues,
-  K extends Path<T>,
```
misicnenad, line 11: Had to remove the 2nd generic parameter to allow for dynamic field names in the bank details form, see how it's [used in RequirementField](https://github.com/AngelProtocolFinance/angelprotocol-web-app/pull/2439/files#diff-c7b73ee138eaa5d5f7f49f7b6fbc0b853f997bfdfdd3718d5209a1fce98666daR50)

src/components/KYC/Form/index.tsx
```diff
@@ -94,7 +94,7 @@ export default function Form({ classes = "", ...props }: Props) {
           <Label htmlFor="usState" className="mb-2" required={false}>
             State
           </Label>
-          <Selector<FV, "usState", string>
```
misicnenad, line 97: See https://github.com/AngelProtocolFinance/angelprotocol-web-app/pull/2439/files#r1371279109

src/pages/Admin/Charity/Banking/RecipientDetails/dot.ts
misicnenad, line 1: Read the comment to understand the reason for using these helpers.
If you think of a better way to handle this problem, let me know and we can work something out.

src/pages/Admin/Charity/Banking/RecipientDetails/RecipientDetailsForm/Form.tsx
```diff
@@ -0,0 +1,65 @@
...
+  formValues: FormValues
+): CreateRecipientRequest {
+  return {
+    accountHolderName: formValues.accountHolderName,
+    currency: formValues.currency,
+    type: formValues.type,
+    details: Object.entries(formValues.requirements).reduce<
+      CreateRecipientRequest["details"]
+    >((details, [key, value]) => {
+      details[redot(key)] = typeof value === "object" ? value.value : value;
```
misicnenad, line 61: For explanation on `undot/redot` see https://github.com/AngelProtocolFinance/angelprotocol-web-app/pull/2439/files#r1372615657

src/pages/Admin/Charity/ProgramEditor/Milestones/Milestone.tsx
```diff
@@ -61,7 +61,7 @@ export default function Milestone({
           classes={{ input: "date-input uppercase", container: "field-admin" }}
           name={`milestones.${idx}.milestone_date`}
           label="Date of milestone"
-          placeholder="e.g. John"
+          placeholder="e.g. 2014-09-23"
```
misicnenad, line 64: misleading placeholder

src/schemas/date.ts
misicnenad, line 1: unused

src/services/aws/bankDetails/bankDetails.ts
```diff
@@ -0,0 +1,19 @@
+import { WiseRequest } from "types/aws";
+import { aws } from "services/aws/aws";
+import { version as v } from "../../helpers";
+
+export const bank_details_api = aws.injectEndpoints({
+  endpoints: (builder) => ({
+    /**
+     * The endpoint returns different types of results depending on the requests sent, but since it's impossible
+     * to specify a generic return argument here, we return `unknown` and cast the return value manually
```
misicnenad, line 9: For the casting logic, see [useTypedWiseMutation](https://github.com/AngelProtocolFinance/angelprotocol-web-app/pull/2439/files#diff-ffd7f4b3dc2dac7b058b975c84453fb7bb7badead212d7afa60b65884f6f4c88R18)

src/pages/Admin/Charity/Banking/Banking.tsx
```diff
@@ -0,0 +1,67 @@
...
+        </label>
+        <input
+          id="amount"
+          type="number"
+          onChange={(event) => setExpectedFunds(Number(event.target.value))}
+          className="field-input"
+        />
+      </div>
+
+      {!!targetCurrency && !!debouncedExpectedFunds && (
```
misicnenad, line 56: Debouncing the input value (see [useBanking.ts](https://github.com/AngelProtocolFinance/angelprotocol-web-app/pull/2439/files#diff-3124142d81fc04300e4adeef06cfd1a62dc84e080341f3dc827489a150e74417R15)) and loading the form afterwards was a creative choice, so that there's no need for an additional button in the UI (and an additional mouse click for the user).
It is logical to expect the user to input this value once or twice, so the UX should be ok.
What do you think @SovereignAndrey @ap-justin?

src/schemas/date.ts
misicnenad, line 1: unused

src/pages/Admin/Charity/Banking/RecipientDetails/dot.ts
misicnenad, line 1: Read the comment to understand the reason for using `undot/redot` helpers.
This works, but if you think of a more elegant way to handle this problem, let me know and we can work something out.

src/pages/Admin/Charity/Banking/CurrencySelector.tsx
```diff
@@ -0,0 +1,97 @@
...
+        </Combobox.Button>
+
+        <Transition
+          as={Fragment}
+          leave="transition ease-in duration-100"
+          leaveFrom="opacity-100"
+          leaveTo="opacity-0"
+          afterLeave={() => setQuery("")}
+        >
+          <Combobox.Options className="absolute top-full mt-2 z-10 w-full bg-white dark:bg-blue-d6 shadow-lg rounded overflow-y-scroll scroller text-base ring-1 ring-black ring-opacity-5 focus:outline-none sm:text-sm">
```
ap-justin, line 72:
```suggestion
          <Combobox.Options className="absolute top-full mt-2 z-10 w-full bg-white dark:bg-blue-d6 shadow-lg rounded overflow-y-scroll {fixedHeight} scroller text-base ring-1 ring-black ring-opacity-5 focus:outline-none sm:text-sm">
```

should add fixed height for scroll to take effect

ap-justin (changes requested):

src/pages/Admin/Charity/Banking/useBanking.ts
```diff
@@ -0,0 +1,87 @@
...
+import { EMAIL_SUPPORT } from "constants/env";
+import { Currency } from "./CurrencySelector";
+
+const DEFAULT_TARGET_CURRENCY = "USD";
+const ERROR_MSG = `Error loading currencies. Please try again later. If the error persists, please contact ${EMAIL_SUPPORT}.`;
+
+export default function useBanking() {
+  const [targetCurrency, setTargetCurrency] = useState<Currency>();
+  const [expectedFunds, setExpectedFunds] = useState<number>();
+  const [debouncedExpectedFunds] = useDebouncer(expectedFunds, 1000);
```
ap-justin, line 15:
```suggestion
  const [debouncedExpectedFunds, isDebouncing] = useDebouncer(expectedFunds, 1000);
```
`isLoading | isDebouncing` to show loading also when debouncing

ap-justin (changes requested):

src/services/aws/bankDetails/constants.ts
```diff
@@ -0,0 +1,40 @@
+import { CreateRecipientRequest, WiseRequest } from "types/aws";
+
+export const REQUESTS = {
```
ap-justin, line 3: it's better to move these requests to redux/query, specially for `GET` requests to take advantage of caching 

with current setup, 
everytime `/banking` is rendered (i.e.  user goes to `/banking`, then `/edit-profile`, then `/banking`), `GET /v1/currencies` is fired

ap-justin (changes requested):

src/pages/Admin/Charity/Banking/CurrencySelector.tsx
```diff
@@ -0,0 +1,97 @@
...
+      ? props.currencies
+      : props.currencies.filter((currency) => {
+          // check whether query matches either the currency name or any of its keywords
+          const formatQuery = query.toLowerCase().replace(/\s+/g, ""); // ignore spaces and casing
+          const matchesName = currency.name
+            .toLowerCase()
+            .replace(/\s+/g, "") // ignore spaces and casing
+            .includes(formatQuery);
+          const matchesKeyword = new RegExp(
+            currency.searchKeywords.join("|"),
```
ap-justin, line 33: seems like `searhKeyWords` is not a good filter attribute 
examples
search for `JPY`, or `PHP`, or `AUD`, and results are still broad

I tried `matchCode || matchName` and it yields more specific filter result

ap-justin (dismissed):

src/pages/Admin/Charity/Banking/CurrencySelector.tsx
misicnenad: > search for `JPY`, or `PHP`, or `AUD`, and results are still broad

I tested these values out with the current setup (available currencies Japanese yen and Australian dollar):
- searching by `JPY` returns only "Japanese yen" 
- searching by `PHP` returns nothing
- searching by `AUD` returns only "Australian dollar" 

What problem did you experience?

#### Explanation of the purpose of `searchKeywords`

Field `searchKeywords` is created by assigning [[wise currency].countryKeywords](https://docs.wise.com/api-docs/api-reference/currencies#object), the purpose of which is to aid the user in finding the desired currency.
You can check this by using Wise's URL to get all currencies:
https://api.sandbox.transferwise.tech/v1/currencies

The result looks like:
```js
[
    // ...
    {
        "code": "IDR",
        "symbol": "Rp",
        "name": "Indonesian rupiah",
        "countryKeywords": ["idn", "IDR", "ID", "Indonesia"],
        "supportsDecimals": false
    },
    {
        "code": "ILS",
        "symbol": "₪",
        "name": "Israeli shekel",
        "countryKeywords": [
            "ILS",
            "IL",
            "PS",
            "pse",
            "isr",
            "Israel",
            "Palestine"
        ],
        "supportsDecimals": true
    },
    // ...
]
```
Will add an explanatory comment to the code.

src/pages/Admin/Charity/Banking/CurrencySelector.tsx
misicnenad: Will be addressed in PR #2446

src/pages/Admin/Charity/Banking/useBanking.ts
misicnenad: Will be addressed in PR #2446

src/services/aws/bankDetails/constants.ts
misicnenad: Reason was that we're targeting one and the same AWS endpoint, but now that you mention caching, I think it might even be better to write all these calls as separate endpoints in `aws.injectEndpoints` :thinking: This would also solve remove the need for [useTypedWiseMutation](https://github.com/AngelProtocolFinance/angelprotocol-web-app/pull/2439/files#diff-ffd7f4b3dc2dac7b058b975c84453fb7bb7badead212d7afa60b65884f6f4c88R18), as each endpoint would have a defined expected return type.
Will update as per your suggestion :+1:

src/pages/Admin/Charity/Banking/Banking.tsx
ap-justin: > It is logical to expect the user to input this value once or twice, so the UX should be ok.
with this, it's even better to make transferring funds multi-step (with buttons `[continue] | [back]`)
1. user sees their balance in USD (e.g. 100USD)
2. decided to withdraw 50USD --> PHP, click `[continue]` and go to next tab/page
`( create quote, recipient requirements )` only fired once

3. user chooses transfer method 
4. user provides details for the chosen transfer method
5. show user summary (fees, exchange rates, expected amount to receive) before they commit to the transfer

misicnenad: @ap-justin 
This Banking page was created with the assumption that it would be used as a standalone page, not as part of other flows. So my question about not needing a button here was asked with this in mind.

@SovereignAndrey is Banking page meant to be a part of the withdrawal flow? Didn't see that in the requirements

src/pages/Admin/Charity/Banking/CurrencySelector.tsx
misicnenad: @ap-justin did my explanation alleviate your concerns?

src/services/aws/bankDetails/constants.ts
misicnenad: See https://github.com/AngelProtocolFinance/angelprotocol-web-app/pull/2439/commits/7b0a6f7f8ec805c18446dd5e5ff6353abf27874d

src/pages/Admin/Charity/Banking/CurrencySelector.tsx
ap-justin: > Field searchKeywords is created by assigning [[wise currency].countryKeywords](https://docs.wise.com/api-docs/api-reference/currencies#object), the purpose of which is to aid the user in finding the desired currency.
You can check this by using Wise's URL to get all currencies:
https://api.sandbox.transferwise.tech/v1/currencies

currently searching experience based on this is not good 
image


image


if you just based it on `code` and `name`, these are the results
image

image

image


it just doesn't make sense for people to use these keywords to search for their currency 
```
 {
    "code": "EUR",
    "symbol": "€",
    "name": "Euro",
    "countryKeywords": [
      "Cambodia",
      "cck",
      "fin",
      "Paraguay",
      "lux",
      "uga",
      "iot",
      "Solomon Islands",
      "Montserrat",
      "deu",
      "lva",
      "egy",
      "Guadeloupe",
      "hnd",
      "nzl",
      "AD",
      "fji",
      "Falkland Islands",
      "AE",
      "AG",
      "Seychelles",
```

src/pages/Admin/Charity/Banking/Banking.tsx
ap-justin: @misicnenad, @SovereignAndrey 

> This Banking page was created with the assumption that it would be used as a standalone page, not as part of other flows.

I'm not sure what this PR ultimately is trying to accomplish, 
is it just trying to save bank details so user would be able to use this information on any operation later on? (hence, the general banking term )

looking at the wise resources used in this PR  -these resources are used for  [transferring money](https://docs.wise.com/api-docs/guides/send-money) (which in our context is like charity withdrawing funds from AP bank )
1. AP profile (source of fund) creates quote 
2. get recipient details based on quote
3. create transfer based on quote and provided recipient details
4. fund transter
5. done 

I am not sure whether recipient created on `2` can be reused on another `quote`(new transfer/different amount). Or if you create a recipient without first creating a quote - would this recipient be valid to be used once you finally create a quote for a later transter ? Also user might opt to receive funds via bank but later on want to receive funds via e-wallet - this is not possible if we only attach a single `recipient_id` on charity's profile

src/pages/Admin/Charity/Banking/RecipientDetails/useRecipientDetails.ts
```diff
@@ -0,0 +1,148 @@
...
+      }
+
+      const requirements = requirementsDataArray.at(selectedIndex);
+      if (!requirements) {
+        throw new UnexpectedStateError("Requirements not loaded.");
+      }
+
+      setSubmitting(true);
+      // refresh requirements if necessary, see https://docs.wise.com/api-docs/api-reference/recipient#account-requirements
+      if (requirements.refreshOnSubmit && !requirements.refreshed) {
```
ap-justin, line 109: @misicnenad, this would mean user would user would need to submit (`handleSubmit`) whenever to see new requirement in case one of the fields `refreshRequirementsOnChange` - what if new requirements requires refresh again ?  user would have to submit form repeatedly

I was expecting the form to refresh when I choose recipient type
image

ap-justin (changes requested):

src/pages/Admin/Charity/Banking/RecipientDetails/useRecipientDetails.ts
misicnenad: You are right, the user *would* need to submit the form to be able to load the additional requirements (if necessary). This is because of Wise API requirements, see the below quote from [2nd paragraph in docs](https://docs.wise.com/api-docs/api-reference/recipient#account-requirements):
> Use the GET endpoint to learn what datapoints are required to send a payment to your beneficiary. As you build that form, use the POST endpoint to learn if any additional datapoints are required as a result of passing a field that has `"refreshRequirementsOnChange": true' in the GET Response. You should be posting the same recipient account payload that will be posted to `v1/accounts`.

> I was expecting the form to refresh when I choose recipient type

I would be expecting the same, but as per the explanation above, this is not possible :confused: 

To add more clarity to this logic, I added a "Check requirements" button and a short explanatory text to the form when requirement refresh is necessary. This should provide more context to our users.
I also updated the relevant logic:
- simplified check for whether refresh is needed
- separated "refresh requirements" logic from "submit form" logic

See commit https://github.com/AngelProtocolFinance/angelprotocol-web-app/pull/2439/commits/52237ae2cf5cabcabfdae6fd69d715635409a230

src/pages/Admin/Charity/Banking/Banking.tsx
misicnenad: > I'm not sure what this PR ultimately is trying to accomplish,

The purpose of this banking page as far as I understand for now (@SovereignAndrey please jump in if I miss anything) is to allow endowments to add their bank details that we can use to:
1. Create a recipient in our Wise account (see https://docs.wise.com/api-docs/api-reference/recipient#create)
2. Send funds to these recipients when relevant endowments request a withdrawal

Regarding 1), you can check the requirements by looking at the big parent Linear issue and its subtasks:
[AP-828: Collect / Setup an Endowments Banking details](https://linear.app/angel-protocol/issue/AP-828/collect-setup-an-endowments-banking-details)

Regarding 2), these are Andrey's words:
image

**For more context on the types used here and the logic itself, see comments under the related ticket linked below - it explains the types used, the expected format, the reasons for using AWS as proxy to Wise, the reason for dynamic fields etc.:**
[AP-829: Define minimum Wise API fields and shape for creating a recipient `Recipient`](https://linear.app/angel-protocol/issue/AP-829/define-minimum-wise-api-fields-and-shape-for-creating-a-recipient)

**P.S.** I updated this PR's ticket to include the decision to add the "What is amount of donations you expect to keep on our platform" input, see [this comment](https://linear.app/angel-protocol/issue/AP-834/admin-banking-page-posts-bank-details-to-aws-endpoint-and-can-show#comment-ce8b6194).

src/pages/Admin/Charity/Banking/CurrencySelector.tsx
misicnenad: Found the main culprit for such a long list of (unintuitive) matches - it was a bug in the filter function. 

Nevertheless, I see what you mean. Even with a shorter list of matches, some of the keywords used by Wise just makes for nonsensical result list. This is because Wise's logic goes like "when a user searches for a currency code in a specific country, we will match not only the currency itself, but all the currencies that are available in said country". This is why whatever you write in the search box, USD and EURO will almost certainly pop up - because they are possible bank account currencies in most of the countries of the world.

Updated currency filter to always match only by code or name (ignore `WiseCurrency.countryKeywords`), see commit https://github.com/AngelProtocolFinance/angelprotocol-web-app/pull/2439/commits/c4b02326a82a0ba562b1af80a756ce0e7d6e4e57

src/pages/Admin/Charity/Banking/Banking.tsx
ap-justin: > The purpose of this banking page as far as I understand for now (@SovereignAndrey please jump in if I miss anything) is to allow endowments to add their bank details that we can use to:

Create a recipient in our Wise account (see https://docs.wise.com/api-docs/api-reference/recipient#create)

> Send funds to these recipients when relevant endowments request a withdrawal

> For more context on the types used here and the logic itself, see comments under the related ticket linked below - it explains the types used, the expected format, the reasons for using AWS as proxy to Wise, the reason for dynamic fields etc.:
AP-829: Define minimum Wise API fields and shape for creating a recipient Recipient

@misicnenad 
I am well aware of these, we need recipient to send money. I am not asking what it is needed to created one but rather why we create `recipient` as a standalone entity
by getting requirements from (without quote context)
`GET /v1/account-requirements?source=EUR&target=USD&sourceAmount=1000`

 which we will save for later use, rather than `recipient` being just ephemeral (the only time they provide these details when they withdraw, just like we repeatedly provide our credit card details when purchasing online) and part of a send money process

I believe the reason is stated here, @SovereignAndrey 
image
which says user's wouldn't need to initiate send money process themselves

but going for this is approach has some [caveat](https://docs.wise.com/api-docs/guides/send-money/recipient#create-recipient)
> These factors additionally change on a regular basis, meaning creating and maintaining recipient account can be difficult.

perhaps its safer to just have the user input their bank details? (which is commonly account number, account name) whenever they withdraw.

src/services/aws/bankDetails/bankDetails.ts
misicnenad: This is outdated logic

src/pages/Admin/Charity/Banking/Banking.tsx
misicnenad: @ap-justin this concern should be directed to @SovereignAndrey

misicnenad:
Added bank verification status, see description to check out screenshots of the component.

NOTE: as was said before, this PR only focuses on the *functionality*. UI will be handled in a separate PR #2446

src/components/Selector/Selector.tsx
ap-justin: @misicnenad, since the dynamic use case was only needed in the requirement field, I would have preferrred to just type the path to `any` (as it is not known in advance)
```ts
<Selector<FV, any, string name={dynamicName} />
```
rather than changing all use cases that already benefit from this interface. It's true the `<Field/>` has less strict interface but it doesn't mean it's okay to level into this. 

now mismatches, `FV.usState -> string`, or `<Selector name="city" options={usState}/>` won't be detected anymore which has proved helpful to me when I did some refactoring in registration and widget

I really hope you could bring this strictness level back as the pros of removing it (`declaration a bit shorter to write`) is aesthetic

ap-justin (dismissed):

src/components/Selector/Selector.tsx
misicnenad: Sure thing, `<Selector<FV, any, string name={dynamicName} />` works fine for bank page needs.

Reverted at https://github.com/AngelProtocolFinance/angelprotocol-web-app/pull/2439/commits/2ea112e2f9593f208bc1f0e2b432d9335560282b

misicnenad:
@ap-justin addressed Selector-related change, see https://github.com/AngelProtocolFinance/angelprotocol-web-app/pull/2439#discussion_r1378640304

misicnenad:
@ap-justin @SovereignAndrey UI part of the task (PR https://github.com/AngelProtocolFinance/angelprotocol-web-app/pull/2446) is ready for review as well, will move it out of "draft" once this PR is merged

src/pages/Admin/Charity/Banking/Banking.tsx
ap-justin: resolving this now as got clarifiation from sprint meeting

PR #2456, opened 2023-11-01
Create reusable `optionType()` schema

misicnenad:
[OptionType<V>](https://github.com/AngelProtocolFinance/angelprotocol-web-app/blob/RC-v1.9/src/components/Selector/types.ts#L6) is a commonly used type in forms, for which we usually create ad-hoc Yup schemas.
This PR intends to create a reusable schema type to be (re)used in these forms.

Schema looks like:
```ts
export const optionType = ({ required } = { required: false }) =>
  Yup.object<any, SchemaShape<OptionType<string>>>({
    label: required ? requiredString : Yup.string(),
    value: required ? requiredString : Yup.string(),
  });
```
And is used like:
- `field: optionType({ required: true })` - a required field
- `field: optionType({ required: false })` - an optional field
- `field: optionType()` - same as previous, an optional field

It is assumed that `OptionType<>.value` is to be of type `string` even when the actual value is of type `number` for the following reasons:
- since we only need to check whether the value is present (i.e. required) we can allow for `number` to automatically be converted to `string` for this purpose (as any number including 0 (zero) is converted to a non-empty string)
- if the actual type should be `number`, the validation would still function and the final value would still be of type `number`
- all the above being the case, it makes the implementation cleaner and easier to write

The reason for wrapping the `required` option in an object is so that when it needs to be used, it makes the intention more explicit:
```ts
  // intention to make the field "required" is clear
  endowDesignation: optionType({ required: true })
  // ...

  // the function accepts some boolean value, purpose of the parameter is not immediately apparent
  endowDesignation: optionType(true)
  // ...
```
Also allows for adding more parameters in the future (doubtful if it'll be necessary, but the option is there).

-

## UI changes for review

### DIFF
--- a/src/schemas/shape.ts
+++ b/src/schemas/shape.ts
@@ -1,7 +1,7 @@
 import * as Yup from "yup";
 import { SchemaShape } from "./types";
 import type { TokenWithAmount as TWA } from "types/tx";
-import { OptionType } from "components/Selector";
+import { OptionType } from "types/utils";
 import { tokenConstraint } from "./number";
 import { requiredString } from "./string";
 
--- a/src/schemas/shape.ts
+++ b/src/schemas/shape.ts
@@ -1,7 +1,9 @@
 import * as Yup from "yup";
 import { SchemaShape } from "./types";
 import type { TokenWithAmount as TWA } from "types/tx";
+import { OptionType } from "components/Selector";
 import { tokenConstraint } from "./number";
+import { requiredString } from "./string";
 
 type Key = keyof TWA;
 type Min = TWA["min_donation_amnt"];
@@ -12,7 +14,7 @@ const balKey: Key = "balance";
 export const tokenShape = (withMin = true): SchemaShape<TWA> => ({
   amount: Yup.lazy((amount: string) =>
     amount === ""
-      ? Yup.string().required("required")
+      ? requiredString
       : tokenConstraint.when([minKey, balKey], (values, schema) => {
           const [minAmount, balance] = values as [Min, Bal];
           return withMin && !!minAmount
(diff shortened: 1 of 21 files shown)


src/schemas/shape.ts
```diff
@@ -1,7 +1,9 @@
 import * as Yup from "yup";
 import { SchemaShape } from "./types";
 import type { TokenWithAmount as TWA } from "types/tx";
+import { OptionType } from "components/Selector";
```
ap-justin, line 4: @misicnenad, everything looks good , except `schemas` now imports from `components`. Though I have not documentented this,  I've been striving to not to make cyclical imports between folders (which having one proved to be so catastrophic when trying to make `package/library` out of a folder in monorepo setup ) and keep it structured e.g

third-party
types
constants
schemas
helpers 
components
pages

though not yet perfect throughout the codebase atm
e.g. u can see `schemas/string` importing from `helpers` 

one solution could be to put `OptionType` into say `types/utils` as I have found that `slices` also imports from components currently 🤦‍♂️

misicnenad: Good catch, I'd like to avoid importing from `components/*` folder inside `src/schemas/*`.

> one solution could be to put `OptionType` into say `types/utils`

Sounds good

misicnenad: See:
https://github.com/AngelProtocolFinance/angelprotocol-web-app/pull/2456/commits/1abe760540aafab25df6c680942cf3ff21ecea09
https://github.com/AngelProtocolFinance/angelprotocol-web-app/pull/2456/commits/539c39cdcf00621751e62519146c3cfeafe59463

SovereignAndrey (approved):

PR #2469, opened 2023-11-13
Component types

ap-justin:
Ticket(s):
- N/A 
- bug: CountrySelector sets `isDirty:true` on mount - related to #2467 

* see comments for reasons of changes

-

## UI changes for review


src/components/CountrySelector/CountrySelector.tsx
```diff
@@ -49,21 +44,6 @@ export default function CountrySelector<
 
   const [query, setQuery] = useState(country.name);
 
-  /**
```
ap-justin, line 52: removing this `useEffect`  as it makes the form it belongs to `dirty` (related to #2467 ), Forms using `<CountrySelector/>` should now take care of setting flags for persistent country records
see changes: 
EditProfile
Documentation

src/types/components.ts
```diff
@@ -0,0 +1,18 @@
+import { FileObject } from "./aws";
```
ap-justin, line 1: as widely used component related type tend to be single for reach component, just group them in this file.

src/pages/Admin/Charity/Banking/RecipientDetails/types.ts
```diff
@@ -1,7 +1,7 @@
 import { CreateRecipientRequest } from "types/aws";
-import { Country } from "types/countries";
-import { OptionType } from "types/utils";
-import { Asset } from "components/registration";
+import { OptionType } from "types/components";
+import { Country } from "types/components";
+import { Asset } from "components/FileDropzone";
```
ap-justin, line 4: moved Filedropzone from `components/registration` to `components/` as it is used also in admin now

src/pages/Admin/Charity/Banking/RecipientDetails/getDefaultValues.ts
```diff
@@ -1,7 +1,7 @@
 import { FormValues } from "./types";
 import { AccountRequirements } from "types/aws";
-import { Country } from "types/countries";
-import { genFileAsset } from "pages/Registration/Steps/getRegistrationState";
```
ap-justin, line 4: moved this helper to `components/Filedropzone` as it now both used by admin and registration

src/components/CountrySelector/helpers.ts
```diff
@@ -0,0 +1,11 @@
+import { Country } from "types/components";
+import countries from "assets/countries/all.json";
+
+export const country = (name: string): Country => {
```
misicnenad, line 4:
```suggestion
export const country = (name = ""): Country => {
```
nit: makes it unnecessary to do `country(props.hq_country ?? "")`

src/types/components.ts
```diff
@@ -0,0 +1,18 @@
...
+
+//selector
+export type ValKey = string | number;
+export type OptionType<V> = { label: string; value: V };
+
+//dropzone
+export type Asset = {
```
misicnenad, line 8: The more I'm looking at it, the more I get the feeling of how generic of a name `Asset` is. Maybe a better one would be `FileDropzoneAsset`?
It'd be a small change, but would go a long way of making it obvious what its purpose is.

misicnenad (approved):
:+1:

src/types/components.ts
ap-justin: I agree with this ser, I'll include this change in #2467

PR #2483, opened 2023-11-15
Remove jwt secrets in favor of CORS

ap-justin:
Ticket(s):
marketplace cards takes forever to load
- https://discord.com/channels/834976399228010527/857111094019883010/1173846552251465759

* remove auto-setting of authorization for all aws endpoints
* only set authorization for select endpoints
* remove secret keys which is now not used  in favor fo CORS

-

## UI changes for review


.env.example
```diff
@@ -4,23 +4,8 @@ GENERATE_SOURCEMAP=<TRUE | FALSE>
 # port to use for serving the application on (defaults to 3000)
 PORT=<port number>
 
+REACT_APP_NETWORK=<TESTNET | MAINNET>
 
 
-REACT_APP_NETWORK=<LOCAL | TESTNET | MAINNET>
```
ap-justin, line 9: we don't need this anymore as `rpcs and lcds` are now defined in `constants/chains`

src/services/aws/registration/registration.ts
```diff
@@ -23,6 +24,7 @@ const registration_api = aws.injectEndpoints({
         url: `${v(4)}/registration`,
         method: "POST",
         body: { Email: email },
+        headers: { authorization: TEMP_JWT },
```
ap-justin, line 27: getting jwt token from current session is async but headers per endpoints can only be defined sync so set `TEMP_JWT` which will be replaced when hits `service.prepareHeaders`

.env.example
SovereignAndrey: Will make sure it's removed from AWS side.

amplify.yml
```diff
@@ -11,9 +11,6 @@ frontend:
         - yarn install
     build:
       commands:
-        - REACT_APP_APES_AUTH_SECRET_KEY=${REACT_APP_APES_AUTH_SECRET_KEY}
-        - REACT_APP_ANGEL_AUTH_SECRET_KEY=${REACT_APP_ANGEL_AUTH_SECRET_KEY}
-        - REACT_APP_KADO_API_KEY=${REACT_APP_KADO_API_KEY}
         - REACT_APP_NETWORK=${REACT_APP_NETWORK}
```
SovereignAndrey, line 14: Can remove this from build I think since we're not using it anymore.
```suggestion
```

amplify.yml
```diff
@@ -11,9 +11,6 @@ frontend:
...
         - REACT_APP_JUNO_LCD_NODE=${REACT_APP_JUNO_LCD_NODE}
```
misicnenad, line 15:
```suggestion
```
Should we remove this line too? 

There are other uses of the same env. var in [.github/workflows/javascript.yml](https://github.com/AngelProtocolFinance/angelprotocol-web-app/blob/auth-token/.github/workflows/javascript.yml)

misicnenad (changes requested):
Other than this, all looks good

amplify.yml
ap-justin: checked that this env profile is still in use for these endpoints
POST `${v(1)}/fiat/stripe-proxy/apes/${network}`,
GET `/${v(5)}/endowments/${network}`
 GET `/${v(2)}/profile/${network}/endowment/${endowId}`
 GET `/${v(1)}/profile/${network}/program/${endowId}/${programId}`
DELETE |  PUT `/${v(3)}/profile/${network}/endowment`

ser @jp-mariano, let me know here the updated endpoints without network

amplify.yml
ap-justin: fixed

amplify.yml
SovereignAndrey: @jp-mariano We can also leave this one alone for now, since it's not sensitive data, it's not a high priority to spend a lot of time refactoring out and risk introducing bugs so close to the launch date. If it's just a single recent lambda to update or all those mentioned above are (or will be) depreciated post-v1.9 launch, then full steam ahead.

.env.example
SovereignAndrey: @ap-justin This is still needed locally for development? I get mainnet endowment data when I remove this from my local `.env`.

amplify.yml
jp-mariano: POST `${v(1)}/fiat/stripe-proxy/apes/${network}`
- For now, we can keep this as is. The `stripe-proxy` Lambda would probably still be refactored and from there we can update the endpoint it uses to one without the `network` parameter.

GET `/${v(5)}/endowments/${network}`
- This is the marketplace endpoint and `network` here is still needed in order to differentiate prod from staging endowments. We can keep this for now.

GET `/${v(2)}/profile/${network}/endowment/${endowId}`
- This is the `get-single-endowment` Lambda's old endpoint which will be deprecated. The new endpoint here should be `GET /v1/profile/endowment/{id}` or `GET /staging/profile/endowment/{id}`

GET `/${v(1)}/profile/${network}/program/${endowId}/${programId}`
- This is currently a working endpoint and `network` here is needed in order to differentiate prod from staging endows. We can also keep this for now.

DELETE | PUT `/${v(3)}/profile/${network}/endowment`
- This endpoint will be deprecated and the new one should be `DELETE | PUT /v1/profile/endowment` or `DELETE | PUT`

ap-justin: got this @jp-mariano

src/services/aws/aws.ts
```diff
@@ -127,7 +127,7 @@ export const aws = createApi({
         error ? [] : ["endowments", "profile", "walletProfile"],
       query: (payload) => {
         return {
-          url: `/${v(3)}/profile/${network}/endowment`,
+          url: `/${v(3)}/profile/endowment`,
```
misicnenad, line 130:
```suggestion
          url: `/${v(1)}/profile/endowment`,
```
As per ser Paolo's [comment](https://github.com/AngelProtocolFinance/angelprotocol-web-app/pull/2483#discussion_r1395103450):
> This endpoint will be deprecated and the new one should be `DELETE | PUT /v1/profile/endowment` or `DELETE | PUT`

misicnenad (changes requested):

SovereignAndrey (approved):

src/services/aws/aws.ts
SovereignAndrey: Oh! Good catch!

ap-justin: fixed https://github.com/AngelProtocolFinance/angelprotocol-web-app/pull/2483#:~:text=edit%20profile%20version-,901e642,-ap%2Djustin%20dismissed

SovereignAndrey (approved):

PR #2488, opened 2023-11-16
AP-867: Registration: Add Banking Details step

misicnenad:

- start registration
- get to step 5 (bank details)

NOTE: You can use https://www.iban.com/structure to get sample IBANs

## UI changes for review

image
image
image

image
image


SovereignAndrey (changes requested):
@misicnenad Just a few items to address and some other questions that came up while going through everything.

1. We should probably restrict payouts to businesses only since these are supposed to be registered Charities and Non-Profits that _should_ have a separate bank account set up. 
image

2. [NOT CRITICAL FOR THIS PR] Bank statements should probably only be allowed in formats common for documents of that type, namely multiple page files: PDF, TIF/TIFF, or ???. We want to avoid people screenshotting a single page of what is, in reality, a multiple page document and asking why they can't upload multiple JPG/PNG/etc. :laughing: 
image

3. Form says that the "Bank code (BIC/SWIFT)" field is optional, so I skipped it, but was then told it was required upon trying to submit the form. 
Screenshot from 2023-11-17 09-22-48
PS. This site is good one to find SWIFT codes: https://www.theswiftcodes.com/

4. When I click "Check Requirements", I'm sent to the top of the page and then have to scroll back down to hit continue again. 
    - If all is good, then it should either: just take the user to the next step (ie. the Summary page) OR stay at the bottom, show a message like "<check mark> All requirements are met! Please click continue." 
    - If there's new info that's needed, show a small message "We need some more information from you, please see the red fields", and have the form note those fields with red outlines. 

5. I finished filling out the form completely and the check requirements passed BUT, if then I change the estimated amount to something big that would require more information:
a) the form gets new fields. This is OK. 
b) I lost all the information I just filled in. This is BAD. I'd expect previously entered data to be mapped over to the new form fields, saving the user from re-entering everything. Might be an edge case where the user changes that estimate number after the fact, but it's a big pain point if they do actually that. :thinking:

SovereignAndrey:
@misicnenad Might be AWS related (cc: @jp-mariano on this) but I get a 400 error when trying to submit the completed banking step (after passing check requirements). 
image

src/pages/Registration/Steps/Banking/useSubmit.ts
```diff
@@ -1,43 +1,71 @@
...
-    formState: { isDirty, isSubmitting },
-  } = form;
-
+  const [createRecipientAccount] = useCreateRecipientAccountMutation();
   const [updateReg] = useUpdateRegMutation();
   const { handleError } = useErrorContext();
   const navigate = useNavigate();
+  const [isSubmitting, setSubmitting] = useState(false);
+
+  const submit = useCallback(
```
ap-justin, line 21: when i try to submit i get error
```
Request URL:
https://kpnxz5rzo2.execute-api.us-east-1.amazonaws.com/staging/wise
Request Method:
POST
Status Code:
422 Unprocessable Content
Remote Address:
44.205.162.160:443
Referrer Policy:
strict-origin-when-cross-origin
```

payload 
```
{
    "method": "POST",
    "url": "/v1/accounts",
    "PK": "a037a29b-f328-46de-94b0-eceb38e81499",
    "payload": "{
    "method": "POST",
    "url": "/v1/accounts",
    "PK": "a037a29b-f328-46de-94b0-eceb38e81499",
    "payload": "{\"accountHolderName\":\"testorg\",\"currency\":\"PHP\",\"type\":\"philippinesmobile\",\"details\":{\"legalType\":\"PRIVATE\",\"email\":\"mail@mail.com\",\"accountHolderName\":\"test last\",\"bankCode\":\"GCASH\",\"accountNumber\":\"9983651234\",\"address.country\":\"PH\",\"address.city\":\"city\",\"address.firstLine\":\"address1\",\"address.postCode\":\"1111\"}}"
}
```

ap-justin, line 21: nit: `useCallback` is [generally not necessary](https://kentcdodds.com/blog/usememo-and-usecallback#why-is-usecallback-worse), as there's no need for `submit` to be referentially equal all the time (unless used in `useMemo` or `useEffect` [deps](https://kentcdodds.com/blog/usememo-and-usecallback#dependencies-lists))

src/types/aws/ap/registration.ts
```diff
@@ -121,7 +122,7 @@ export type TDocumentation = {
 };
 
 export type BankingDetails = {
-  BankStatementPDF: FileObject;
```
ap-justin, line 124: `BankStatementPDF` is still used in lambda `hubspot-company-updatooor @jp-mariano

misicnenad:
@SovereignAndrey re https://github.com/AngelProtocolFinance/angelprotocol-web-app/pull/2488#pullrequestreview-1735838383:

> 1. We should probably restrict payouts to businesses only...

I'd agree, but this field is one of the dynamically loaded ones we get from Wise. This means we can't know in advance:
- whether we'll get this field
- what its key would be (used as ID/key for the form)
- what the available values will be

Note: I noticed that occasionally we'd get this field, but the only available value would be just one of the two from the image you attached.

We could update the logic so that if we get a field with the key being the most common one for "Recipient Type", we immediately set it to "Business". But what if "Business" is not available? Do we let the user choose then? What value will they choose?

This might be best handled during bank details review. We'd then be able to reject the submission if we deem so necessary based on the value they chose here (e.g. "Personal") and let them know what we expect of them to do.

Note 2: More often than not the user will *have* to choose "Business" as they'd be inputting their charity's bank details, which by definition should be using a "business account".

> 2. Bank statements should probably only be allowed in formats common for documents of that type

Will update.

> 3. Form says that the "Bank code (BIC/SWIFT)" field is optional, so I skipped it, but was then told it was required upon trying to submit the form.

:open_mouth: will update.

> 4. When I click "Check Requirements", I'm...

This will probably be more complicated to implement than it sounds. Will work on this today.

> 5. I finished filling out the form completely...

Same as above, only more so.

The thing is that I tried to make it so that switching between requirement types (ACH/WIRE/SWIFT etc.) doesn't wipe out the previously inputted data in not-currently-selected requirement types. I gave some effort to doing the same when changing the estimated amount, but after realizing it would result in an even more complex implementation and suck even more of my time, all for a relative edge case, I figured it might be better to push something to RC branch *now* and implement all the nice-to-haves *later*.
I have a short list of these nice-to-haves for which I intend to create tickets, so that we (or just I) can focus on those more complicated bits completely instead of trying to juggle them with other requirements.

src/pages/Registration/Steps/Banking/useSubmit.ts
misicnenad: You're right, removed.

See https://github.com/AngelProtocolFinance/angelprotocol-web-app/pull/2488/commits/6394ab9d733a292ae9e90dae55d419ed7c3e6e38

misicnenad: Can you post the exact repro steps please?

ap-justin: REPRO in registration banking page
* select `PHP` as currency
* any amount
* choose `e-wallet` 
* recipient account number `9983651234` postcode `1111`, country: Philippines, city: city
* check requirements button
* continue

misicnenad: Working on this, will coordinate with Paolo if necessary

misicnenad:
Converted to draft as this is still in progress https://github.com/AngelProtocolFinance/angelprotocol-web-app/pull/2488#discussion_r1397016238
Experiencing some authorization issues atm, coordinating with Paolo

src/pages/Registration/Steps/Banking/useSubmit.ts
misicnenad: Solved all the (apparent) errors on webapp side, now getting `400 Bad Request`.
This is most likely due to a missing Profile ID assignment on AWS side, waiting on Paolo to implement this.

SovereignAndrey:
> This might be best handled during bank details review. We'd then be able to reject the submission if we deem so necessary based on the value they chose here (e.g. "Personal") and let them know what we expect of them to do.

I think this is fine for now. :+1: The manual review process can determine if it's acceptable based on the size of the charity, how new it is, and the country HQ.

SovereignAndrey:
> **EDIT 3: judging by [create recipient docs](https://docs.wise.com/api-docs/api-reference/recipient#create) (section _Request > details.legalType_), it might be that `legalType` is always returned for every currency 🤔 Do you understand this the same way @SovereignAndrey?**

I read that as all Recipient's have both a `legalType` [personal || business] and a `currency` field associated with them. Currency object itself doesn't have the concept of legal Type, as it is just a string code, a name, and symbol prefix char.

jp-mariano:
> @misicnenad Might be AWS related (cc: @jp-mariano on this) but I get a 400 error when trying to submit the completed banking step (after passing check requirements). image
> 
> **UPDATE:** Might just be CORS related from my local machine (though I'm running on port 4200). Seeing a 403 on file-upload endpoint.

I have removed all non-Cognito JWT Authorizers in AP AWS and in some APES AWS endpoints. Today, I'll start working on implementing the source domain checks that will replace those non-Cognito JWT Authorizers.

src/types/aws/ap/registration.ts
jp-mariano: This is also the case for `registration-update` Lambda's schema. I will update both files. 👍🏾

misicnenad:
Fixed all issues with front-end code, the recipient should be created successfully after Paolo implements fixes on AWS side
@SovereignAndrey @ap-justin

PR #2511, opened 2023-11-23
AP-904: Page authorization

ap-justin:
Ticket(s):
-

* when user is ap-admin, will see applications link
image

* when user is not authorized (e.g. going to `/applications` but not ap-admin)
image

-

## UI changes for review


src/store/store.ts
```diff
@@ -1,25 +1,53 @@
...
     getDefaultMiddleware().concat([aws.middleware, apes.middleware]),
 });
+
+config.oauth.redirectSignIn =
+  window.location.origin + `${appRoutes.auth_redirector}/`;
+config.oauth.redirectSignOut = window.location.origin + "/";
+Amplify.configure(config);
+
+store.dispatch(loadSession());
+Hub.listen("auth", async ({ payload }) => {
```
ap-justin, line 34: upgraded to amplify v6 (stable now) as Hub has better types

src/slices/auth.ts
```diff
@@ -0,0 +1,106 @@
...
+        id: attributes.email || _user.username,
+        isSigningOut: false,
+      };
+    } catch (err) {
+      return null;
+    }
+  }
+);
+
+const auth = createSlice({
```
ap-justin, line 77: ditch `useAuthenticator` hook for user state management due to limitations
* in `v5` retrieval of claims from jwt is of node style instead of react
```
getSession(callback: ((error: Error, session: null) => void) | ((error: null, session: CognitoUserSession) => void), options?: GetSessionOptions | undefined): void
```
* in `v6`, need to call `fetchAuthSession` separately 


instead created `auth` slice which centrally manages user state (updated via auth events)

src/services/aws/aws.ts
```diff
@@ -32,9 +33,13 @@ const awsBaseQuery = retry(
   fetchBaseQuery({
     baseUrl: APIs.aws,
     mode: "cors",
-    async prepareHeaders(headers) {
+    async prepareHeaders(headers, { getState }) {
+      const {
+        auth: { user },
+      } = getState() as RootState;
```
ap-justin, line 39: since auth state is now on redux, can just reuse

src/slices/auth.ts
```diff
@@ -0,0 +1,106 @@
...
+        user ? Promise.resolve(user) : getCurrentUser(),
+      ]);
+
+      //remove standard claims
+      const payload = session.tokens.accessToken.payload;
+      const token = session.tokens.accessToken.toString();
+
+      const credentials = [
+        Array.isArray(payload["cognito:groups"]) &&
+        payload["cognito:groups"].includes("AngelProtocolAdmin")
```
ap-justin, line 60: FUTURE: jwt contains list of endowments user is authorized to access @jp-mariano

SovereignAndrey:
@ap-justin How does a user get set as `ap-admin`? Users table in AP AWS or elsewhere? Thanks! 
Will try to start looking over this after dinner. :+1:

ap-justin:
> @ap-justin How does a user get set as `ap-admin`? Users table in AP AWS or elsewhere? Thanks! Will try to start looking over this after dinner. 👍

in `Users` table set attribute `admin` -> `true`, and sign-in again, so that pre-token-generation lambda would attach the admin status

src/contexts/Auth.tsx
```diff
@@ -0,0 +1,65 @@
...
+    }
+
+    /**
+     * if wish to access user via props wrap component:
+     * withAuth(function Component({user, ...restProps}){
+     * })
+     */
+    return (
+      <Context.Provider value={user}>
+        <Component {...props} user={user} />
```
misicnenad, line 49: Isn't assigning `user={user}` prop to `Component` redundant?
Can't the component use `useAuthenticatedUser` to access user context data?

src/contexts/Auth.tsx
```diff
@@ -0,0 +1,65 @@
...
+      </Context.Provider>
+    );
+  };
+}
+
+const INIT = "__init__" as unknown as AuthenticatedUser;
+const Context = createContext<AuthenticatedUser>(INIT);
```
misicnenad, line 56: Just to make sure I understood this code, the reason for assigning `INIT` here it so that `AuthenticatedUser` is always defined?

src/pages/OAuthRedirector.tsx
```diff
@@ -1,17 +1,6 @@
...
-  useEffect(() => {
-    const from = localStorage.getItem(OAUTH_PATH_STORAGE_KEY);
-    navigate(from || "/");
-  }, [navigate]);
-  return (
-    <div className="grid place-items-center">
-      <LoaderRing thickness={12} classes="w-32" />
-    </div>
-  );
+  return <Navigate to={localStorage.getItem(OAUTH_PATH_STORAGE_KEY) ?? "/"} />;
```
misicnenad, line 5: Cleaner :+1:

src/services/aws/aws.ts
```diff
@@ -32,9 +33,13 @@ const awsBaseQuery = retry(
   fetchBaseQuery({
     baseUrl: APIs.aws,
     mode: "cors",
-    async prepareHeaders(headers) {
+    async prepareHeaders(headers, { getState }) {
```
misicnenad, line 36:
```suggestion
    prepareHeaders(headers, { getState }) {
```
no longer `async`

src/slices/auth.ts
```diff
@@ -0,0 +1,106 @@
...
+      ].filter((item): item is Credential => item !== null);
+
+      return {
+        token,
+        credentials,
+        id: attributes.email || _user.username,
+        isSigningOut: false,
+      };
+    } catch (err) {
+      return null;
```
misicnenad, line 72: Should handle this error or at leas rethrow it

misicnenad (changes requested):
Looks good, some comments

SovereignAndrey:
@ap-justin Works good, per blocking non-ap-admin users and redirecting non-logged in users to signin as well for the Applications area.

Found another admin, but not this ap-admin related bug. Will post in discord chat.

src/contexts/Auth.tsx
ap-justin: > so that AuthenticatedUser is always defined?
another reason is to have a way to distinguish init context value as `useAuthenticatedUser` would result to `INIT` if used outside `withAuth`

we could set this to just `undefined` but would also need to coerce to silence typescript error
```ts
const Context = createContext<AuthenticatedUser>() //ts-error
const Context = createContext<AuthenticatedUser>(undefined as unknown as AuthenticatedUser) 
```
and check in hook as 
```js
//useAuthenticatedUser

if val = useContext(context)
if(val === undefined) // error: used outside withAuth
```

we can't however type like this 
```ts
const Context = createContext<AuthenticatedUser | undefined>() 
```
as consumers would then receive
```
AuthenticatedUser | undefined 
```
which is a case that isn't a thing but still needs to be checked everytime the hook is used

src/contexts/Auth.tsx
ap-justin: yes, it's redundant which adds flexibility , simple pages need not import accessor hook 
e.g. Donations
```ts
export default withAuth(function Donations({ user }) {
  const donorAddress = user.email;
```

but deeply nested components could use `useAuthenticatedUser` hook

src/services/aws/aws.ts
ap-justin: fixed

src/App/Header/UserMenu/Menu.tsx
```diff
@@ -1,40 +1,56 @@
...
+      <div className="hidden p-4 border-t border-prim">
+        <Link
+          to={appRoutes.donations}
+          className="text-orange hover:text-orange-l2 text-sm uppercase"
+        >
+          My Donations
+        </Link>
+      </div>
+
+      {/** TODO: temp hidden until user endowments are in jwt claim  */}
```
ap-justin, line 32: @SovereignAndrey, temp set this to `hidden` until proper authorization is attached in the authenticated user

SovereignAndrey (approved):

src/contexts/Auth.tsx
misicnenad: @ap-justin 
> we could set this to just undefined but would also need to coerce to silence typescript error
> ```ts
> const Context = createContext<AuthenticatedUser>() //ts-error
> const Context = createContext<AuthenticatedUser>(undefined as unknown as AuthenticatedUser)
> ```

That's of course necessary and the same exact thing is done for `INIT`
```ts
const INIT = "__init__" as unknown as AuthenticatedUser;
const Context = createContext<AuthenticatedUser>(INIT);
```

I understand what you were trying to achieve now, doesn't really matter which default value we use

src/slices/auth.ts
misicnenad: @ap-justin errors are always a possibility

PR #2517, opened 2023-11-24
AP-886: Scrolling up/down over expected monthly donations input field should not be changing the input value

misicnenad:
The simplest solution would've been:
```tsx
<input type="number" {/* ...other props... */} onWheel={(e) => e.currentTarget.blur()} />
```
But this would've caused the input to lose focus each time mouse wheel was used.

Opted for just using turning the input into a `type="text"` and using a custom logic to simulate `type="number`" behavior to ensure valid number value is used.

-

## UI changes for review

### DIFF
--- a/src/components/BankDetails/ExpectedFunds.tsx
+++ b/src/components/BankDetails/ExpectedFunds.tsx
@@ -45,13 +45,20 @@ export default function ExpectedFunds({ disabled, onChange }: Props) {
       <div className="w-60 md:w-80">
         <input
           id="amount"
-          type="number"
+          type="text"
           placeholder="10.000"
-          onChange={(event) => onChange(Number(event.target.value))}
+          onChange={(event) => {
+            const value = Number(event.target.value);
+            if (!isNaN(value)) {
+              onChange(value);
+            }
+          }}
           className="field-input text-field"
           autoComplete="off"
           spellCheck={false}
           disabled={disabled}
+          inputMode="numeric"
+          pattern="[0-9]*"
         />
       </div>
     </div>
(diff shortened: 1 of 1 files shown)


ap-justin (dismissed):

src/components/BankDetails/ExpectedFunds.tsx
```diff
@@ -45,13 +46,22 @@ export default function ExpectedFunds({ disabled, onChange }: Props) {
...
+              return event.preventDefault();
+            }
+            setValue(event.target.value);
+            onChange(tvalue);
+          }}
           className="field-input text-field"
           autoComplete="off"
           spellCheck={false}
           disabled={disabled}
+          inputMode="numeric"
```
misicnenad, line 64: > The inputmode [global attribute](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes) is an [enumerated](https://developer.mozilla.org/en-US/docs/Glossary/Enumerated) attribute that hints at the type of data that might be entered by the user while editing the element or its contents. This allows a browser to display an appropriate virtual keyboard.

See https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/inputmode

src/components/BankDetails/ExpectedFunds.tsx
```diff
@@ -45,13 +46,22 @@ export default function ExpectedFunds({ disabled, onChange }: Props) {
...
         <input
           id="amount"
-          type="number"
+          type="text"
+          value={value}
           placeholder="10.000"
-          onChange={(event) => onChange(Number(event.target.value))}
+          onChange={(event) => {
+            const tvalue = Number(event.target.value);
+            if (isNaN(tvalue)) {
```
misicnenad, line 54: Simulates `type="number"` behavior

ap-justin (approved):

PR #2526, opened 2023-11-27
Check req updates

misicnenad:
Ticket(s):
-

-

## UI changes for review


src/components/Selector/Selector.tsx
```diff
@@ -46,7 +46,7 @@ export function Selector<
           aria-invalid={!!get(errors, valuePath)?.message}
           aria-disabled={isDisabled}
           as="button"
-          className={`${button} ${styles.selectorButton} peer-focus:border-gray-d1 peer-focus:dark:border-blue-l2`}
+          className={`${button} ${styles.selectorButton}`}
```
misicnenad, line 49: When a form containing a `Selector` was submitted and validation errors occur, the screen scrolls up to the first field with an error and the UI is as the following:
- with these classes
image
- without:
image

@ap-justin is it necessary to keep these classes present for `Selector`?

src/components/BankDetails/RecipientDetails/useRecipientDetails.ts
```diff
@@ -38,7 +39,6 @@ export default function useRecipientDetails(
   const [selectedIndex, setSelectedIndex] = useState(0);
   const [quote, setQuote] = useState<Quote>(); // store quote to use to refresh requirements
   const [isLoading, setLoading] = useState(true); // starts in loading state, loads only once
-  const [isSubmitting, setSubmitting] = useState(false);
```
misicnenad, line 41: redundant

src/components/Selector/Selector.tsx
ap-justin: I think it's necessary to distinguish which one of the invalid fields are currently in focus - more evident if user then tabs to fill out other field

misicnenad: That makes sense :thinking:
A box shadow indicator might be a good indicator of a focused field while being more appealing to the eye (see image below), what do you think @ap-justin @SovereignAndrey ?
image

SovereignAndrey (approved):

PR #2536, opened 2023-11-28
AP-795: Modal filter

ap-justin:
Ticket(s):
-

* change filter icon
image

* sidebar now on modal and style changes. Country goes first so has room for its dropdown options
image

* selecting a filter immediately applies it (no need for separate submit step - unless filter is dynamic like typing a `keyword` which needs validation prior to application)
image

* user dismissing the modal sees previously applied filters and can clear accordingly
image

* when only single filter is active, don't show `clear all` button
image

-

## UI changes for review


SovereignAndrey:
@ap-justin This looks beautiful! :heart_eyes: 
Two quick comments from the screenshots alone (will dig into code later on):
1. Maybe change modal header "Country" >> "Countries" or maybe "Country(s)" so users know it supports more than 1 at a time?
2. Need a `<hr>` below the Country select and before "NPO Type" header.

SovereignAndrey:
> the divider looks redundant when near the countries box

@ap-justin Hmm... it could go either way. I guess we can leave it off then and let rest of the team make that call when they review. :+1:

SovereignAndrey (approved):

PR #2544, opened 2023-11-29
Fix: edit profile selected file preview not showing

ap-justin:
Ticket(s):
- https://discord.com/channels/834976399228010527/1098542651218866208/1179079645786812436

x is not of format `type/subtype` but of abstracted `SUBTYPE` e.g. `JPEG | PNG`
```ts
//useImgEditor.ts
      //preview & crop valid formats only
      if (!!accept.find((x) => x === newFile.type)) {
```

* mime type lists usage now refers to bare `type/subtype` format (created type for autocompletion)

## Others
* `Filedropzone` now creates tooltip internally by accepting props: file size , and valid mimetypes

-

## UI changes for review

### DIFF
--- a/src/types/lists.ts
+++ b/src/types/lists.ts
@@ -23,6 +23,7 @@ export type UNSDG_NUMS =
 export type TransactionStatus = "open" | "approved" | "expired";
 export type EndowmentType = "charity" | "ast" | "daf";
 
+//https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/MIME_types
 type ImageSubType = "svg" | "jpeg" | "png" | "webp";
 type ApplicationSubType = "pdf";
 
--- a/src/types/lists.ts
+++ b/src/types/lists.ts
@@ -22,3 +22,11 @@ export type UNSDG_NUMS =
 
 export type TransactionStatus = "open" | "approved" | "expired";
 export type EndowmentType = "charity" | "ast" | "daf";
+
+type ImageSubType = "svg" | "jpeg" | "png" | "webp";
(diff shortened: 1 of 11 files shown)


src/types/lists.ts
```diff
@@ -22,3 +22,12 @@ export type UNSDG_NUMS =
 
 export type TransactionStatus = "open" | "approved" | "expired";
 export type EndowmentType = "charity" | "ast" | "daf";
+
+//https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/MIME_types
+type ImageSubType = "svg" | "jpeg" | "png" | "webp";
+type ApplicationSubType = "pdf";
```
SovereignAndrey, line 28: @ap-justin 
Nit: We have "image" subtype. Maybe call the other generic files "documents" or "documentation" subtypes instead of a naming based on the place it's used in? We could have this support files like doc,xlsx,tiff,etc in the future and use it in more places. It would align will with the eventual "video" subtypes needed by Media. 
```suggestion
type DocumentationSubType = "pdf";
```

ap-justin: the keyword `application` is one of the [discreet types](https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/MIME_types#application) where it's subtypes are `pdf, zip, pkc8 ...`, 

other formats like `doc` and `docx ` is also a subtype of `application`
image

SovereignAndrey: Ah! :bulb: _That_ Application. Got it! All good then. :)

SovereignAndrey (approved):
Looks good!

PR #2569, opened 2023-12-05
AP-952: wording changes

ap-justin:
Ticket(s):
-

* NOTE: changes only rendered text, front-end variable names, db attribute names changing could be done on another PR

-

## UI changes for review

### DIFF
--- a/src/pages/Admin/Charity/Dashboard/Dashboard.tsx
+++ b/src/pages/Admin/Charity/Dashboard/Dashboard.tsx
@@ -12,7 +12,7 @@ export default function Dashboard() {
 
   return (
     <div className="@container w-full max-w-4xl grid content-start mt-6">
-      <Seo title="Charity Dashboard" />
+      <Seo title="Nonprofit Dashboard" />
       <h3 className="uppercase font-extrabold text-2xl mb-4">Dashboard</h3>
       <QueryLoader
         queryState={queryState}
--- a/src/App/constants.ts
+++ b/src/App/constants.ts
@@ -15,7 +15,7 @@ type LINKS = {
 
 export const CHARITY_LINKS: LINKS = {
   HEADER_LINKS: [
-    { title: "For nonprofits", href: BASE_URL, external: true },
+    { title: "For Nonprofits", href: BASE_URL, external: true },
     {
       title: "Marketplace",
       href: appRoutes.marketplace,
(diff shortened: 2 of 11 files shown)


src/pages/Admin/Charity/Dashboard/Dashboard.tsx
```diff
@@ -12,7 +12,7 @@ export default function Dashboard() {
 
   return (
     <div className="@container w-full max-w-4xl grid content-start mt-6">
-      <Seo title="Charity Dashboard" />
+      <Seo title="Nonprofit Dashboard" />
```
misicnenad, line 15: Do we need to update "Endowment Designation"' Selector's `"Charity"` value as well?
See:
- https://github.com/AngelProtocolFinance/angelprotocol-web-app/blob/master/src/types/aws/ap/index.ts#L30
- https://github.com/AngelProtocolFinance/angelprotocol-web-app/blob/master/src/pages/Admin/Charity/EditProfile/Form.tsx#L26
- https://github.com/AngelProtocolFinance/angelprotocol-web-app/blob/master/src/pages/Marketplace/Filter/Designations.tsx#L7

misicnenad, line 15: @SovereignAndrey are we dropping the term "charity" completely?
If so, it would be prudent to update our internal variable/field names from `*charity*` to `*nonprofit*` to avoid confusion later (e.g. `charityName -> nonprofitName`, `CHARITY_LINKS -> NONPROFIT_LINKS` etc.)

misicnenad (changes requested):

src/pages/Admin/Charity/Dashboard/Dashboard.tsx
ap-justin: > Do we need to update "Endowment Designation"' Selector's "Charity" value as well?
I guess `charity` is one of nonprofit designations, so didn't touch that one

ap-justin: we could change everything in frontend, but would eventually come into inconsistency with backend as we can't change db `attribute` names easily

SovereignAndrey: As I understand, yes, we're moving to the more inclusive "nonprofit" term. We can and should move to change internal vars and such too, but it's def a low priority thing (esp with the issues we'll have changing the DB fields as Justin mentioned. Wen Postgres... :sweat: ). :+1:

SovereignAndrey (approved):

PR #2574, opened 2023-12-06
Amplify setup envs

SovereignAndrey:
Ticket(s):
-

sets up `staging` and `production` backend environments and links relevant Cognito Auth User Pool to each. 

-

## UI changes for review


src/aws-exports.js
```diff
@@ -8,16 +8,16 @@ const awsmobile = {
     "aws_user_pools_id": "us-east-1_ukOlQeQIM",
     "aws_user_pools_web_client_id": "4lkqi6s5qp6g48tot3em1cabkt",
     "oauth": {
-        "domain": "oauth.better.giving",
+        "domain": "gc8hpcg5cpes-prod.auth.us-east-1.amazoncognito.com",
         "scope": [
-            "phone",
+            "aws.cognito.signin.user.admin",
```
misicnenad, line 13: Presumably these reorders are not just to make the code "prettier"?

.gitignore
```diff
@@ -25,7 +25,7 @@ codecov-report.json
 
 
 package-lock.json
-
+amplify-backup/
```
misicnenad, line 28: Does this get created by the Amplify CLI?

amplify/backend/auth/bettergivingwebapp59590dc1/parameters.json
```diff
@@ -0,0 +1,31 @@
...
+    "Requires Symbols"
+  ],
+  "mfaConfiguration": "OPTIONAL",
+  "autoVerifiedAttributes": [
+    "email"
+  ],
+  "mfaTypes": [
+    "TOTP"
+  ]
+}
```
misicnenad, line 31: nit: please add missing newline characters in to all PR files that are missing it

misicnenad (changes requested):

.gitignore
SovereignAndrey: Yes, I gitignored it since our build was complaining about it's existence. Digging into the why now. I think ultimately it can stay in and I'll need to see about modifying the `amplify.yml`. :crossed_fingers:

ap-justin (dismissed):

src/store/store.ts
```diff
@@ -25,11 +22,6 @@ export const store = configureStore({
     getDefaultMiddleware().concat([aws.middleware, apes.middleware]),
 });
 
-config.oauth.redirectSignIn =
-  window.location.origin + `${appRoutes.auth_redirector}/`;
-config.oauth.redirectSignOut = window.location.origin + "/";
-Amplify.configure(config);
```
ap-justin, line 31: @SovereignAndrey, `Amplify` need to be configured first before being used (required by `<Authenticator/>` component )

src/store/store.ts
```diff
@@ -25,11 +22,6 @@ export const store = configureStore({
     getDefaultMiddleware().concat([aws.middleware, apes.middleware]),
 });
 
-config.oauth.redirectSignIn =
-  window.location.origin + `${appRoutes.auth_redirector}/`;
```
ap-justin, line 29: when using legacy config from `aws-pull`, oauth signin gets error when fed with csv redirectURLs, so before being passed to `Amplify.configure(..)` we replaced this csv with a single link. 

I'm not sure if this is still the case with `aws-amplify` v6, but it's good to note that this behavior exist

ap-justin (changes requested):

src/store/store.ts
SovereignAndrey: @ap-justin That context helps, thanks. This bit was all removed due errors around missing `config.oauth`. Still trying to figure out a way around that. Revert this group of commits, but builds will fail now due to aforementioned error.

PR #2575, opened 2023-12-06
AP-885: Prefill the "expected monthly donations" amount field so that the form is loaded on page load

misicnenad:

- comment out:
   - `Authenticator.Provider` wrapper in *App.tsx*
   - all user setup code in *Auth.tsx* except for the context; use the following `AuthenticatedUser` placeholder:
   ```ts
   const user: AuthenticatedUser = {
      email: "some@email.com",
      endowments: [],
      groups: [],
      isSigningOut: false,
      token: "",
   };
   ```
- go to registration step 5 (banking)
- **verify the form displays even when no value is set for "expected monthly donations"**

## UI changes for review

image

### DIFF
--- a/src/components/BankDetails/BankDetails.tsx
+++ b/src/components/BankDetails/BankDetails.tsx
@@ -47,8 +47,8 @@ function Content({
   isSubmitting,
   onSubmit,
 }: Omit<Props, "shouldUpdate" | "onInitiateUpdate">) {
-  const [expectedMontlyDonations, setExpectedMontlyDonations] =
-    useState<number>();
+  const [expectedMontlyDonations, setExpectedMontlyDonations] = useState(10000); // in USD
+
   const [debounce, isDebouncing] = useDebounce();
 
   const { currencies, isLoading, targetCurrency, setTargetCurrency } =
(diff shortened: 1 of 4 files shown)


src/components/BankDetails/BankDetails.tsx
```diff
@@ -79,35 +87,28 @@ function Content({
...
+          // but still call the function to cancel the previous debounce call
+          const delay = newValue === expectedMontlyDonations ? 0 : 1000;
+          debounce(() => setExpectedMontlyDonations(newValue), delay);
         }}
       />
 
-      {/* Display disabled form buttons by default, this is necessary 
-          to be able to show "Back" button during registration */}
-      {!expectedMontlyDonations ? (
-        <FormButtons disabled refreshRequired />
```
misicnenad, line 92: `expectedMonthlyDonations` will always be defined, so this will never show

src/components/BankDetails/BankDetails.tsx
```diff
@@ -79,35 +87,28 @@ function Content({
...
-      ) : (
-        <>
-          <Divider />
-          <div className="min-h-[20rem]">
```
misicnenad, line 96: Moved this `min-height` container inside `RecipientDetails`

PR #2595, opened 2023-12-13
FIX: reg summary/dashboard not disabled after submission

ap-justin:
* changes in https://github.com/AngelProtocolFinance/devops/pull/394, not reflected in web-app
* since `EndowmentId` is not returned on `registration-submit` use `Email` instead as indicator that application is approved
https://github.com/AngelProtocolFinance/devops/blob/main/src/AP/lambdas/registration-submit/index.js#L37

* fix 2: check if user signed `FSA` (if applicable) before doc considered complete

TODO:
@jp-mariano, noticed that `EndowmentId` is no longer reflected on  registration record upon approval. Web-app still needs this attribute to redirect users to `profile/{endowId}` on successful registration prompt

-

## UI changes for review

### DIFF
--- a/src/types/aws/ap/registration.ts
+++ b/src/types/aws/ap/registration.ts
@@ -235,7 +235,12 @@ export function isDoneFSAInquiry(
 }
 
 export function isDoneDocs(data: SavedRegistration): data is DoneDocs {
-  return !!(data.Registration as TDocumentation).Documentation;
+  const reg = data.Registration as Partial<TDocumentation>;
+
+  if (reg.Documentation?.DocType === "FSA") {
+    return !!reg.Documentation.SignedFiscalSponsorshipAgreement;
+  }
+  return !!reg.Documentation;
 }
 
 export function isDoneBanking(data: SavedRegistration): data is DoneBanking {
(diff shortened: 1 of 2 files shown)


src/types/aws/ap/registration.ts
```diff
@@ -137,7 +137,7 @@ export type DoneFSAInquiry = Append<DoneOrgDetails, FSAInquiry, {}>;
 export type DoneDocs = Append<DoneFSAInquiry, TDocumentation, {}>;
 export type DoneBanking = Append<DoneDocs, BankingDetails, {}>;
 
-export type SubmissionDetails = { Email: string; EndowmentId: number };
+export type SubmissionDetails = { Email: string; EndowmentId?: number };
```
ap-justin, line 140: probably best to add `Approved` state type after `InReview` to make things clearer but this would require more changes - could be in another PR

jp-mariano:
I have added a commit in AngelProtocolFinance/devops#408 that addresses the `EndowmentId` issue and also updated the Lambda `admin-approvals`.

SovereignAndrey (approved):

PR #2596, opened 2023-12-13
Fix: edge case - handle unfinished FSA

ap-justin:
`isDoneDocs` passes if FSA information (ids, proof of registration) is saved - even if FSA agreement is not signed. 

after
* #2595

`isDoneDocs` enforces FSA agreement to be signed before state is registered to be `4`. This caused saved FSA information to not be shown. 

As a middle ground - always show saved FSA information but don't allow going to banking details if not signed.

Expected edge case flow :
1. user registers and needs FSA
2. user generates signing URL agreement, and redirected there to sign
3. for some reason, user didn't sign the FSA and went back to registration
4. upon resuming, user will be directed first to step `3`(FSA Inquiry) and is free to go to step `4` (FSA docs) - which is prefilled with previously inputed data. However user can't go to step `5` (banking) as registration state is still `3`

5. since user already generated FSA signing URL - he will always be directed to this previous URL until he signs it. FSA url can be re-created if it times out or user explicitly opt out of FSA (clears FSA docs) and opt-in again. 

-

## UI changes for review

### DIFF
--- a/src/pages/Registration/Steps/getRegistrationState.ts
+++ b/src/pages/Registration/Steps/getRegistrationState.ts
@@ -63,9 +63,13 @@ export function getRegistrationState(reg: SavedRegistration): {
 
   if (isDoneDocs(reg)) {
     const { ContactPerson: c, Registration: r } = reg;
+    const isSignedFSA =
+      r.Documentation.DocType === "FSA"
+        ? !!r.Documentation.SignedFiscalSponsorshipAgreement
+        : true;
     return {
       state: {
-        step: 4,
+        step: (isSignedFSA ? 4 : 3) as 4,
         data: {
           init: initReg(c),
           contact: { ...c, orgName: r.OrganizationName },
--- a/src/pages/Registration/Steps/getRegistrationState.ts
+++ b/src/pages/Registration/Steps/getRegistrationState.ts
@@ -67,14 +67,17 @@ export function getRegistrationState(reg: SavedRegistration): {
       r.Documentation.DocType === "FSA"
         ? !!r.Documentation.SignedFiscalSponsorshipAgreement
         : true;
+
     return {
       state: {
+        //cast to 4 (type only) to conform to type `RegStep4` which accepts documentation
         step: (isSignedFSA ? 4 : 3) as 4,
         data: {
           init: initReg(c),
           contact: { ...c, orgName: r.OrganizationName },
           orgDetails: orgDetails(r),
           fsaInquiry: fsaInquiry(r),
+          //even step is has value of `3` user could still go to step 4 with documentation pre-filled from previous uploads
           documentation: docs(r),
         },
       },
--- a/src/types/aws/ap/registration.ts
+++ b/src/types/aws/ap/registration.ts
@@ -235,12 +235,7 @@ export function isDoneFSAInquiry(
 }
 
 export function isDoneDocs(data: SavedRegistration): data is DoneDocs {
-  const reg = data.Registration as Partial<TDocumentation>;
-
-  if (reg.Documentation?.DocType === "FSA") {
-    return !!reg.Documentation.SignedFiscalSponsorshipAgreement;
-  }
-  return !!reg.Documentation;
+  return !!(data.Registration as TDocumentation).Documentation;
 }
 
 export function isDoneBanking(data: SavedRegistration): data is DoneBanking {


src/types/aws/ap/registration.ts
```diff
@@ -235,12 +235,7 @@ export function isDoneFSAInquiry(
 }
 
 export function isDoneDocs(data: SavedRegistration): data is DoneDocs {
-  const reg = data.Registration as Partial<TDocumentation>;
```
ap-justin, line 238: revert to - `docs` is complete if user uploaded all necessary files,info to generate signing URL

src/pages/Registration/Steps/getRegistrationState.ts
```diff
@@ -63,9 +63,13 @@ export function getRegistrationState(reg: SavedRegistration): {
...
   if (isDoneDocs(reg)) {
     const { ContactPerson: c, Registration: r } = reg;
+    const isSignedFSA =
+      r.Documentation.DocType === "FSA"
+        ? !!r.Documentation.SignedFiscalSponsorshipAgreement
+        : true;
     return {
       state: {
-        step: 4,
+        step: (isSignedFSA ? 4 : 3) as 4,
```
ap-justin, line 72: only mark `4` is signed - allowing user to go to banking step.

src/pages/Registration/Steps/getRegistrationState.ts
```diff
@@ -63,9 +63,13 @@ export function getRegistrationState(reg: SavedRegistration): {
 
   if (isDoneDocs(reg)) {
     const { ContactPerson: c, Registration: r } = reg;
+    const isSignedFSA =
```
misicnenad, line 66: nit:
```suggestion
    const isSignedFSA = r.Documentation.DocType !== "FSA" || !!r.Documentation.SignedFiscalSponsorshipAgreement;
```

src/pages/Registration/Steps/getRegistrationState.ts
misicnenad: I understand that this works, but... casting `3 as 4` is just weird...
Looks like something that would be a part of an article titled "Bizarre code I've seen in my career" haha

misicnenad: Casting `3 as 4` looks like something that would be make it into an article titled "Bizarre code I've seen in my career" haha
I understand that this works, but... it's just weird... Is there a way to make this work without such a hack?

misicnenad (changes requested):

src/pages/Registration/Steps/getRegistrationState.ts
SovereignAndrey: @ap-justin @misicnenad At the very very least, should we leave that all as is, def add have a comment explaining what's going on and why we are doing that. We'll thank ourselves 3 months later when we revisit it or a new team member starts working on it. :)

src/pages/Registration/Steps/getRegistrationState.ts
ap-justin: saves a few character of code but I can't understant it the first time I read it 

original:
FSA is signed  - `DocType` is `FSA` ? check if FSA is signed , otherwise no need to check

suggested:
FSA is signed - `DocType` is `not` `FSA` `or` `FSA` is signed

I also noticed you changed a lot of these type constructs on your previous PR, which I believe is not necessary

src/pages/Registration/Steps/getRegistrationState.ts
ap-justin: > Casting 3 as 4 looks like something that would make it into an article titled "Bizarre code I've seen in my career" haha

`as 4` is type construct and doesn't actually make `3` transform to `4` (actual value) , it's needed to conform this particular block to 

```ts
type RegStep4 = {
  step: 4;
  data: Step4Data;
};
```
which accepts `documentation` , if I don't put this casting this step would of of type `3 | 4` which doesn't accept `documentation`. This would mean even documentation is previously done (but not signed), we can't put it in `getRegistrationState`

> . it's just weird... Is there a way to make this work without such a hack?

Of course I am thinking of a better way to handle this edge case, but considering this is on production - can't afford  to have it there for a long time, that's why I presented the least intrusive and fastest solution I know.  


> @ap-justin @misicnenad At the very very least, should we leave that all as is, def add have a comment explaining what's going on and why we are doing that. 

@SovereignAndrey, would add some temp comments.

src/pages/Registration/Steps/getRegistrationState.ts
misicnenad: It's just a suggestion, you don't have to implement it.

> I also noticed you changed a lot of these type constructs on your previous PR, which I believe is not necessary

I'd agree, in retrospect maybe it wasn't necessary. Had you or Andrey pointed this out, I've would've removed those changes from the refactor PR

src/pages/Registration/Steps/getRegistrationState.ts
```diff
@@ -63,14 +63,21 @@ export function getRegistrationState(reg: SavedRegistration): {
...
       state: {
-        step: 4,
+        //cast to 4 (type only) to conform to type `RegStep4` which accepts documentation
+        step: (isSignedFSA ? 4 : 3) as 4,
         data: {
           init: initReg(c),
           contact: { ...c, orgName: r.OrganizationName },
           orgDetails: orgDetails(r),
           fsaInquiry: fsaInquiry(r),
+          //even step is has value of `3` user could still go to step 4 with documentation pre-filled from previous uploads
```
misicnenad, line 80:
```suggestion
          // even if step has value of `3` user could still go to step 4 with documentation pre-filled from previous uploads
```

misicnenad (approved):

PR #2597, opened 2023-12-14
BG-1002: update network param `testnet | mainnet` to `staging | production`

ap-justin:
reflects: 
* https://github.com/AngelProtocolFinance/devops/pull/417

* replace `network` with `apiEnv`
- [x] test endpoints in use that is changed to use `apiEnv` to still work

-

## UI changes for review

### DIFF
--- a/src/services/apes/apes.ts
+++ b/src/services/apes/apes.ts
@@ -3,7 +3,7 @@ import { Token } from "types/aws";
 import { ChainID } from "types/chain";
 import { appRoutes } from "constants/routes";
 import { APIs } from "constants/urls";
-import { network } from "../constants";
+import { apiEnv } from "../constants";
 import { version as v } from "../helpers";
 import { tags } from "./tags";
 
@@ -27,7 +27,7 @@ export const apes = createApi({
     stripeSessionURL: builder.mutation<{ url: string }, StripeSessionURLParams>(
       {
         query: ({ endowId, liquidSplitPct }) => ({
-          url: `${v(1)}/fiat/stripe-proxy/apes/${network}`,
+          url: `${v(1)}/fiat/stripe-proxy/apes/${apiEnv}`,
           method: "POST",
           body: JSON.stringify({
             endowmentId: endowId,
--- a/src/constants/env.ts
+++ b/src/constants/env.ts
@@ -1,5 +1,4 @@
 import isMobile from "is-mobile";
-import { NetworkType } from "types/lists";
 
 export const IS_MOBILE = isMobile();
 
@@ -16,6 +15,3 @@ export const AWS_S3_PUBLIC_BUCKET = "https://endow-profiles.s3.amazonaws.com";
 const ENV = process.env.REACT_APP_ENVIRONMENT;
 
 export const IS_TEST = ENV === "STAGING";
-export const EXPECTED_NETWORK_TYPE: NetworkType = IS_TEST
-  ? "testnet"
-  : "mainnet";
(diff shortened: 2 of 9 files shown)


jp-mariano (approved):

src/services/apes/apes.ts
```diff
@@ -27,7 +27,7 @@ export const apes = createApi({
     stripeSessionURL: builder.mutation<{ url: string }, StripeSessionURLParams>(
       {
         query: ({ endowId, liquidSplitPct }) => ({
-          url: `${v(1)}/fiat/stripe-proxy/apes/${network}`,
+          url: `${v(1)}/fiat/stripe-proxy/apes/${apiEnv}`,
```
misicnenad, line 30: @ap-justin @jp-mariano does this mean that [version(v: number)](https://github.com/AngelProtocolFinance/angelprotocol-web-app/blob/46c5cd78580bee4b20ad3784f41f78973977ef3a/src/services/helpers.ts#L8) is redundant and that both prod and staging endpoints will be starting with the version part (v1, v2 et.)?
So instead of these being the prod and staging endpoints:
```
staging/fiat/stripe-proxy/apes/staging
v1/fiat/stripe-proxy/apes/production
```
it would be:
```
v1/fiat/stripe-proxy/apes/staging
v1/fiat/stripe-proxy/apes/production
```
The former now makes little sense (`/staging/.../staging` is redundant), while the latter allows for versioning the staging lambdas too (not just prod ones).

misicnenad (changes requested):

src/services/apes/apes.ts
jp-mariano: This will be the case for the meantime. The `staging` version of the endpoints were made because it was a necessity back then. Those endpoints were then updated to handle the current processes and logic. This is just a band aid fix that addresses the issues that will be encountered due to us having a live endpoint and the change from using `testnet/mainnet` to `staging/production`.

Moving forward, we can go back to the old type of API versioning we're previously using. For example, `stripe-proxy` will be updated to use `PaymentIntent` instead of `CheckoutSession`. We can capitalize on this by changing/refactoring the Lambda's logic to handle `v{n}` versioned APIs.

ap-justin: @misicnenad, in this url `${v(1)}/fiat/stripe-proxy/apes/${apiEnv}`
`${v(1)}` still pertains to `v1/staging` path configured in API Gateway

we could certainly remove this, in favor of `apiEnv` but that would also require creating new routes and re-attaching lambdas, authorizers etc.. 

while `apiEnv` is a replacement for `network = testnet | mainnet` use as `chain_id` key to db tables

jp-mariano: @misicnenad and @ap-justin ,
Moving forward, if we're going to work on new versions of Lambdas, we can use the format `/v{n}/.../{network}`. Where `network` can either be staging or production. There's no problem to create new routes and attaching authorizers since they will be attached to new Lambda versions and will be used for testing, initially. This (`/v{n}/.../{network}`) was the format I was referring to when I said:
> Moving forward, we can go back to the old type of API versioning we're previously using.

This was our old way of doing things because back then we only had to think about mainnet and testnet. As we move along, we can sunset the use of `staging`-versioned endpoints (`/staging/.../{network}`) for good.

What I believe this PR addresses is the live endpoints that are still using the `mainnet` network parameter.

ap-justin: @misicnenad, I updated the PR title to make it clear that this intends to reflect only aws updates with `{network}` path parameter

misicnenad (approved):

SovereignAndrey (approved):

PR #2598, opened 2023-12-14
BG-958: Stripe Embedded form for credit card payments on existing "credit/debit card" tab

misicnenad:
Related to https://github.com/AngelProtocolFinance/devops/pull/420

- open *.env* and add:
```.env
PUBLIC_STRIPE_KEY=pk_test_51MGl2PJBsJzBirEIi04vNpC8N1pvsmDb8zUHwBV0O47wLoi3atillyJSEz0Syb7dGSOoNEZmFEjZzkKVsWEOI5AP00MP1H0S2R
```
- go to http://localhost:4200/donate/53
- insert any amount in "donation amount" field
- click "Pay with Card"
- wait for Stripe form to load
- insert test card data (see docs https://stripe.com/docs/testing)
- click "Pay Now"
- **verify "Thank you for your donation" page appears**
- **verify the flow works when accessed through a widget (example URL http://localhost:4200/donate-widget/33?isDescriptionTextShown=true&advancedOptionsDisplay=collapsed&liquidSplitPct=50)**
- **verify that submitted values get pre-loaded when switching between Form, KYC and Checkout steps**

## UI changes for review

Desktop:
Authenticated (no email field):
image
Non-authenticated (has email field):
image
image
(note: if authenticated, the default email value here will be the user's email; this value can be changed)
image

Mobile:
Authenticated (no email field):
image
Non-authenticated (has email field):
image
image
image

When loading the checkout page (note: left the "back" button so that the user can still navigate back if needed):
image


src/components/donation/Steps/DonateMethods/Stripe/Form.tsx
```diff
@@ -0,0 +1,119 @@
...
+      handleError(err, "Failed to load payment platform");
+    }
+  });
+
+  return (
+    <FormProvider {...methods}>
+      <form onSubmit={submit} className="grid gap-4">
+        <Field<FormValues>
+          name="amount"
+          label="Donation amount"
```
ap-justin, line 58: should add hint that this is in `usd`

misicnenad: Fixed https://github.com/AngelProtocolFinance/angelprotocol-web-app/pull/2598/commits/c27ebc5e18130c772c49717638166f6d817cbe83

.env.example
```diff
@@ -6,6 +6,5 @@ PORT=<port number>
 
 REACT_APP_ENVIRONMENT=<STAGING | PRODUCTION>
 
-
-
-
+REACT_APP_PUBLIC_STRIPE_KEY_TEST=""
```
ap-justin, line 9: we also need to add this in `amplify.yml`
like
```
- REACT_APP_SENTRY_DSN=${REACT_APP_SENTRY_DSN}
```

I'll add this envs to aws amplify

src/services/apes/apes.ts
```diff
@@ -23,26 +22,35 @@ export const apes = createApi({
...
+      query: ({ paymentIntentId }) => ({
+        url: `v2/fiat/stripe-proxy/apes/staging?paymentIntentId=${paymentIntentId}`,
+      }),
+    }),
+    createStripePaymentIntent: builder.mutation<
+      { clientSecret: string },
+      StripePaymentIntentParams
+    >({
+      query: ({ amountInCents, endowmentId, liquidSplitPct }) => ({
+        url: `v2/fiat/stripe-proxy/apes/staging`,
```
ap-justin, line 38:
```suggestion
        url: `v2/fiat/stripe-proxy/apes/{staging}`,
```
also add `production`

ap-justin (changes requested):

src/services/apes/apes.ts
```diff
@@ -23,26 +22,35 @@ export const apes = createApi({
...
-            endowmentId: endowId,
-            splitLiq: liquidSplitPct,
-            redirectUrl: `${window.location.origin}${appRoutes.donate_fiat_thanks}`,
-          }),
+    getStripePaymentStatus: builder.query<
+      Pick<PaymentIntent, "status">,
+      { paymentIntentId: string }
+    >({
+      query: ({ paymentIntentId }) => ({
+        url: `v2/fiat/stripe-proxy/apes/staging?paymentIntentId=${paymentIntentId}`,
```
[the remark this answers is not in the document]
misicnenad: :man_facepalming:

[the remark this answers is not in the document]
misicnenad: Added

ap-justin (dismissed):

SovereignAndrey (dismissed):
@misicnenad This is looking very good. Only one thing I noticed, not your fault since it was not on the original COS, but we should be sure the new embedded flow for credit/debit is asking if user wants a tax receipt on the first screen, just like the crypto flow (ie. the little checkbox that causes the KYC page to show up before a final summary and submission page), so they too can get a tax receipt.

misicnenad:
Converting to draft until KYC is implemented.

misicnenad:
Ready for review @SovereignAndrey @ap-justin

src/components/donation/Steps/DonateMethods/Stripe/Form.tsx
```diff
@@ -0,0 +1,166 @@
...
+
+  useEffect(() => {
+    fetch("https://ipapi.co/json/")
+      .then<{ country_code: string }>((response) => response.json())
+      .then((data) => setCountryCode(data.country_code))
+      .catch((err) => handleError(err, GENERIC_ERROR_MESSAGE));
+  });
+
+  return (
+    <QueryLoader
```
ap-justin, line 45: there's brief error before form gets loaded
REPRO:
go to other tab,
go back to credit/debit tab

image

src/components/donation/Steps/DonateMethods/Stripe/Stripe.tsx
```diff
@@ -0,0 +1,125 @@
...
+              setStep({ ...step, type: "checkout", clientSecret, ...fv });
+            } catch (err) {
+              handleError(err, "Failed to load payment platform");
+            }
+          }}
+        />
+      );
+    case "kyc":
+      return (
+        <KYCForm
```
ap-justin, line 50: kyc looks cramped

image

src/components/donation/Steps/DonateMethods/Stripe/Form.tsx
```diff
@@ -0,0 +1,166 @@
...
+  });
+
+  const isInsideWidget = widgetConfig !== null;
+
+  const submit = methods.handleSubmit(onSubmit);
+
+  return (
+    <FormProvider {...methods}>
+      <form onSubmit={submit} className="grid gap-4">
+        <Field<FormValues>
```
ap-justin, line 81: I tried to donate `$0.01` and gets error - should add min amount validation

image

src/components/donation/Steps/DonateMethods/Stripe/Form.tsx
```diff
@@ -0,0 +1,166 @@
...
+  const [countryCode, setCountryCode] = useState<string>();
+
+  const queryState = useStripeCurrenciesQuery(
+    { country_code: countryCode || "" },
+    { skip: !countryCode }
+  );
+  const { handleError } = useErrorContext();
+
+  useEffect(() => {
+    fetch("https://ipapi.co/json/")
```
ap-justin, line 38: if `countryCode` is dependence of `stripeCurrency`, should merge to one query like `aws.bankingApplications.bankingApplication` , so that we don't fetch on every render

ap-justin (approved):

src/components/donation/Steps/DonateMethods/Stripe/Form.tsx
misicnenad: Good catch.

I'd agree that the error modal can be made more clear/helpful by showing this specific error. And although validating a field even before the submission is made would be even better, there is a drawback, namely that we need to get an exchange rate dynamically.

To do this we need to get the current exchange rate between whatever currency was loaded for the users location and USD. Stripe has no API that would allow reading such a value, as they depend on 3rd parties for this data, so we'd have to access some 3rd party as well. 
I couldn't find a free API that would serve our purposes, only payed ones and these were the most often recommended ones:
- https://exchangeratesapi.io - pricing goes from $10/month for 10,000 API calls per month to $80/month for 500,000 API calls per month
- https://openexchangerates.org - $12/month for 10,000 API calls -> $97/month for unlimited API calls

@SovereignAndrey @ap-justin @jp-mariano do you know of any other APIs?

@SovereignAndrey so these are the 2 options for handling the problem:
1. After submit show the error modal stating "Amount must be at least $0.50 usd"
   - pros: cheap, easy to implement
   - cons: no field validation (better said, postponed field validation), the same error would show for every currency requiring the user to look up the current exchange rate themselves before inputting a value
2. Use a 3rd party API to get the current exchange rate
   - pros: field validation, minimum amount can be tailored to the selected currency ("Amount must be at least 51 RSD" or "...156 GRD)
   - cons: paid service

src/components/donation/Steps/DonateMethods/Stripe/Stripe.tsx
misicnenad: It does look ugly.
Copied classes used in donation `CurrentStep.tsx`, looks much better.
image

src/components/donation/Steps/DonateMethods/Stripe/Form.tsx
SovereignAndrey: @misicnenad If we tweak the setup and make the assumption that we don't nesessarily need live rates for checkout, this becomes more manageable. 

I'd suggest we use OpenExchangeRates.org (herein: OER) for a source of FX rates. We can do up to 1000 requests per month for their free tier. After that, it's $10/month for up to 10K requests. 

Periodically, an APES AWS Lambda (via a 2-3x day EventBridge trigger) would hit the OER API and update a DB table with the latest rates.

We would have an AWS endpoint for the web app to query the most recent FX rate for a given currency from AWS DB. This rate would be recent enough I think for most cases.

At 3x a day for 10 currencies, that's under the 1000 / month free tier limit! I think we can get away with 2x a day, since fiat FX rates do not move nearly as much as crypto. Wdyt?

package.json
```diff
@@ -55,6 +55,8 @@
     "@keplr-wallet/wc-qrcode-modal": "0.12.23",
     "@reduxjs/toolkit": "1.9.7",
     "@sentry/react": "7.81.0",
+    "@stripe/react-stripe-js": "^2.4.0",
```
ap-justin, line 58: nit: preferrably exact version, as it's easy to see which we are running without looking at the lock file.

src/components/CurrencySelector.tsx
```diff
@@ -44,21 +45,25 @@ function CurrencySelector(props: Props) {
...
+        className={`relative items-center grid grid-cols-[1fr_auto] field-container ${
+          props.classes?.combobox || ""
+        }`}
       >
         <Combobox.Input
           ref={inputRef}
           className="w-full border-r border-gray-l3 dark:border-bluegray px-4 py-3.5 text-sm leading-5 text-gray-900 focus:ring-0"
           displayValue={(currency: WiseCurrency) =>
-            `${currency.code} - ${currency.name}`
+            !!currency.name
```
ap-justin, line 64: displayValue should be typed as `Currency` instead of `WiseCurrency`

ap-justin (approved):

src/components/donation/Steps/DonateMethods/Stripe/Form.tsx
jp-mariano: We could use Wise for forex rates too. For reference, https://docs.wise.com/api-docs/api-reference/rate

misicnenad: This would work. It would be even better if we were querying the exchange rates, but storing the minimum donation amounts (MDA) in this DB table. That way our webapp can simply check and validate the amount the user has input in the donation field.

> We would have an AWS endpoint for the web app to query the most recent FX rate for a given currency from AWS DB. This rate would be recent enough I think for most cases.
> 
> At 3x a day for 10 currencies...

No need to query for multiple currencies, we only need the current exchange rate for USD which returns *exchange rates for all currencies against USD* (see https://docs.openexchangerates.org/reference/api-introduction#api-response-formats).
We could query 3x/day only against USD, use the exchange rates to recalculate MDA and store those values in a table, so the webapp can query MDA against this table for any possible currency.

Edge cases:
1. the client wants to donate in a currency for which we do not have a minimum amount as _openexchangerates_ didn't provide us with the exchange rate.
   - **solution:** we stop using _openexchangerates_ as they should support all possible currencies; we switch to another provider.
2. the exchange rate changes sufficiently enough for a given currency to cause the "Amount must be at least $0.50 usd" error to be thrown in certain cases, until our lambda updates MDA table; this might not even be a black swan event, just a natural change of the rate.
   - **solution:** for the error to appear in such cases would require a user to want to donate the bare minimum in their currency, which is not going to happen that often anyway. If it does then it would be understandable to show the error to users. We might implement certain steps to handle this though, as it would be bad UX to keep showing this error to all users for a couple of hours, before our lambda updates the rates table.
   Potential steps:
      - **set MDA to always be a bit higher than the $0.50 rate** -> if calculate the current exact MDA for PHP to be 27.85 (~$0.50), we increase it by 10% just to be sure and store 30 in our DB; we do this for every currency. This measure should be enough to avoid MDA errors as the exchange rate rarely changes 10% or more in a single day (even Venezuelan Bolivar doesn't change this drastically); we could increase the value by even less, maybe 5%.
      NOTE: we could round this value down, as showing a message like "Mimimum donation amount is 30" is much more pleasant than "Mimimum donation amount is 30.7615".
      - **if a user experiences the "Amount must be at least $0.50 usd" error has occurred, we automatically trigger the MDA updater lambda** -> if upon submitting the donation we see that the error above has occurred, this must mean there was a wild swing in exchange rate(s) and that we should immediately trigger the MDA updater lambda to update the MDA table; we do this so that future potential donors **do not** experience the same error. Despite this, the MDA updater lambda continues triggering as scheduled.
      
NOTE:  **we should add a message directly on screen or as part of an info popup explaining why MDA might be different than the last time they were donating** -> users might find it confusing to see that the minimum donation amount is 30 when the day before it was 29; adding a message explaining the relationship between MDA and the current exchange rate would help them understand that MDA didn't just change because we're greedy.

misicnenad: @SovereignAndrey

src/components/CurrencySelector.tsx
misicnenad: good catch

src/components/donation/Steps/DonateMethods/Stripe/Form.tsx
SovereignAndrey: > if we were querying the exchange rates, but storing the minimum donation amounts (MDA) in this DB table

@misicnenad If we store __both__ items, then we can validate MDA is hit and also can show an FX rate in UI (if it becomes important in the future).

misicnenad: > @SovereignAndrey , @misicnenad We could use Wise for forex rates too. For reference, https://docs.wise.com/api-docs/api-reference/rate

@jp-mariano missed this comment as GH doesn't update comment threads in real-time, only on reload.

I think this solves our problem then. We could still have a special lambda that updates the MDA table just to make fetching MDA faster. Then we also update [stripe-currencies](https://us-east-1.console.aws.amazon.com/lambda/home?region=us-east-1#/functions/stripe-currencies?tab=code) to fetch MDAs for all supported currencies based on user's locale and voila.

@SovereignAndrey 
> If we store both items, then we can validate MDA is hit and also can show an FX rate in UI (if it becomes important in the future).
> ...
> Ex: The minimum donation amount is 1 USD or 420.69 NZD (calculated based on a rate of 841.38 NZD/USD).

We can do that.

> We could start with US$1 as our internal MDA for fiat...

Sounds good :+1:

SovereignAndrey:
@misicnenad Please address the merge conflicts. Thanks!

misicnenad:
> @misicnenad Please address the merge conflicts. Thanks!

@SovereignAndrey will do so and will also convert this to draft until the "minimum donation amount" logic is setup (finishing up the responsible lambda, creating a table and then implementing the appropriate field validation in this PR, after which it will be reopened).
See https://github.com/AngelProtocolFinance/angelprotocol-web-app/pull/2598#discussion_r1444436448

src/components/CurrencySelector/CurrencyOptions.tsx
```diff
@@ -1,35 +1,16 @@
...
 import { isEmpty } from "helpers";
 
 type Props = {
   classes?: string;
-  searchText?: string;
+  currencies: Currency[];
 };
 
-export default function Options({ classes = "", searchText = "" }: Props) {
-  const { data: currencies = [] } = useCurrencisQuery(
```
misicnenad, line 11: Loading currencies inside `CurrencySelector` makes it impossible to have different sources of currency data - this becomes a problem when we want to load Stripe-supported currencies, as opposed to all Wise currencies.

By allowing the parent to load currencies from whichever source they need and pass them down, we address this problem and make `CurrencySelector` more flexible.

src/components/KYCForm/useKYCForm.ts
```diff
@@ -1,10 +1,10 @@
 import { Props } from "./types";
 import { useModalContext } from "contexts/ModalContext";
-import Receipter from "./index";
+import KYCForm from "./KYCForm";
```
misicnenad, line 3: Align names inside the hook file to increase clarity about what is what

src/components/QueryLoader.tsx
```diff
@@ -35,6 +35,7 @@ export default function QueryLoader<T>({
     );
   }
   if (isError || !data) {
+    if (isError) logger.error(error);
```
misicnenad, line 38: Needed this when I was debugging something, left it here as I found it useful

src/components/donation/Steps/DonateMethods/Stripe/Form.tsx
misicnenad: @SovereignAndrey 
> If we store both items, then we can validate MDA is hit and also can show an FX rate in UI (if it becomes important in the future).
...
Ex: The minimum donation amount is 1 USD or 420.69 NZD (calculated based on a rate of 841.38 NZD/USD).

Take a look at the current tooltip for the minimum donation amount:
image

Do you want me to add the `"... (calculated based on a rate of 107 RSD/USD)"` part as well?

---
Currently we store rates with 2 decimals, but minimum donation amount with rounded to the nearest integer (so 106.91 RSD/USD becomes 107 minimum donation amount). Do you want us to store the minimum donation amounts with 2 decimals as well?

SovereignAndrey: @misicnenad 
> Take a look at the current tooltip for the minimum donation amount:

Looks great! No need to add the other words.

> Do you want us to store the minimum donation amounts with 2 decimals as well?

Agreed! We prob should just use the nearest whole number.

src/components/donation/Steps/DonateMethods/Stripe/Stripe.tsx
```diff
@@ -0,0 +1,126 @@
...
+import { KYC } from "slices/donation";
+import { PUBLIC_STRIPE_KEY } from "constants/env";
+import Checkout from "./Checkout";
+import Form from "./Form";
+
+const stripePromise = loadStripe(PUBLIC_STRIPE_KEY);
+
+// Followed Stripe's custom flow docs
+// https://stripe.com/docs/payments/quickstart
+export default function Stripe(props: Props) {
```
ap-justin, line 17: @misicnenad, I got error when choosing credit/debit
image

```

data
: 
{message: 'Failed to get currencies'}
status
: 
400
```

repro:
`git branch -D stripe-embed` - delete local  

`git pull `RC-v1.9.2` - fetch latest remotes  

`git checkout stripe-embed`  

`yarn && yarn start`  

`go to donate/33`

ap-justin (changes requested):

src/components/donation/Steps/DonateMethods/Stripe/Stripe.tsx
misicnenad: Will be addresses as part of https://linear.app/better-giving/issue/BG-1072/fix-bug-in-fetch-currencies

@SovereignAndrey @ap-justin in the meantime to test things out:
- open [src/services/apes/apes.ts](https://github.com/AngelProtocolFinance/angelprotocol-web-app/blob/stripe-embed/src/services/apes/apes.ts)
- swap out `${country_code}` on line 57 for a hardcoded value of `RS` (this works for me), so that the URL looks like:
``url: `v2/fiat/stripe-proxy/apes/${apiEnv}/currencies/RS`,``
- reload the donation page
- verify everything works.

SovereignAndrey (changes requested):
Blocking bug with loading the fiat tab. See this ticket please: https://linear.app/better-giving/issue/BG-1072/fix-bug-in-fetch-currencies

misicnenad:
> Blocking bug with loading the fiat tab. See this ticket please: https://linear.app/better-giving/issue/BG-1072/fix-bug-in-fetch-currencies

Created a fix for this https://github.com/AngelProtocolFinance/devops/pull/464

src/components/donation/Steps/DonateMethods/Stripe/Stripe.tsx
misicnenad: Created a fix https://github.com/AngelProtocolFinance/devops/pull/464, currency loading should work now

src/components/donation/Steps/DonateMethods/Stripe/Checkout.tsx
```diff
@@ -0,0 +1,94 @@
...
+} from "@stripe/react-stripe-js";
+import { FormEventHandler, useState } from "react";
+import { useErrorContext } from "contexts/ErrorContext";
+import LoadText from "components/LoadText";
+import { GENERIC_ERROR_MESSAGE } from "constants/common";
+import { appRoutes } from "constants/routes";
+
+// Code inspired by React Stripe.js docs, see:
+// https://stripe.com/docs/stripe-js/react#useelements-hook
+export default function Checkout({ onBack }: { onBack: () => void }) {
```
ap-justin, line 14: after specifying amount and currency, the next screen is blank
image

ap-justin (changes requested):

src/components/donation/Steps/DonateMethods/Stripe/Checkout.tsx
misicnenad: Dear Lord, again... Looking into it!

misicnenad: @ap-justin can't repro, can you post exact repro steps?

Also, did you pull the latest branch changes before testing? 
And did you run `yarn install` when switching to this branch? Please try `rm -rf node_modules && yarn install` and then test things out, maybe there occurred a corruption in your *node_modules*

ap-justin: weird, I always delete local branch and pull latest remote. works now👍

src/services/apes/apes.ts
```diff
@@ -1,15 +1,20 @@
 import { createApi, fetchBaseQuery } from "@reduxjs/toolkit/query/react";
-import { EndowmentBalances, Token } from "types/aws";
+import { PaymentIntent } from "@stripe/stripe-js";
+import { EndowmentBalances, FiatCurrencyData, KYCData, Token } from "types/aws";
 import { ChainID } from "types/chain";
-import { appRoutes } from "constants/routes";
+import { version as v } from "services/helpers";
```
ap-justin, line 5: nit: as apes is also inside `/services`
```suggestion
import { version as v } from "../helpers";
```

src/components/BankDetails/BankDetails.tsx
```diff
@@ -14,39 +14,64 @@ const DEFAULT_EXPECTED_MONTHLY_DONATIONS_AMOUNT = "1000";
...
+    try {
+      setSubmitting(true);
+      await onSubmit(...params);
+    } catch (error) {
+      // All errors should be handled in `onSubmit`.
+      // This try/catch is just to ensure that `isSubmitting`
+      // is set to false at the end of the operation.
+      console.error(error);
+    } finally {
+      setSubmitting(false);
```
ap-justin, line 43: comment: I didn't stress too much on disabling currency and amount as user is way below the dynamic form when they hit submit - and probably won't gain any benefit changing mid-submit if they can

ap-justin (dismissed):
tested donation and works 🚀

src/components/BankDetails/BankDetails.tsx
misicnenad: Understandable. I think of this more as a UX aspect, to indicate to the user they shouldn't be touching anything while the form is submitting or loading.

src/pages/Admin/Charity/Banking/Banking.tsx
```diff
@@ -28,9 +28,9 @@ export default function Banking() {
 
       const { id, details, currency } = recipient;
       //creating account return V1Recipient and doesn't have longAccount summary field
-      const bankSummary = `${currency.toUpperCase()} account ending in ${details.accountNumber?.slice(
-        -4 || "0000"
-      )} `;
+      const bankSummary = `${currency.toUpperCase()} account ending in ${
+        details.accountNumber?.slice(-4) || "0000"
```
misicnenad, line 32: @ap-justin was this here by mistake?
Seemed to be that the intention was that `"0000"` was to be used in case the previous value is falsy.

ap-justin: right should be outside

SovereignAndrey (changes requested):
@misicnenad The payment flow worked overall very well for me. Nice work on this. :) Just a couple smaller things to change:
- [ ] First screen: the button should read "Continue" or "Next" because the user is not paying...yet. That happens in later submission step after entering card details

- [ ] First screen: Switch order of currency input and amount. Currency choice drives the UI changes (help note below on minimums) for amount box.
- [ ] We need to require entering an email somewhere (first screen or payment screen whichever is easier). We need to have a way for users to get KYC receipt later if they want one and didn't complete it at time of payment. If the user is logged in, prefill with that email. That email field needs to be posted to AWS along with other donation info.

misicnenad:
@SovereignAndrey 
> If the user is logged in, prefill with that email(and disable field).

What if the user wants to use a different email address for KYC?

SovereignAndrey:
> 1. Is disabling necessary? What if the user wants to use a different email address for KYC?

Yes, I think so. This field is so that non-logged in users can link the donation to a user `email` field. Different from the KYC email field. If a logged in user wants to use a different KYC email they can request KYC receipt at time of donation OR later. When they log in the `email` field will allow them to pull that donation up in My Donations area and request receipt to wherever they want to send it.    

> 2. There's currently no way for a non-signed in user to fill out KYC after donating (they don't have access to "My Donations" page), do we need to ask them for their email at all?

Yes. By asking for an email we can allow them the chance to pull that record up when they do make a User account. 

> 3. For signed in users:
>    
>    1. After signing in and donating through Stripe, I don't see my donation in "My Donations", is this WIP or am I missing something?
>    2. We don't even have to ask signed in users to input their email, we can just include it by default in the donate request. Even this might be redundant, depending on the logic we're gonna use to display data in "My Donations" (if we know which donation is from who, then we know all their data, including email)

3.1 - You don't see it because the email of your user is not being sent to AWS in the POST payload for `email` correctly? If sent that field will be what pulls records into a user's My Donations page table. 
3.2 - That's a good idea. Omit the need to see the email field if user is logged in since this will get set to their user email on record.

misicnenad:
Converting to draft: need to make `KYC.email` a required field on both webapp and AWS.

misicnenad:
Created a new lambda update for the relevant change request
https://github.com/AngelProtocolFinance/devops/pull/467

misicnenad:
Converting to draft: need to enable storing email address when switching between Form, KYC and Checkout steps

misicnenad:
Updated the code, ready for review:
- added email field
- differentiate between KYC email and user's email (if authenticated)
- store and preload data when switching between steps

SovereignAndrey (approved):

PR #2629, opened 2024-01-04
BG-1005: Fix /auth-redirector URL appearing after redirecting

misicnenad:
Had an [interesting finding](https://linear.app/better-giving/issue/BG-1005/after-signing-in-and-being-redirected-to-marketplace-the-url-should-be#comment-aeeaecf0), quoting below:
> Interesting finding: when I throttle the network, the issue does not occur - maybe the issue lies with aws-amplify redirecting twice to the same /auth-redirector page :thinking: 
> 
> See repro:
https://www.loom.com/share/10480dbb1f584085a3254dacdb504d38?sid=fa93ce3b-fb10-48f0-9fce-2f1ccb19cd2a

Interested in your thoughts on this @SovereignAndrey @ap-justin @jp-mariano , maybe you know something that could shed light on this issue?

In the meantime, added a 300 ms delay before redirecting in `OAUTHRedirector` which completely avoids the bug. This is a band-aid (and hopefully temporary) solution.

- make sure you're signed out
- click on "login" button (@SovereignAndrey we should align terms we use, either it's "login & logout" or it's "sign in & sign out", but currently we use a mix of both)
- sign in using your Google account
- **verify the bug no longer appears**

### DIFF
--- a/src/pages/OAuthRedirector.tsx
+++ b/src/pages/OAuthRedirector.tsx
@@ -1,6 +1,25 @@
-import { Navigate } from "react-router-dom";
+import { useEffect } from "react";
+import { useNavigate } from "react-router-dom";
 import { OAUTH_PATH_STORAGE_KEY } from "constants/auth";
 
+// For some reason aws-amplify redirects to this page twice; the slowness of the 2nd
+// redirect causes the final URL to be of this page even though the rendered page is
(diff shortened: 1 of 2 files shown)


misicnenad:
*yarn.lock* got updated after I ran `yarn install`

ap-justin (approved):

PR #2652, opened 2024-01-17
Custom signup confirm text

ap-justin:
since this text is generated internal by `authenticator-ui-react` - a way to override this is to find the generated text and override it. This is hacky and not so maintainable (as lib authors might change the className of generated text). A more 
sustainable but longer approach is to entirely switch to custom UI. 

Screenshot 2024-01-17 at 10 50 11

-

## UI changes for review


SovereignAndrey (approved):
Hacky indeed, but it works. :+1:

PR #2653, opened 2024-01-19
FIX: client side vaildation not running

ap-justin:
turned out setting `register.options.disabled` also turns off validation

* instead of disabled each field, set disable in `<fieldset/>`, alternative would be `readOnly` 

### Others
* create form with fieldset to avoid nesting which affects diff view
* move individual `shouldUnregister` to central `useForm`

-

## UI changes for review


misicnenad (approved):
Nice and simple :+1:
I knew [I had tunnel vision](https://linear.app/better-giving/issue/BG-1085/bankdetails-validation-does-not-trigger-prior-to-submit#comment-ebc08598) with regards to this.

PR #2660, opened 2024-01-23
Update PUBLIC_ENVIRONMENT env. var to 'dev | production'

misicnenad:
All APES endpoints seem to have changed to contain as a prefix either "dev" or "production".


SovereignAndrey (approved):
@misicnenad @jp-mariano We'll need to consider other `staging | production` uses in the lambdas that will be impacted by the change to `dev`. Opened a ticket for this in Linear.

src/constants/urls.ts
```diff
@@ -1,8 +1,8 @@
-import { BASE_URL } from "./env";
+import { BASE_URL, ENVIRONMENT } from "./env";
 
 export const APIs = {
   aws: "https://kpnxz5rzo2.execute-api.us-east-1.amazonaws.com",
```
misicnenad, line 4: @jp-mariano should these endpoints have the `dev/production` prefix as well?

PR #2661, opened 2024-01-23
Towards BG-1056, Closes BG-1094: Donation state re-structure for multiple methods

ap-justin:
* kindly see code comments for code changes explanations
* add additional tab buttons layout for wide screen
image
* update tab options for mobile 
image

* modify donation state to accomodate layout 
from:
```yaml
1 - FORM
  -Crypto
  -Stripe -> KYC -> Checkout
  -Paypal -> KYC -> Checkout
  ...others
  
2 - CRYPTO_KYC
 
3 - CRYPTO_SUBMIT
 
4 - CRYPTO_RESULT
```
to: 
```yaml
1 - FORM
  -Crypto
  -Stripe
  -Paypal
  ...others
  
2 - KYC

3 - SUBMIT
 -Crypto (sendTx)
 -Stripe Checkout
 -Paypal Checkout
 
4 - CRYPTO_RESULT

```
this would prove useful with the plan to have a standalone `splits` screen and summary page for `stocks`. Also `KYC` need not be aware of payment method used. 

-

## UI changes for review


src/components/donation/Steps/CurrentStep.tsx
```diff
@@ -12,34 +12,34 @@ export default function CurrentStep({ config }: Props) {
   const state = useGetter((state) => state.donation);
   const dispatch = useSetter();
 
+  if (state.step === "init") return <></>; // <Steps /> sets to step 1 onMount
```
ap-justin, line 15: reduce nesting

src/components/donation/Steps/Submit/StripeCheckout/Checkout.tsx
```diff
@@ -61,16 +61,17 @@ export default function Checkout({ onBack }: { onBack: () => void }) {
           handleError(error, GENERIC_ERROR_MESSAGE);
         }}
       />
+
       {isLoading ? (
         <button
-          className="btn-outline-filled btn-donate w-1/2"
+          className="btn-outline-filled btn-donate w-1/2 mt-auto"
```
ap-justin, line 67: min-height + auto to prevent layout shift
image

src/components/donation/Steps/Submit/Submit.tsx
```diff
@@ -1,128 +1,13 @@
...
 import { SubmitStep } from "slices/donation";
-import { chains } from "constants/chains";
-import Breakdown from "./Breakdown";
-import Container from "./Container";
-import WalletSelection from "./WalletSelection";
+import Crypto from "./Crypto";
+import StripeCheckout from "./StripeCheckout";
 
 export default function Submit(props: SubmitStep) {
-  const wallet = useWalletContext();
```
ap-justin, line 14: this `<Submit/>` is now moved to `Submit/Crypto`

src/components/CurrencySelector/CurrencyOptions.tsx
```diff
@@ -1,39 +1,73 @@
...
-import { Currency } from "./types";
-import { isEmpty } from "helpers";
+import { Currency } from "types/components";
+import { QueryState } from "types/third-party/redux";
+import QueryLoader from "components/QueryLoader";
 
 type Props = {
   classes?: string;
-  currencies: Currency[];
+  currencies: Currency[] | QueryState<Currency[]>;
```
ap-justin, line 8: move loading of currency options inside `<CurrencySelector/>` so whole form need not wait for currencies to load - most users could certainly start with USD

src/components/donation/Steps/DonateMethods/DonateMethods.tsx
```diff
@@ -12,41 +12,64 @@ type Props = {
...
 
 export default function DonateMethods({ donaterConfig, state }: Props) {
   return (
     <Tab.Group
+      manual
       as="div"
-      className="grid content-start mt-2"
-      defaultIndex={state.details?.method === "crypto" ? 1 : 0}
+      className="grid @md:grid-cols-[auto_1fr]"
+      defaultIndex={tabIdx(state.details?.method)}
```
ap-justin, line 41: when user goes on with `stripe` and goes back, stripe should still be the selected method

src/components/donation/Steps/DonateMethods/DonateMethods.tsx
```diff
@@ -12,41 +12,64 @@ type Props = {
...
-            state={state}
           />
         </Tab.Panel>
         <Tab.Panel>
-          <Donater {...state} config={donaterConfig} />
+          <Crypto
+            recipient={state.recipient}
+            step={state.step}
+            details={
+              state.details?.method === "crypto" ? state.details : undefined
```
ap-justin, line 69: if user initially goes on with payment option A, and goes back and goes on to another option B, option A data (`amount`,`currency`...) is lost

src/components/donation/Steps/Submit/Crypto/Crypto.tsx
```diff
@@ -0,0 +1,128 @@
+import { useState } from "react";
```
ap-justin, line 1: this file is just moved and renamed

src/components/donation/Steps/Submit/PaypalCheckout/Checkout.tsx
```diff
@@ -77,6 +69,14 @@ export default function Checkout({ orderId, onBack }: Props) {
               return orderId;
             }}
           />
+          <button
```
ap-justin, line 72: this now looks
image

src/services/apes/apes.ts
```diff
@@ -46,31 +47,27 @@ export const apes = createApi({
...
         headers: { authorization: TEMP_JWT },
         body: JSON.stringify(params),
       }),
+      transformResponse: (res: { orderId: string }) => res.orderId,
     }),
-    createStripePaymentIntent: builder.mutation<
-      { clientSecret: string },
-      StripePaymentIntentParams
-    >({
+    stripePaymentIntent: builder.query<string, StripePaymentIntentParams>({
```
ap-justin, line 64: `POST` inside of "query" looks weird at first but this setup works and has these advantages:
1. payment intent can be created when page loads
2. as this is a query, intent can be cached for up to [60s](https://redux-toolkit.js.org/rtk-query/usage/cache-behavior#default-cache-behavior:~:text=(default%2060%20seconds)). And would only revalidate if user changed `StripePaymentIntentParams` - amount, currency...etc.

src/components/BankDetails/BankDetails.tsx
```diff
@@ -1,9 +1,10 @@
 import { useState } from "react";
 import { IFormButtons, OnSubmit } from "./types";
+import { Currency } from "types/components";
```
ap-justin, line 3: @SovereignAndrey, opening this PR as preparatory for further redesign works. I unfortunately have made design UI changes (methods sidebar, ) before I discovered the need for restructuring. Let me know if I should exclude those to this PR.

misicnenad (commented):
Not commenting on the changes themselves as the code seems fine on first look, but the amount of not-necessarily-related changes in a single PR is very hard to review with any quality, it's difficult to keep track of everything.
Big PRs with many unrelated updates incentivize skimming and approving just to save time and effort, instead of actually reviewing the PR.  

@ap-justin would you at least consider breaking it up into smaller, more focused PRs? Would help make reviews easier.

@SovereignAndrey we could create a larger Linear ticket that keeps track of all the restructuring changes.

src/components/BankDetails/BankDetails.tsx
ap-justin: > Not commenting on the changes themselves as the code seems fine on first look, but the amount of not-necessarily-related changes in a single PR is very hard to review with any quality, it's difficult to keep track of everything.
Big PRs with many unrelated updates incentivize skimming and approving just to save time and effort, instead of actually reviewing the PR.

> @ap-justin would you at least consider breaking it up into smaller, more focused PRs? Would help make reviews easier and actually useful.

@misicnenad, changes generally falls into these sequential parts
1. donor tabs + common panels container - Towards `BG-1056`
2. I noticed that `stripe` and `paypal` has KYC form inside donate methods and to bring them out, some changes to `DonationState` should be made (also means standalone state inside stripe/paypal is now removed ) - Closes `BG-1094`
3. As paypal/stripe checkout is now under Submit step, make sure they are correctly rendered there (standalone) - so some spacing changes.

src/components/donation/Steps/Submit/Crypto/Crypto.tsx
```diff
@@ -0,0 +1,128 @@
...
+import { ConnectedWallet, isCosmos, isEVM, isTerra } from "types/wallet";
+import { isDisconnected, useWalletContext } from "contexts/WalletContext";
+import { Info, LoadingStatus } from "components/Status";
+import { CryptoSubmitStep } from "slices/donation";
+import { chains } from "constants/chains";
+import Breakdown from "./Breakdown";
+import Container from "./Container";
+import WalletSelection from "./WalletSelection";
+
+export default function Crypto(props: CryptoSubmitStep) {
```
misicnenad, line 13: Should align the button that navigates back to the profile page:
Crypto donation - "Cancel + Continue" buttons:
image

PayPay & Stripe - "Back + Continue" buttons:
image

DAF - no buttons to navigate back:
image

Stocks - no buttons to navigate back:
image

-------

1. Either all buttons are named "Back" or all are named "Cancel"
2. Either all donation methods should have the "Back" button or none should (user can still go back using by clicking the built-in back button in their browser/mobile phone OR by using breadcrumbs)
image

src/components/donation/Steps/CurrentStep.tsx
misicnenad, line 15: @SovereignAndrey unrelated to the current PR, but I really think when a user gets to the "Thanks for donating" page, there should be a button to go back to the page of the endowment to which the user just donated instead of "going back to the platform" -> the user is *already* on the platform, if they wished to go back they would probably want to go to the endowment's profile and furthermore, if they wished to go to the Marketplace, they can just click the appropriate button in the navbar (so no need for an additional one on the page).
There were many times when I needed such a button while working on donation flows, but instead I had to manually input the URL in the navbar (much faster than going through Marketplace). 
image

misicnenad (changes requested):
Seems good, I had just one change request

src/components/donation/Steps/Submit/Crypto/Crypto.tsx
ap-justin: @misicnenad, I believe "cancel" was the one used on prev design as this is the first part of the multistep, so no prev step to go "back" to - switching all to "cancel" for now for consistency. Ultimately, all would be transferred a top level "cancel" (fixes DAF and Stocks concern) if its the intention - documented [here](https://linear.app/better-giving/issue/BG-1104/move-cancel-button-at-the-top-of-the-page)

src/components/donation/Steps/CurrentStep.tsx
SovereignAndrey: @ap-justin This is looking good overall! Lot of little things here and there that can be improved to get further inline style and UI wise, but this is a great base to work off of for future (smaller) PRs and shouldn't be a blocker to moving forward. 

Re: @misicnenad's comments about syncing up the payment methods, I agree it jumped out at me too when reviewing the UI. To that end, I think we can incorporate a "Back" link and "Cancel" link at the top of the section/page, much like every.org does. This would mean there's only one button at the bottom that moves the user forward, "Continue"/"Pay"/etc. It probably helps with donation conversions to keep users from dropping off as readily when the back and cancel are not _quite_ easy to get to, but there nonetheless.

Here's the screenshot from their site (having clicked into the Venmo section's flow by at least one step):
image

One important distinction that we'll need to consider is the "Cancel" link placement outside the middle area. Since the widget will use this UI too, we'll need to make sure we consider that the widget view will NOT have the flanking side bars (unlike the official donation page...eventually) and we'll need to move that over closer to the middle. Our view will thus be like their mobile view I think regardless of desktop/mobile: 
image

If we move it down a bit that might be the ticket:
image

SovereignAndrey: >  there should be a button to go back to the page of the endowment to which the user just donated instead of "going back to the platform" -> the user is already on the platform, if they wished to go back they would probably want to go to the endowment's profile and furthermore, if they wished to go to the Marketplace, they can just click the appropriate button in the navbar (so no need for an additional one on the page).

@misicnenad Fair point! Though I'd argue that if a user has just finished donating to a nonprofit they are less likely to go back to their profile page and instead, as I think the original intent was for this link, can be encouraged to go donate more (ie. back to the Marketplace to find new nonprofits).

ap-justin: @SovereignAndrey 
> One important distinction that we'll need to consider is the "Cancel" link placement outside the middle area. Since the widget will use this UI too, we'll need to make sure we consider that the widget view will NOT have the flanking side bars (unlike the official donation page...eventually) and we'll need to move that over closer to the middle. Our view will thus be like their mobile view I think regardless of desktop/mobile:

got this ser, may not be a problem since "cancel" is just a link to profile and would now be owned by the container of multistep form. 

image

For the `/donate-page` - just how it is handled by every org (at the top). For the widget, may not be needed as currently, we don't even have a "cancel" button. 

The title could even be a link if there's a need for widget donor to go to the original profile page 
image

SovereignAndrey: @ap-justin Some other smaller changes needed are (example in screenshot):
- [ ] the use of icons next to payment options
- [ ] collapse/show more options when in space restricted mobile view (no more than 4 buttons / two rows) 

image

SovereignAndrey: >For the /donate-page - just how it is handled by every org (at the top). For the widget, may not be needed as currently, we don't even have a "cancel" button.
> [sic]
> For the /donate-page - just how it is handled by every org (at the top). For the widget, may not be needed as currently, we don't even have a "cancel" button.

@ap-justin All good regarding the above. :+1: 
No need to link to original profile from widget for now, though that could be something to consider doing from the widget "Thank You" page to bring users to the main app.

ap-justin: @SovereignAndrey , applied additional requirements (icons, and more options)
image

image

src/components/donation/Steps/Submit/Crypto/Crypto.tsx
SovereignAndrey: 💯

SovereignAndrey (approved):

PR #2665, opened 2024-01-26
Biome format & checks first run

SovereignAndrey:
First run of Biome based formatting changes & all fixable lint fixes applied

### TO DO:
- [ ] fix last of the lint errors needing manual review

-

## UI changes for review
None


src/App/Footer/Footer.tsx
```diff
@@ -15,7 +15,7 @@ function Footer({ linkGroups, socials }: Props) {
         const intercom = document.querySelector(".intercom-launcher");
         intercom?.classList.toggle("hidden", e.isIntersecting);
       },
-      { threshold: [0.5] }
+      { threshold: [0.5] },
```
ap-justin, line 18: @SovereignAndrey, some formatting of biome could be ignored just like this one: can be turned off via `javascript.formatter.trailingComma = es5`

src/App/App.tsx
```diff
@@ -48,12 +48,12 @@ export default function App() {
...
           />
           <Route
-            path={appRoutes.banking_applications + "/:id"}
+            path={`${appRoutes.banking_applications}/:id`}
             element={<BankingApplication />}
           />
           <Route path={appRoutes.applications} element={<Applications />} />
           <Route
-            path={appRoutes.applications + "/:id"}
+            path={`${appRoutes.applications}/:id`}
```
ap-justin, line 56: some lint rules are just style related that we  can safely stick to our existing code. I tried biome last night with less invasive rules #2666 , that we can make stricter incrementally prioritizing performance, then correctness, then lastly style related

another point to consider is switching to biome formatter is best with the biome vscode extension as replacement for `prettier` - so if we're going to adopt this on web-app, we should also adopt on devops, devops-ses at once so we don't need to switch from prettier - biome from time to time

PR #2689, opened 2024-02-05
Update Split.tsx to handle 'stocks' separately when reading amount+symbol

misicnenad:
Fixing build errors in *Split.tsx*:
image

### DIFF
--- a/src/components/donation/Steps/Splits/Splits.tsx
+++ b/src/components/donation/Steps/Splits/Splits.tsx
@@ -24,6 +24,10 @@ export default function Split({
         const { amount, symbol } = details.token;
         return [amount, symbol];
       }
+      case "stocks": {
+        const { numShares, symbol } = details;
+        return [numShares, symbol];
+      }
       default:
         const { amount, currency } = details;
         return [amount, currency.code.toUpperCase()];


src/components/donation/Steps/Splits/Splits.tsx
```diff
@@ -24,6 +24,10 @@ export default function Split({
         const { amount, symbol } = details.token;
         return [amount, symbol];
       }
+      case "stocks": {
+        const { numShares, symbol } = details;
+        return [numShares, symbol];
```
misicnenad, line 29: @SovereignAndrey wouldn't it make sense to display crypto/stock symbols and currency code in upper case by default [in the code below](https://github.com/AngelProtocolFinance/angelprotocol-web-app/pull/2689/files#diff-9687f930a1e28a056462d274d81e5b3d8aca1c4a5dcc5593f45eda1a1b865c9aR86-R97):
image

ap-justin (approved):

src/components/donation/Steps/Splits/Splits.tsx
SovereignAndrey: > @SovereignAndrey wouldn't it make sense to display crypto/stock symbols and currency code in upper case by default
It most definitely would be good to treat those as needing to be uppercase whenever displayed to users. :100:

misicnenad: Created a ticket https://linear.app/better-giving/issue/BG-1134/liquidlocked-split-component-display-cryptostock-symbols-and-currency

PR #2709, opened 2024-02-07
BG-1107, BG-1126: Fiat checkout summary

ap-justin:
* include summary in stripe checkout
image

* include summary in paypal checkout
image

* include summary in DAF checkout 
image

-

## UI changes for review

### DIFF
--- a/src/components/BankDetails/BankDetails.tsx
+++ b/src/components/BankDetails/BankDetails.tsx
@@ -1,7 +1,7 @@
 import Divider from "components/Divider";
 import useDebouncer from "hooks/useDebouncer";
 import { useState } from "react";
-import { useCurrencisQuery } from "services/aws/wise";
+import { useWiseCurrenciesQuery } from "services/aws/wise";
 import { Currency } from "types/components";
 import CurrencySelector from "../CurrencySelector";
 import ExpectedFunds from "./ExpectedFunds";
@@ -24,6 +24,7 @@ export default function BankDetails({ FormButtons, onSubmit }: Props) {
   const [currency, setCurrency] = useState<Currency>({
     code: "USD",
     name: "United States Dollar",
+    rate: 1,
   });
   const [amount, setAmount] = useState(
     DEFAULT_EXPECTED_MONTHLY_DONATIONS_AMOUNT
@@ -45,7 +46,7 @@ export default function BankDetails({ FormButtons, onSubmit }: Props) {
     }
   };
 
-  const currencies = useCurrencisQuery({});
+  const currencies = useWiseCurrenciesQuery({});
 
   return (
     <div className="grid gap-6">
--- a/src/components/donation/Steps/Submit/ChariotConnectCheckout.tsx
+++ b/src/components/donation/Steps/Submit/ChariotConnectCheckout.tsx
@@ -1,3 +1,4 @@
+import ContentLoader from "components/ContentLoader";
 import { GENERIC_ERROR_MESSAGE } from "constants/common";
 import { CHARIOT_CONNECT_ID } from "constants/env";
 import { appRoutes, donateWidgetRoutes } from "constants/routes";
@@ -9,7 +10,9 @@ import { ChariotCheckoutStep, setStep } from "slices/donation";
 import { useSetter } from "store/accessors";
 import BackBtn from "../BackBtn";
 import Err from "./Err";
-import Loader from "./Loader";
+import Currency from "./common/Currrency";
+import Heading from "./common/Heading";
+import SplitSummary from "./common/SplitSummary";
 
 // Followed Stripe's custom flow docs
 // https://stripe.com/docs/payments/quickstart
@@ -33,8 +36,13 @@ export default function ChariotConnectCheckout(props: ChariotCheckoutStep) {
     };
   };
 
+  const currency = details.currency;
+  const total = +details.amount;
+  const liq = total * (liquidSplitPct / 100);
+  const locked = total - liq;
+
   return (
-    <div className="grid content-start gap-8 p-4 @md:p-8">
+    <div className="flex flex-col content-start p-4 @md:p-8 group">
       <BackBtn
         type="button"
         onClick={() => {
@@ -45,8 +53,20 @@ export default function ChariotConnectCheckout(props: ChariotCheckoutStep) {
         }}
       />
 
+      <Heading classes="my-4" />
+
+      <SplitSummary
+        classes="mb-auto"
+        total={<Currency {...currency} amount={total} classes="text-gray-d2" />}
+        liquid={<Currency {...currency} amount={liq} classes="text-sm" />}
+        locked={<Currency {...currency} amount={locked} classes="text-sm" />}
+      />
+
+      {/** <CharioConnect/> is not yet rendered */}
+      <ContentLoader className="rounded h-14 w-full group-has-[chariot-connect]:hidden" />
+
       {isLoading ? (
-        <Loader msg="Processing donation..." />
+        <ContentLoader className="rounded h-12 w-full" />
       ) : isError ? (
         <Err error={error} />
       ) : (
--- a/src/components/donation/Steps/Submit/ChariotConnectCheckout.tsx
+++ b/src/components/donation/Steps/Submit/ChariotConnectCheckout.tsx
@@ -43,18 +43,8 @@ export default function ChariotConnectCheckout(props: ChariotCheckoutStep) {
 
   return (
     <div className="flex flex-col content-start p-4 @md:p-8 group">
-      <BackBtn
-        type="button"
-        onClick={() => {
-          const action = details.userOptForKYC
-            ? setStep("kyc-form")
-            : setStep("donate-form");
-          dispatch(action);
-        }}
-      />
-
+      <BackBtn type="button" onClick={() => dispatch(setStep("splits"))} />
       <Heading classes="my-4" />
-
       <SplitSummary
         classes="mb-auto"
         total={<Currency {...currency} amount={total} classes="text-gray-d2" />}
--- a/src/components/donation/Steps/Submit/StripeCheckout/StripeCheckout.tsx
+++ b/src/components/donation/Steps/Submit/StripeCheckout/StripeCheckout.tsx
@@ -7,6 +7,9 @@ import { useSetter } from "store/accessors";
 import BackBtn from "../../BackBtn";
 import Err from "../Err";
 import Loader from "../Loader";
+import Currency from "../common/Currrency";
+import Heading from "../common/Heading";
+import SplitSummary from "../common/SplitSummary";
 import Checkout from "./Checkout";
 
 const stripePromise = loadStripe(PUBLIC_STRIPE_KEY);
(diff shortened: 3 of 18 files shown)


src/components/donation/Steps/Submit/common/SplitSummary.tsx
```diff
@@ -0,0 +1,29 @@
+import { ReactNode } from "react";
+
+type Props = {
+  total: ReactNode;
+  locked: ReactNode;
+  liquid: ReactNode;
+  classes?: string;
+};
+
+export default function SplitSummary(props: Props) {
```
misicnenad, line 10: Suggestion: related to the Split component, it would make sense to turn the cursor into pointer when hovering over the slider:
image

src/components/donation/Steps/Submit/ChariotConnectCheckout.tsx
```diff
@@ -33,8 +36,13 @@ export default function ChariotConnectCheckout(props: ChariotCheckoutStep) {
...
 
+  const currency = details.currency;
+  const total = +details.amount;
+  const liq = total * (liquidSplitPct / 100);
+  const locked = total - liq;
+
   return (
-    <div className="grid content-start gap-8 p-4 @md:p-8">
+    <div className="flex flex-col content-start p-4 @md:p-8 group">
       <BackBtn
```
misicnenad, line 46: Bug: when the user clicks "Back", they should be redirected to the Split component (as is the case with all other donation methods); instead, the user is navigated all the way back to currency+donation amount screen.

src/components/donation/Steps/Submit/StripeCheckout/StripeCheckout.tsx
```diff
@@ -41,9 +44,22 @@ export default function StripeCheckout(props: StripeCheckoutStep) {
 
   const dispatch = useSetter();
 
+  const currency = details.currency;
+  const total = +details.amount;
+  const liq = total * (liquidSplitPct / 100);
```
misicnenad, line 49: Small inconsistency: amounts shown in the "Split donation" step can sometimes differ from amounts shown in the Checkout; this is probably a rounding issue.

Repro steps:
1. Select "Card" donation option
2. Select USD currency
3. Input `4` as amount
4. Click "Continue"
5. Set split to 66/34% (locked/liquid) -> note the amounts:
- Compounded Forever: 2.6400
- Instantly Available: 1.3599
image
6. Click "Continue"
7. Note the amounts differ by $0.01:
- Compounded Forever: 2.63
- Instantly Available: 1.36
image

@SovereignAndrey if we're fine with these small rounding inconsistencies, then you can ignore the comment.

src/services/aws/wise.ts
```diff
@@ -29,8 +30,10 @@ export const wise = aws.injectEndpoints({
     recipient: builder.query<V2RecipientAccount, string>({
       query: (id: string) => `${baseURL}/v2/accounts/${id}`,
     }),
-    currencis: builder.query<WiseCurrency[], unknown>({
+    wiseCurrencies: builder.query<Currency[], unknown>({
       query: () => `${baseURL}/v1/currencies`,
+      transformResponse: (res: WiseCurrency[]) =>
+        res.map((r) => ({ rate: null, code: r.code, name: r.name })),
```
misicnenad, line 36:
```suggestion
        res.map((r) => ({ code: r.code, name: r.name })),
```
We can update the type of `Currency.rate` to be `rate?: number`, so there would be no need to set it here.

src/types/components.ts
```diff
@@ -18,4 +18,9 @@ export type Country = {
 };
 
 //currency selector
-export type Currency = { code: string; name?: string; min?: number };
+export type Currency = {
+  code: string;
+  name?: string;
+  min?: number;
+  rate: number | null;
```
misicnenad, line 25:
```suggestion
  rate?: number;
```

misicnenad (changes requested):
Tested this out, everything mostly seems to work fine.
Reported some minor (potential) issues and suggestions.

src/components/donation/Steps/Submit/StripeCheckout/StripeCheckout.tsx
SovereignAndrey: @misicnenad @ap-justin We do want to ensure that previous values match the donation summary, so as not a cause donors alarm. This is likely a rounding/display issue that would be solved by making sure all displayed values are rounded to:
- 2 decimal places for fiat
- 4 decimal places for crypto

src/types/components.ts
ap-justin: had I added `rate` as optional property, I would have to manually search throughout where `type Currency` is used and added `rate` accordingly which is easy to miss. 

adding `rate` as explicit `number | null` - suddenly typechecker is so helpful
image

there's no way to miss where `rate` is needed - transform values, default values etc.

src/services/aws/wise.ts
ap-justin: see [related comment](https://github.com/AngelProtocolFinance/angelprotocol-web-app/pull/2709#discussion_r1482357610) - having `null` here isn't critical (a bit helpful, tells otherwise that wise currencies no need to utilize `rate` property ) that we need to `rate` optional across.

here's a [youtube video](https://www.youtube.com/watch?v=X_tvQoW4ioI) that expounds this nuance

src/components/donation/Steps/Submit/ChariotConnectCheckout.tsx
ap-justin: fixed: https://github.com/AngelProtocolFinance/angelprotocol-web-app/pull/2709/commits/c87b9b48868d61e31e5d00199520180f907ecff0

src/components/donation/Steps/Submit/StripeCheckout/StripeCheckout.tsx
ap-justin: found this inconsistency was caused by how `liquid, locked` is determined in splits and checkout  
In splits step:
```js
//get locked first
const locked = total * ( pctLocked  / 100 )
//derive liq
const liq = total - locked
```
while in checkout 
```js
//get liq first
const liq = total * ( pctLiq  / 100 )
//derive locked
const locked = total - liq
```
Nonetheless, still applied 4 decimals for crypto and 2 decimals for fiat @SovereignAndrey

src/services/aws/wise.ts
misicnenad: I watched the video, the creator recommends the use of union with undefined (or null in this case) arguments only if there's an explicit use case for it; he goes on to give an example that's similar to our `ErrorContext.handleError`:
```ts
function logError(error: Error | undefined, message: string) { /** ... */ }
```
In all other cases even he recommends the use of optional arguments (`?`)

src/types/components.ts
misicnenad: That was a useful and once you were done adding the missing property, you could've safely made `rate` optional again.
Leaving it as a union type only forces verbose code.

ap-justin: @misicnenad, I used the video because I thought you need more context on the difference between `rate?:` and `rate: x | undefined` regarding the use of it in this PR. 

> Leaving it as a union type only forces verbose code.
Again, on the context of what's being done here - I would argue it's not even verbose - only the wise api needs an explicit `null`. most of the consumers need to add it anyway. 

On the otherhand, I would have considered `rate?:` if most of the consumers doesn't need to set it.

src/components/BankDetails/BankDetails.tsx
```diff
@@ -24,6 +24,7 @@ export default function BankDetails({ FormButtons, onSubmit }: Props) {
   const [currency, setCurrency] = useState<Currency>({
     code: "USD",
     name: "United States Dollar",
+    rate: 1,
```
misicnenad, line 27:
```suggestion
    rate: null,
```
nit: you don't need a rate in this component.

misicnenad (changes requested):

src/components/BankDetails/BankDetails.tsx
ap-justin: value `1` won't hurt either to conform the `type Currency` or to say that rate of `$1` is obviously 1

src/components/donation/Steps/Submit/common/SplitSummary.tsx
misicnenad: @ap-justin pinging you in case you missed this

src/types/components.ts
ap-justin: @misicnenad 
> I think you too understand that this could now safely be made an optional type and you also know that it wouldn't take more than a minute of your time to make the change

So we just temp opt for a type variation and once satisfied we add another effort to revert to a more lax variant? Is this how you propose to maintain our interfaces? 

> just as I knew in similar situations when you asked me to make trivial changes in my own PRs,

The difference between you and me is that I don't persist on trivial things. When I mark my comment as `nit` I don't go around forcing other developer to make that change to suit my style as it is more of a suggestion. I don't understand why you are going so much with this trivial `number | null` choice even I explained to you the reason I did in the first place.

I assume it came from when you unneccesarily changed `DonaterConfig | null` to `DonaterConfig?` which if you haven't done in the first place, would have not caught my attention. Futhermore, I didn't relentlessly force you to change it - you just reverted it on your own with some suble insult
> I guess this is just a matter of preference, so will revert the change so as not to hold up the whole PR for a trifle.

If you are making changes for the sake of personal preference - probably its better not to change it in the first place?


> that's why I no longer rename your classes: string props into className, even though I prefer that name; 

Again, what benefit would you get from doing so? - so you're doing yourself a good favor on leaving it be.

> that's why I'd ask you to reconsider your position on this topic and make this an optional type, at least as a sign of goodwill.

If what you are suggesting is objective and has effect on the workings of the web-app, I would have accepted this in no time - but suggesting something subjective (because you can't stand seeing a single `null` ? ) How do you expect me to follow this? 

> Otherwise we're just destined to continue the cycle. The result? PRs will be halted for silly reasons, resentment will build, time (and nerves) will be wasted, code quality will drop. It doesn't have to be this way.

Take a look back, when you saw `number | null` you could have just let it go as there's really no substance into it. But you did push on it - I provided with you an objective reason for such decision and you go about finding more holes into it further pushing this futile argument to consume more of our time.

misicnenad: @ap-justin 

> I assume it came from when you unnecessarily changed DonaterConfig | null to DonaterConfig? which if you haven't done in the first place, would have not caught my attention. Futhermore, I didn't relentlessly force you to change it - you just reverted it on your own with some suble insult
> 
> > I guess this is just a matter of preference, so will revert the change so as not to hold up the whole PR for a trifle.

This wasn't meant as any type of insult and I'm sorry that you took it that way. I reverted the change, as you said - "on my own", because I didn't want to create *this* type of discussion for nothing/for a trifle.

> I provided with you an objective reason for such decision and you go about finding more holes into it further pushing this futile argument to consume more of our time.

You may be right about me pushing. Occasionally I find your reasons misguided or unconvincing, and especially so after I feel (rightly or wrongly) like you're ignoring the "holes" that I find in them - this causes me to get frustrated, so I push you to address the "holes" I had found. 

I'm sorry our relations are so low that we can't even have an honest discussion about how low they are; and I'm tired of pointing fingers and remembering every slight. I'm not excusing myself though, I contributed my fair share to the current state of things. Maybe the cause of us consistently misunderstanding each other is the baggage our relationship carries.

You're a good guy Justin, I can hear and see it on our calls. You're also a good and dedicated dev, that I can see in our code. That's enough for me to continue working with you, despite our arguments.
I hope, with time, we get to a point where we at least interpret each other's intentions as well-meaning; that's the least colleagues can expect of each other.

I won't be holding the PR any longer over this, the changes are overall good.

misicnenad (approved):

PR #2707, opened 2024-02-07
Closes BG-1132: Fix donation validation

ap-justin:
regex used to validate donation amount allows input so long as contains number e.g. `abcd12` (valid). Could update the regex but opt to use `yup` constructs instead - this also add more specific errors
`blank` -> `required`
`adfs`. -> `invalid number`
`-1, 0 ..` -> `must be greater than 0`

unlike patter matching
`adfs`. -> `must be a number`
`-1, ..` -> `must be a number`

## Explanation of the solution
* use yup validators instead of pattern matching
* create extensible convenience schema builder `stringSchema` and `schema`
* update other simple schemas to use `stringSchema` and `schema` builders

## Instructions on making this work

- run `yarn` or `yarn install` to install npm dependencies
- run `yarn run test --watchAll` to verify all tests still pass
- (optional) run `yarn run build` to verify the build passes
- run `yarn start` to start the webapp
-

## UI changes for review

When major UI changes are introduced with a PR, please include links to URLS to compare or screenshots demonstrating the difference and notify on design changes

PR #2710, opened 2024-02-07
Remove unused registerOptions prop from Field.tsx

misicnenad:
Extends on work in PR https://github.com/AngelProtocolFinance/angelprotocol-web-app/pull/2707

### DIFF
--- a/package.json
+++ b/package.json
@@ -14,11 +14,7 @@
     "preview": "rsbuild preview"
   },
   "browserslist": {
-    "production": [
-      ">0.2%",
-      "not dead",
-      "not op_mini all"
-    ],
+    "production": [">0.2%", "not dead", "not op_mini all"],
     "development": [
       "last 1 chrome version",
       "last 1 firefox version",
--- a/package.json
+++ b/package.json
@@ -14,7 +14,11 @@
     "preview": "rsbuild preview"
   },
   "browserslist": {
-    "production": [">0.2%", "not dead", "not op_mini all"],
+    "production": [
+      ">0.2%",
+      "not dead",
+      "not op_mini all"
+    ],
     "development": [
       "last 1 chrome version",
       "last 1 firefox version",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -14,7 +14,11 @@
     "preview": "rsbuild preview"
   },
   "browserslist": {
-    "production": [">0.2%", "not dead", "not op_mini all"],
```
SovereignAndrey, line 17: @misicnenad I think you're missing some formater related changes? I recall this bit was just formatted to the original code here and this PR reverts it back.

SovereignAndrey (changes requested):
Other than formatting related stuff the actual change at hand is OK. 👍

package.json
misicnenad: I definitely am missing something. I left a related comment on one of the recent PRs https://github.com/AngelProtocolFinance/angelprotocol-web-app/pull/2707/files#r1481187581

@ap-justin do you know why this might be happening?

misicnenad:
> Other than formatting related stuff the actual change at hand is OK. 👍

@SovereignAndrey will wait on Justin to review this, maybe he knows why the formatting issue keeps happening (see https://github.com/AngelProtocolFinance/angelprotocol-web-app/pull/2710#discussion_r1481188443)

package.json
ap-justin: @misicnenad, @SovereignAndrey - might be because we are using biome, but our vscode extension is still prettier. I am using the biome vscode extension and `biome check` and format on save is the same
image

misicnenad: I removed `prettier` extension and installed `biome` as soon as we made the switch

SovereignAndrey: Hey now! Don't lump all me in with that VS code stuff. ;)

SovereignAndrey (approved):

package.json
misicnenad: Very retro of you Andrey :P

misicnenad: @ap-justin I figured out *when* this formatting happens - whenever I run `yarn install`, do you know what's causing this to happen? :open_mouth: Opening *package.json* and formatting the file inlines this back :thinking:

ap-justin (approved):

PR #2711, opened 2024-02-07
Closes BG-1111: Redesign of Header Navbar

misicnenad:

### Signed in:
- desktop:
image
image

- tablet (note: this is old logo, it is now the same as on desktop/mobile): 
image
image

- mobile:
image
image

### Signed out:

**NOTE: "Login" was renamed "Log in" and "Sign out" was renamed "Log out".**

- desktop:
image
image

- tablet: 
image
image

- mobile:
image
image

### Transition from fixed non-full width header to sticky full width one
https://www.loom.com/share/3a913705f338429fbb741a7b3aa247bd?sid=172802ce-ac4b-4ded-8912-98bbaea713cc


yarn.lock
```diff
@@ -1653,15 +1653,15 @@ __metadata:
   linkType: hard
 
 "@npmcli/agent@npm:^2.0.0":
-  version: 2.2.0
-  resolution: "@npmcli/agent@npm:2.2.0"
+  version: 2.2.1
```
misicnenad, line 1656: Not sure why pkg versions were updated, as all packages other than `typescript@<5.2.0` use fixed version numbers - `yarn install` shouldn't have done anything. Maybe the non-fixed TS pkg version is affecting this? :thinking:

package.json
```diff
@@ -14,7 +14,11 @@
     "preview": "rsbuild preview"
   },
   "browserslist": {
-    "production": [">0.2%", "not dead", "not op_mini all"],
+    "production": [
```
misicnenad, line 17: This keeps getting reformatted every other PR, not sure why

src/App/Header/UserMenu/UserMenu.tsx
```diff
@@ -14,20 +14,32 @@ export default function UserMenu() {
...
+          state={state}
+          className="btn text-base normal-case max-sm:hidden bg-blue-d1 hover:bg-blue text-white text-nowrap px-5 sm:px-6 lg:px-7 py-3 rounded-full"
+        >
+          Sign up
+        </Link>
+      </>
     );
   }
 
+  if (user === "loading") {
```
misicnenad, line 39: @SovereignAndrey added this loading indicator as it makes it clearer to users that some process is running after they sign in. 
If you prefer the former behavior (while signing in, login/sign up buttons still appear but disabled) will revert this.

src/App/Header/index.tsx
```diff
@@ -1,68 +1,33 @@
...
-    >
-      <div className="grid items-center gap-4 px-5 grid-cols-[auto_1fr_auto] h-full bg-white">
-        <DappLogo />
-        <DesktopNav
-          classes="hidden lg:flex font-heading font-bold uppercase"
-          links={links}
-        />
-        <div className="flex gap-4 justify-self-end items-center">
+    <header id={HEADER_ID} className={`${classes} ${headerClassNames}`}>
+      <div className="grid place-items-center gap-4 pl-5 pr-10 grid-cols-3 h-full w-full max-w-6xl mx-auto rounded-full bg-white">
```
ap-justin, line 20: @SovereignAndrey @misicnenad, is redesign  a total package for RC-v1.9.2 ? - which warrants temporarily shipping to staging inconsistent UIs
image

Perhaps we ship incrementally e.g. , retain the current header layout (full width rectangle) but only update its contents 
like `login/ signup btns` dropdown links
image


The totality of this new header 
also requires blue background which is not the case on every page of our app (e.g. admin has white background)

other UI inconsistency would be corner roundness - most of our UI is `rounded` while we are introduction some `rounded-full`. Perhaps these could be in batches 
task 1: update theme colors 
task 2: update container roundness

misicnenad: It shouldn't be a problem for the staging site to have inconsistent UI, but it would be a problem to have this in production.

If we wanted to keep staging site "consistent", we could create a special branch to hold all the design-related changes; that way we can merge UI rework in a batch. The downside would be that we would be setting ourselves up for some long conflict resolution sessions when trying to merge this new branch into `RC-v1.9.2`, which would still receive non-UI related updates.

----
**@SovereignAndrey @ap-justin what do you think of a completely new convention for how we organize our branches:**
- **`master` -> branch that holds production-ready code, the one accessible on the production site**
- **`staging` -> branch that holds staging code, accessible on `https://staging.better.giving/`**
- **RC-v1.9.2` -> development branch, changes that are not yet ready for the staging site.**

Changes would be merged into the "upper layer" branch in logical batches in the following order:
`RC-v1.9.2 -> staging -> master`

Merging would happen only when we are completely sure that the current batch (of changes) is ready to move to the next "layer".

misicnenad: I've been thinking about this some more and I think the proposed branch organization (the latter one) could actually complicate things a bit. In our particular case, we'd then need to spin up a new "development" webapp stage, where everyone could access the in-development webapp - this is actually what we use "staging" for atm.
In short, this would only be adding one more step to the overall development and deployment processes.

The original proposal (having a `ui-rework` branch to hold all UI changes) is much more useful IMHO. If we kept this branch up-to-date with the most recent `RC-v1.9.2` changes, then there wouldn't be that many conflicts to solve; the only thing to note here is that *all UI-related changes* (**all of them**) would need to come from and into this `ui-rework` branch.

SovereignAndrey: @misicnenad @ap-justin Will try to answer all the questions and provide feedback:
1. UI Redesign IS the total package for RC-v1.9.2 shipping (at this point) and what launch will be waiting on along with wrapping up multiple donations and the new payment options.
2. As for branch use/naming proposals here's my take:
  - `master` -> branch that holds production-ready code
  - `RC-vX.X.X` -> Release Candidate branch that holds staging server version of code
  - `FB-<name>` -> Feature branches that hold (1 or more merged in working branches). These are the branches that we can build and refine with multiple, smaller PRs to progress a feature along until we're all happy enough with everything and know with some confidence that it won't terribly break staging (or cause Operations folks to freak out over known bugs).
   - `<working-branch>` -> linked to Linear tickets and flow into a FB (for bigger items like a UI overhaul) OR into a RC branch for smaller, one-off issues or features.

misicnenad: Do we want to create this `FB-<name>` branch for the UI rework?
Or merge this into `RC-v1.9.2`?

SovereignAndrey: I feel that, since we've pushed a ton of re-design work into the `RC-v1.9.2` branch already, we can continue to do so. We'll take a `master < RC < FB < WB` approach for the next bigger features, in future releases, that would warrant that approach (Gift Cards, Leaderboard, Admin area improvements, etc).

misicnenad: Will open the PR for review then and resolve this conversation.

SovereignAndrey (commented):
@misicnenad I think the logo needs more left padding (on desktop for sure)? It seems off-balanced when compared to the white space from the right side drop-down to the edge of the white bar. Wdyt?

SovereignAndrey (changes requested):
@misicnenad One other issue: We need to allow the hero banner to flow to the top, past the navbar. Maybe can use a top offset for that with absolute positioning to get that floating effect while letting the hero area (or whatever component is first on a page) to flow as normal? 🤔 

Here's the landing page desktop view example for reference. Note how the background goes behind the nav bar.

src/App/Header/index.tsx
ap-justin: > I feel that, since we've pushed a ton of re-design work into the RC-v1.9.2 branch already, we can continue to do so. 

Yup, I'm only suggesting a more strategic way to do it so that internal testing team which constantly looking at staging won't create tickets like `Hey, staging nav bar is currently broken` - and could potentially hang in there at the moment while follow up works is being done. We could keep this confidence that what's up there is being evaluated for production and that bugs/inconsistencies are not liberally put up. 

As an example, I didn't settle on adding the [donate `Cancel` button](https://linear.app/better-giving/issue/BG-1104/move-cancel-button-at-the-top-of-the-page) without updating the layout alongside as it is out of place if I would just put it up there on the corner. 

In case of this navbar - shipping bit by bit is not that problematic: even so makes PRs even smaller. 
1.  PR1: inside header, update auth buttons, move links to dropdown
2. PR2: make blue background common to pages, make it navbar floating (not full width)
3. PR3: apply change in roundess of containers across app
4. etc..

misicnenad:
> @misicnenad One other issue: We need to allow the hero banner to flow to the top, past the navbar. Maybe can use a top offset for that with absolute positioning to get that floating effect while letting the hero area (or whatever component is first on a page) to flow as normal? 🤔
> 
> Here's the landing page desktop view example for reference. Note how the background goes behind the nav bar. 

Judging by the mockups, the background image should be set not on any specific component of the page, but the *whole page*.
This can be achieved by setting the bg image on [src/App/Layout.tsx](https://github.com/AngelProtocolFinance/angelprotocol-web-app/blob/RC-v1.9.2/src/App/Layout.tsx), which encompasses every other component (header, body, footer). This way, the current Header implementation works just fine.

SovereignAndrey:
@nenad Perfect! :)

SovereignAndrey (dismissed):
@misicnenad Some more changes needed. Notes on the attached screenshot (if it doesn't make sense lmk).

misicnenad:
> @misicnenad Some more changes needed. Notes on the attached screenshot (if it doesn't make sense lmk).
> 
> 

@SovereignAndrey 
1. You mean to make it look like this?
image
2. So the Marketplace will more-or-less remain as-is, so we can update the header width set in mockups (current one that you asked to widen) to align with it?

misicnenad:
@SovereignAndrey for tablets I used the mockup logo version; if we do not plan to implement the "search" feature *at all*, then I'd recommend we change this to the regular Logo (w/ name)
image

SovereignAndrey (dismissed):
Looks great! 
Only one change: Let's add the "better.giving" text on tablet mode as well. It seems odd to only have that view without it. It's not like we don't have the space for it. 😅

misicnenad:
@SovereignAndrey 

> Looks great! Only one change: Let's add the "better.giving" text on tablet mode as well. It seems odd to only have that view without it. It's not like we don't have the space for it. 😅

Looks good?
image

SovereignAndrey (approved):

SovereignAndrey (approved):

PR #2727, opened 2024-02-13
Closes BG-1071: BankDetails: improve selector styles

misicnenad:
See https://github.com/AngelProtocolFinance/angelprotocol-web-app/pull/2727#discussion_r1487805046

- start registration
- go to step 5 http://localhost:4200/register/steps/5
- **verify UI matches the expectations**

## UI changes for review

As requested in the [ticket](https://linear.app/better-giving/issue/BG-1071/bankdetails-improve-selector-styles)

### DIFF
--- a/src/components/BankDetails/RecipientDetails/RecipientDetailsForm/RecipientDetailsForm.tsx
+++ b/src/components/BankDetails/RecipientDetails/RecipientDetailsForm/RecipientDetailsForm.tsx
@@ -1,5 +1,6 @@
 import { ErrorMessage } from "@hookform/error-message";
 import { FetchBaseQueryError } from "@reduxjs/toolkit/query";
+import { SelectorUncontrolled } from "components/Selector/SelectorUncontrolled";
 import { Label } from "components/form";
 import { GENERIC_ERROR_MESSAGE } from "constants/common";
 import { useErrorContext } from "contexts/ErrorContext";
@@ -12,7 +13,6 @@ import {
 import { Group, ValidationContent } from "types/aws";
 import { ApplicationMIMEType } from "types/lists";
 import { IFormButtons, OnSubmit } from "../../types";
-import SelectorUncontrolled from "../SelectorUncontrolled";
 import Form from "./Form";
 
 type Props = {
--- a/src/components/BankDetails/RecipientDetails/RecipientDetailsForm/RecipientDetailsForm.tsx
+++ b/src/components/BankDetails/RecipientDetails/RecipientDetailsForm/RecipientDetailsForm.tsx
@@ -1,6 +1,6 @@
 import { ErrorMessage } from "@hookform/error-message";
 import { FetchBaseQueryError } from "@reduxjs/toolkit/query";
-import { SelectorUncontrolled } from "components/Selector/SelectorUncontrolled";
+import { NativeSelect } from "components/Selector";
 import { Label } from "components/form";
 import { GENERIC_ERROR_MESSAGE } from "constants/common";
 import { useErrorContext } from "contexts/ErrorContext";
@@ -138,7 +138,7 @@ export default function RecipientDetailsForm({
                 }}
                 render={({ field: { name, value: key, onChange, ref } }) => (
                   <>
-                    <SelectorUncontrolled
+                    <NativeSelect
                       aria-invalid={!!get(errors, name)?.message}
                       id={name}
                       onChange={(optValue) => {
--- a/src/components/BankDetails/RecipientDetails/RecipientDetailsForm/RecipientDetailsForm.tsx
+++ b/src/components/BankDetails/RecipientDetails/RecipientDetailsForm/RecipientDetailsForm.tsx
@@ -1,10 +1,14 @@
+import { Listbox } from "@headlessui/react";
 import { ErrorMessage } from "@hookform/error-message";
 import { FetchBaseQueryError } from "@reduxjs/toolkit/query";
+import { DrawerIcon } from "components/Icon";
+import { FocusableInput } from "components/Selector";
+import { styles } from "components/Selector/constants";
 import { Label } from "components/form";
 import { GENERIC_ERROR_MESSAGE } from "constants/common";
 import { useErrorContext } from "contexts/ErrorContext";
 import { isEmpty } from "helpers";
-import { useForm } from "react-hook-form";
+import { Controller, get, useForm } from "react-hook-form";
 import {
   useCreateRecipientMutation,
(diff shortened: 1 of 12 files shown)


src/components/form/Label.tsx
```diff
@@ -3,6 +3,13 @@ export function Label({
...
   ...props
 }: React.LabelHTMLAttributes<HTMLLabelElement> & {
+  /**
+   * UI depending on the value:
+   *
+   * - `true` - a red '*' (star) next to the label
+   * - `false` - text '(optional)' next to the label
+   * - `undefined` - nothing appears next to the label
+   */
   required?: boolean;
```
misicnenad, line 13: Found it confusing that UI was different for `required` values `false` and `undefined`, so added this comment for future clarity.

src/pages/Registration/Steps/Documentation/FSA/Form/index.tsx
```diff
@@ -71,7 +71,7 @@ export default function Form(props: Props) {
 
       <div className="grid grid-cols-2 sm:flex gap-2 mt-8">
         <Link
-          aria-disabled={isSubmitting}
+          aria-disabled={isSubmitting || isRedirecting}
```
misicnenad, line 74: Noticed the button becoming "enabled" when the form was redirecting, so added this check to keep it disabled.

src/components/BankDetails/RecipientDetails/RecipientDetailsForm/RecipientDetailsForm.tsx
```diff
@@ -123,31 +125,44 @@ export default function RecipientDetailsForm({
         const labelRequired = f.required ? true : undefined;
         if (f.type === "select") {
           return (
-            <div key={f.key} className="">
-              <Label required={labelRequired} htmlFor={f.key} className="mb-1">
+            <div key={f.key} className="field">
+              <Label required={labelRequired} htmlFor={f.key}>
                 {f.name}
               </Label>
-              <select
```
misicnenad, line 130: The reason I opted for creating a shared "uncontrolled" selector component instead of just manipulating native `select`'s CSS is that we already have a shared [Selector](https://github.com/AngelProtocolFinance/angelprotocol-web-app/blob/RC-v1.9.2/src/components/Selector/Selector.tsx) component with the desired styles and behaviors defined, so having additional ad hoc implementations of select, with their own CSS, would increase maintenance cost.

You may be asking "but hey, you say you wanted to avoid ad hoc implementations, but you created yet a new Selector implementation!" and you'd be right, I did. There are 2 reasons for this:
1. This "uncontrolled" Selector can be used in places where we want to use a selector without a surrounding `FormProvider`, like here in `BankDetails` (for an example of `FormProvider` usage see [ContactDetails/index.tsx](https://github.com/AngelProtocolFinance/angelprotocol-web-app/blob/RC-v1.9.2/src/pages/Registration/Steps/ContactDetails/index.tsx#L47-L49)).
2. The current `Selector.tsx` can be rewritten to use this "uncontrolled" selector as its nested component (as they share all of the styles and behavior) - that way webapp-wide Selector changes can be done by manipulating _one single component_.
The way this would be done is that:
   - `SelectorUncontrolled` would contain the UI and the behavior
   - `Selector` would set up the `react-hook-form` wires and pass them to the actual selector.
I left this for a separate Linear ticket / PR, as the change would require thorough testing, so it's outside the scope of this PR.
P.S. I'm open to name suggestions for these components, I find `SelectorUncontrolled` too long winded, but hadn't though of any better (shorter) name.
3. The "uncontrolled" Selector can in the future be used in places where we only need a plain (yet aligned) selector component.

src/components/Selector/NativeSelect.tsx
```diff
@@ -0,0 +1,63 @@
...
+  value?: OptionType<V>;
+  onChange: (value: OptionType<V>) => void;
+};
+
+export const NativeSelect = forwardRef(function Select<V extends ValKey>(
+  props: Props<V>,
+  ref: React.Ref<HTMLInputElement>
+) {
+  return (
+    <Listbox
```
ap-justin, line 22: technically, this is still controlled component (manifested by `value + onChange` ), which result to this bug

REPRO: 
- go to `admin/banking/new`
- select PHP, amount:50, 
- select E-wallet

image

image


headlessui/listbox can however be [uncontrolled](https://headlessui.com/react/listbox#using-as-an-uncontrolled-component) but I'm not sure if it could be used just like native `<select/>`
```
<select {...register(...)} /> 
```

ap-justin (changes requested):

src/components/Selector/NativeSelect.tsx
misicnenad: @ap-justin by "uncontrolled" I meant "not dependent on ('controlled by') `react-hook-form`"

About the bug - good catch, will address.

misicnenad: > headlessui/listbox can however be [uncontrolled](https://headlessui.com/react/listbox#using-as-an-uncontrolled-component) but I'm not sure if it could be used just like native `<select/>`
> ```
> <select {...register(...)} /> 
> ```

I did try some things out unsuccessfully, but I just thought of an another approach to attempt, will test it out :+1:

misicnenad:
Converted to draft, for the reason see https://github.com/AngelProtocolFinance/angelprotocol-web-app/pull/2727#discussion_r1489044931

src/components/Selector/NativeSelect.tsx
misicnenad: No luck.
Created a new branch to show the changes (see [native-select-with-register](https://github.com/AngelProtocolFinance/angelprotocol-web-app/tree/native-select-with-register)).

The attempt consisted of updating [NativeSelect to be able to accept `{...register(...)}` props](https://github.com/AngelProtocolFinance/angelprotocol-web-app/blob/native-select-with-register/src/components/Selector/NativeSelect.tsx).
Then [I updated the relevant usage of the component so that no `Controller` is necessary](https://github.com/AngelProtocolFinance/angelprotocol-web-app/blob/native-select-with-register/src/components/BankDetails/RecipientDetails/RecipientDetailsForm/RecipientDetailsForm.tsx#L131-L143).
But unfortunately, `react-hook-form > register` always expects to work with native components (its `onChange` always expects native event parameter, see [docs](https://react-hook-form.com/docs/useform/register)), so it didn't know how to handle the actual `ValKey` value.

Resolved the bug you reported by setting the initial selector value to be an empty string.

SovereignAndrey (approved):

PR #2730, opened 2024-02-15
Use native Key type instead of ValKey

misicnenad:
- ~~React already has a native type that is equal to our custom `ValKey = string | number`.~~ Revert this change, see https://github.com/AngelProtocolFinance/angelprotocol-web-app/pull/2730#issuecomment-1945623808.
- narrowed down `OptionType<V>` to be `OptionType<V extends ValKey | symbol>` (added `symbol` as well due to [this](https://github.com/AngelProtocolFinance/angelprotocol-web-app/pull/2730#issuecomment-1945729400))

### DIFF
--- a/src/components/Selector/MultiSelector.tsx
+++ b/src/components/Selector/MultiSelector.tsx
@@ -156,13 +156,17 @@ export function MultiSelector<
   );
 }
 
-type SelectedProps<T> = {
+type SelectedProps<T extends Key> = {
   option: OptionType<T>;
   selected: OptionType<T>[];
   onChange(value: OptionType<T>[]): void;
 };
 
-function SelectedOption<T>({ selected, onChange, option }: SelectedProps<T>) {
+function SelectedOption<T extends Key>({
+  selected,
+  onChange,
+  option,
+}: SelectedProps<T>) {
   const handleRemove = (value: T) =>
     onChange(selected.filter((s) => s.value !== value));
 
--- a/src/components/Selector/MultiSelector.tsx
+++ b/src/components/Selector/MultiSelector.tsx
@@ -2,7 +2,7 @@ import { Combobox } from "@headlessui/react";
 import { ErrorMessage } from "@hookform/error-message";
 import Icon, { DrawerIcon } from "components/Icon";
 import { isEmpty } from "helpers";
-import { PropsWithChildren, useState } from "react";
+import { Key, PropsWithChildren, useState } from "react";
 import {
   FieldValues,
   Path,
@@ -10,7 +10,7 @@ import {
   useController,
   useFormContext,
 } from "react-hook-form";
-import { OptionType, ValKey } from "types/components";
+import { OptionType } from "types/components";
 import FocusableInput from "./FocusableInput";
 import { styles, valueKey } from "./constants";
 import { MultiselectorProps } from "./types";
@@ -18,7 +18,7 @@ import { MultiselectorProps } from "./types";
 export function MultiSelector<
   T extends FieldValues,
   K extends Path<T>,
-  V extends ValKey,
+  V extends Key,
 >({
   name,
   disabled,
--- a/src/components/Selector/MultiSelector.tsx
+++ b/src/components/Selector/MultiSelector.tsx
@@ -2,7 +2,7 @@ import { Combobox } from "@headlessui/react";
 import { ErrorMessage } from "@hookform/error-message";
 import Icon, { DrawerIcon } from "components/Icon";
 import { isEmpty } from "helpers";
-import { Key, PropsWithChildren, useState } from "react";
+import { PropsWithChildren, useState } from "react";
 import {
   FieldValues,
   Path,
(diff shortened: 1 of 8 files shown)


ap-justin:
This is unnecessary, we shoudn't use react types that is meant for components to our option interface that just denotes shape of option just because they look the same. 

> narrowed down OptionType<V> to be OptionType<V extends Key | symbol>

I'm not sure the use case of `symbol` as a dropdown option

misicnenad:
> This is unnecessary, we shoudn't use react types that is meant for components to our option interface that just denotes shape of option just because they look the same

Fair, will revert

ap-justin:
> > This is unnecessary, we shoudn't use react types that is meant for components to our option interface that just denotes shape of option just because they look the same
> 
> Fair point, will revert.
> 
> > > narrowed down OptionType to be OptionType<V extends Key | symbol>
> > 
> > 
> > I'm not sure the use case of symbol as a dropdown option value
> 
> _Note: `OptionType` is now defined as `OptionType<V extends ValKey | symbol>` (see commit [e95a615](https://github.com/AngelProtocolFinance/angelprotocol-web-app/commit/e95a61565103d745b41b5bb9b5d508a49740e79c))._
> 
> Symbols are basically strings in this use case, so there shouldn't be any harm in allowing them. But to the point, we need to allow symbols, otherwise the following error appears in [ContactDetails/constants.ts](https://github.com/AngelProtocolFinance/angelprotocol-web-app/blob/RC-v1.9.2/src/pages/Registration/Steps/ContactDetails/constants.ts#L33): image

@misicnenad, I don't see this error appearing in `RC-v1.9.2`

misicnenad:
> @misicnenad, I don't see this error appearing in `RC-v1.9.2`

@ap-justin the error appears if you narrow down the allowed values for `V` in `OptionType<V>`.

ap-justin:
> > @misicnenad, I don't see this error appearing in `RC-v1.9.2`
> 
> @ap-justin the error appears if you narrow down the allowed values for "V" in `OptionType<V>` just to `V extends ValKey`:
> 
> ```ts
> export type OptionType<V extends ValKey> = { label: string; value: V }; // error appears in ContactDetails/constants.ts
> ```

perhaps there's no need to narrow it down? - as `<Selector />`'s `V` generic extends `ValKey` so putting some `Option<SomeContrivedType>`  would be rejected

src/components/Selector/MultiSelector.tsx
```diff
@@ -156,13 +156,17 @@ export function MultiSelector<
   );
 }
 
-type SelectedProps<T> = {
+type SelectedProps<T extends ValKey> = {
```
ap-justin, line 159: no need to narrow this down as this is internal to `Multiselector` that is already bound by `ValKey`

ap-justin (changes requested):

misicnenad:
> perhaps there's no need to narrow it down? - as `<Selector />`'s `V` generic extends `ValKey` so putting some `Option<SomeContrivedType>` would be rejected

@ap-justin `OptionType` is only ever used with `Selector`, so it is more type-safe to have their generic type definitions match.

src/components/Selector/MultiSelector.tsx
misicnenad: No harm in narrowing it down.

ap-justin:
> > perhaps there's no need to narrow it down? - as `<Selector />`'s `V` generic extends `ValKey` so putting some `Option<SomeContrivedType>` would be rejected
> 
> @ap-justin `OptionType` is only ever used with `Selector`, so it is more type-safe to have their generic type definitions match. Removed `symbol` and updated _ContactDetails/constants.ts_ to avoid the error, see commit [c953b8c](https://github.com/AngelProtocolFinance/angelprotocol-web-app/commit/c953b8c27389dc126e76b53426d8204d79469c50)

could you give at least a working example that `OptionType<V>` would cause some bug in the current setup ?

src/components/Selector/MultiSelector.tsx
ap-justin: Why are we suddenly narrowing types of an internal implementation of `Multiselector` which api is already narrowed down for safe use? Do we plan to use `<SelectedProps/>` outside of `Multiselector` ?  I mean, what's even the point of this PR? I've been going here back and forth away from more important task. @SovereignAndrey

SovereignAndrey (approved):
@misicnenad Seems OK. Though this PR ought to have been a low-priority backlog issue that was simply opened instead, as it is not a user facing bug, nor is it a critical refactoring, and could have waited.

misicnenad:
@SovereignAndrey @ap-justin 

This small PR is getting more attention than I anticipated and I'm not sure that attention is warranted.

The PR was created because I had 3 minutes to spare and we never created special Linear tickets for these types of changes, as the changes are as small as minor package version updates:
- adds some minor type-safety to an internal component
- adds some minor type-safety to an in-use type
- adds a single comment

I'm not sure I understand the aversion towards it, it's not like the PR includes some major logic overhaul.

PR #2744, opened 2024-02-26
Update RichTextEditor content parameter to always keep track of number of characters

misicnenad:
Closes BG-1147

Source of the problem was obvious - rich text strings contain more than plain user input; they are actually a stringified object that contains styling-related fields and subfields. Calculating the char length for rich texts would require parsing the rich text string, iterating over all of its fields, accessing the appropriate subfields, removing trailing newline chars.

Instead, we can use the fact that we already calculate the exact char length in `RichText`, so just passed this value up as part of the `onChange` hook.

Added a comment explaining this [to the type itself](https://github.com/AngelProtocolFinance/angelprotocol-web-app/pull/2744/files#diff-21bebfd3054f5334cdbb7e9249a7ffd6b118b46ecced5f45ef61d54aefd519f0R21).
 
- sign in with an account that has an associated endowment where you can edit profile and/or add/edit programs
- go to http://localhost:4200/admin/{{ID}}/program-editor/new
- scroll down to description editor
- paste the text that failed validation (see issue's description for the Loom link)
```
Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua. Viverra maecenas accumsan lacus vel facilisis volutpat. Laoreet non curabitur gravida arcu ac tortor dignissim convallis aenean. Vitae aliquet nec ullamcorper sit. Eget velit aliquet sagittis id consectetur purus ut. Blandit volutpat maecenas volutpat blandit aliquam. Lectus sit amet est placerat. Amet dictum sit amet justo donec. Tortor condimentum lacinia quis vel eros
```
- **verify the issue no longer appears**

### DIFF
--- a/src/pages/Admin/Charity/EditProfile/schema.ts
+++ b/src/pages/Admin/Charity/EditProfile/schema.ts
@@ -16,7 +16,7 @@ export const VALID_MIME_TYPES: ImageMIMEType[] = [
 ];
 
 export const MAX_SIZE_IN_BYTES = 1e6;
-export const MAX_CHARS = 4000
+export const MAX_CHARS = 4000;
 
 // we only need to validate the pre-crop image and if we confirm it is valid
 // we can be sure that the cropped image is valid too
@@ -46,5 +46,5 @@ export const schema = object<any, SchemaShape<FV>>({
     youtube: url,
     tiktok: url,
   }),
-  overview: richTextContent({ maxChars: MAX_CHARS })
+  overview: richTextContent({ maxChars: MAX_CHARS }),
 }) as ObjectSchema<FV>;
--- a/src/pages/Admin/Charity/EditProfile/schema.ts
+++ b/src/pages/Admin/Charity/EditProfile/schema.ts
@@ -1,12 +1,12 @@
+import { ObjectSchema, array, object } from "yup";
+import { FV } from "./types";
+import { SchemaShape } from "schemas/types";
+import { ImageMIMEType } from "types/lists";
 import { ImgLink } from "components/ImgEditor";
-import { MAX_SDGS } from "constants/unsdgs";
 import { genFileSchema } from "schemas/file";
 import { optionType, richTextContent } from "schemas/shape";
 import { requiredString, url } from "schemas/string";
-import { SchemaShape } from "schemas/types";
-import { ImageMIMEType } from "types/lists";
-import { ObjectSchema, array, object } from "yup";
-import { FV } from "./types";
+import { MAX_SDGS } from "constants/unsdgs";
 
 export const VALID_MIME_TYPES: ImageMIMEType[] = [
   "image/jpeg",
--- a/src/pages/Admin/Charity/EditProfile/schema.ts
+++ b/src/pages/Admin/Charity/EditProfile/schema.ts
@@ -1,12 +1,12 @@
-import { ObjectSchema, array, object } from "yup";
-import { FV } from "./types";
+import { ImgLink } from "components/ImgEditor";
+import { MAX_SDGS } from "constants/unsdgs";
+import { genFileSchema } from "schemas/file";
+import { optionType, richTextContent } from "schemas/shape";
+import { requiredString, url } from "schemas/string";
 import { SchemaShape } from "schemas/types";
 import { ImageMIMEType } from "types/lists";
-import { ImgLink } from "components/ImgEditor";
-import { genFileSchema } from "schemas/file";
-import { optionType } from "schemas/shape";
-import { requiredString, url } from "schemas/string";
-import { MAX_SDGS } from "constants/unsdgs";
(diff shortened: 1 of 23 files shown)


SovereignAndrey (approved):
Beautiful! Please be sure to cherry pick the needed changes from here over to `RC-v1.9.2`. Thanks!

src/pages/Admin/Charity/EditProfile/schema.ts
```diff
@@ -16,6 +16,7 @@ export const VALID_MIME_TYPES: ImageMIMEType[] = [
 ];
 
 export const MAX_SIZE_IN_BYTES = 1e6;
+export const MAX_CHARS = 4000;
```
ap-justin, line 19: bug: 
image

REPRO:
put more than `4000` chars - error appears
delete to less than `4000` - error should disappear

ap-justin (changes requested):

src/pages/Admin/Charity/EditProfile/schema.ts
ap-justin, line 19: bug: chars is more than `4000` but can still submit

ap-justin (changes requested):

src/pages/Admin/Charity/EditProfile/schema.ts
misicnenad: Can you post repro steps?

misicnenad: This is a bug on `master`, so unrelated to this PRs focus. @SovereignAndrey do we have a ticket for this?

ap-justin: > Can you post repro steps?

1. go to edit-profile
2. input more than 4000 chars on overview
3. submit

ap-justin: This PR introduces new schema structure for Richtext, no? Would it make sense to expect that those schema handle error (whether previously existing) as they arise and resolved ?

misicnenad: This also seems to be a bug inherited from `master`, so outside the scope of this PR.

misicnenad: Created a ticket https://linear.app/better-giving/issue/BG-1173/edit-profile-stop-submission-when-overview-contains-more-than-the

ap-justin: you applied explicit validation via schema `richTextContent` to `profile.overview`. Do we really need another PR just to say that previous  `RichTextEditor.onError` is not needed anymore? - can't we just delete that in this PR ?

ap-justin: you applied `richTextContent` to `EditProfile`, no ? So we just accept that it doesn't work (can't detect error) and approve this PR ? Are we now okay to ship something that doesn't work? @SovereignAndrey  - with the reason the we could just create another PR to fix it. Isn't the point of this PR is to fix validation concern? 

Even new programs that I created exceeds `500` chars , and it submitted (not catched by `richTextContent`).

ap-justin (changes requested):

src/pages/Admin/Charity/EditProfile/schema.ts
ap-justin: created PR to do just this
* https://github.com/AngelProtocolFinance/angelprotocol-web-app/pull/2751

ap-justin: created pr to do just this 
* https://github.com/AngelProtocolFinance/angelprotocol-web-app/pull/2751

SovereignAndrey:
@ap-justin Thank you for the quick PR fix the few bugs found. Tested locally with staging and looks like everything works as it should with excess 4000 >> block submission >> back under 4000 no error around box & submission now works. :+1:

SovereignAndrey (approved):

src/pages/Admin/Charity/EditProfile/schema.ts
misicnenad: @SovereignAndrey @ap-justin 3 reasons I like to keep my PRs focused:
1. Easier to track in Linear what was done, when and by whom
2. Easier to review
3. Easier to revert if something goes wrong

misicnenad: @ap-justin there's no need to make a big deal out of this. I fixed the issue of RichTextEditor not allowing the max. number of characters it's expected to allow and not to fix an unrelated issue where a form would submit even when one of its fields is invalid. Also see https://github.com/AngelProtocolFinance/angelprotocol-web-app/pull/2744#discussion_r1505455417

PR #2753, opened 2024-02-29
Closes BG-1152: Signup on success

ap-justin:
* create multistep signup component : signup-form -> confirm-email-code-form -> success-page (contains a link that redirects to `/signin` page)
* persist donor info in local storage for fiat checkouts, so that redirect page can have access later on

* create temp mail
* donate with `card` ( crypto and paypal will error out as server payload is different from RC-v1.9.2  web-app)

-

## UI changes for review

donate page success
image

fiat success redirect page
image

confirm code page
image


SovereignAndrey (changes requested):
@ap-justin I'm unable to get through the donation flow (with crypto). 
- `"tipAmount" is required` error is encountered 



From this URL endpoint: https://fctqkloitc.execute-api.us-east-1.amazonaws.com/dev/crypto-donation

ap-justin:
> @ap-justin I'm unable to get through the donation flow (with crypto).
> 
> * `"tipAmount" is required` error is encountered
> 
> 
> 
> From this URL endpoint: https://fctqkloitc.execute-api.us-east-1.amazonaws.com/dev/crypto-donation

@SovereignAndrey, I'm not sure if [lambda code updating](https://linear.app/better-giving/issue/BG-1164/v2-donations-pipeline-in-apes-aws) is already done @jp-mariano (as stripe donation is not erroring out about the `tipAmount`), so I could open PR 
* https://github.com/AngelProtocolFinance/angelprotocol-web-app/pull/2719
that contains the compatible payload

a temp workaround would be to modify `src/slices/donation/sendDonation/index.ts: 53` to  (apologies: not included in PR description)
```js
  body: JSON.stringify({ ...payload, tipAmount: 0, splitLiq: 50 }),
```

SovereignAndrey:
> > @ap-justin I'm unable to get through the donation flow (with crypto).
> > 
> > * `"tipAmount" is required` error is encountered
> > 
> > 
> > From this URL endpoint: https://fctqkloitc.execute-api.us-east-1.amazonaws.com/dev/crypto-donation
> 
> @SovereignAndrey, I'm not sure if [lambda code updating](https://linear.app/better-giving/issue/BG-1164/v2-donations-pipeline-in-apes-aws) is already done @jp-mariano (as stripe donation is not erroring out about the `tipAmount`), so I could open PR
> 
> * [Towards BG-1106, Closes BG-1131: Donate to BG step #2719](https://github.com/AngelProtocolFinance/angelprotocol-web-app/pull/2719)
>   that contains the compatible payload
> 
> a temp workaround would be to modify `src/slices/donation/sendDonation/index.ts: 53` to (apologies: not included in PR description)
> 
> ```js
>   body: JSON.stringify({ ...payload, tipAmount: 0, splitLiq: 50 }),
> ```

Makes sense! Thanks. I'll give fiat a shot first and then try crypto with the patch. :+1:

SovereignAndrey (changes requested):
@ap-justin Was able to do a complete donation with fiat. :+1: 
Found an issue with how we check if a user exists. In the below screenshot I managed to do the following: 
Not logged, in completed the donations email with `andrey@better.giving` (an email that exists in the DB - google auth). The final success screen shows me the signup form to set a password. This is not good, as it would then have setup two accounts for the same email if I confirmed it(I suspect) with possibly differing passwords. 

Screenshot from 2024-02-29 15-34-24

We should check if the email entered is in the UsersV2 DB as part of deciding  which success version to show. 
- [ ] If the email is in the DB, then show the logged-in version (w/o signup option)

ap-justin:
> @ap-justin Was able to do a complete donation with fiat. 👍 Found an issue with how we check if a user exists. In the below screenshot I managed to do the following: Not logged, in completed the donations email with `andrey@better.giving` (an email that exists in the DB - google auth). The final success screen shows me the signup form to set a password. This is not good, as it would then have setup two accounts for the same email if I confirmed it(I suspect) with possibly differing passwords.
> 
> Screenshot from 2024-02-29 15-34-24
> 
> We should check if the email entered is in the UsersV2 DB as part of deciding which success version to show.
> 
> * [ ]  If the email is in the DB, then show the logged-in version (w/o signup option)

@SovereignAndrey, I believe this is a cognito issue (though not sure if really an issue) in general - I noticed before that account created with `username + password` is different with federated account (from google). Could verify this on `/signin` our signin page.  

I view it as: an entiy `x@mail.com` can either login by providing password (`username+password` account), or just by federated signin (no need for password - as already taken care of by provider)

src/components/Signup/ConfirmForm.tsx
```diff
@@ -0,0 +1,97 @@
...
+            !result.isSignUpComplete
+          ) {
+            return handleError("Code confirmation failed");
+          }
+
+          props.setSignupState("success");
+        } catch (err) {
+          const message =
+            err instanceof AuthError ? err.message : GENERIC_ERROR_MESSAGE;
+          handleError(message);
```
misicnenad, line 53:
```suggestion
          handleError(err, message);
```
We should still log the original error.

src/components/Signup/ConfirmForm.tsx
```diff
@@ -0,0 +1,97 @@
...
+            //no need to inspect result
+            await resendSignUpCode({
+              username: props.codeRecipientEmail.raw,
+            });
+
+            return alert("New code has been sent to your email.");
+          } catch (err) {
+            const message =
+              err instanceof AuthError ? err.message : GENERIC_ERROR_MESSAGE;
+            handleError(message);
```
misicnenad, line 87:
```suggestion
            handleError(err, message);
```
Same as above.

src/components/Signup/SignupForm.tsx
```diff
@@ -0,0 +1,108 @@
...
+  } = useForm({
+    resolver: yupResolver(
+      object({
+        password: requiredString
+          .min(8, ({ min }) => `must have at least ${min} characters`)
+          .matches(/[a-z]/, "must have lowercase letters")
+          .matches(/[A-Z]/, "must have uppercase letters")
+          .matches(/\d/, "must have numbers")
+          .matches(
+            /[!@#$%^&*()_+{}\[\]:;<>,.?~\\/-]/,
```
misicnenad, line 35:
```suggestion
            /[! @#$%^&*()_+{}\[\]:;<>,.?~\\/-]/,
```
Should include space characters

src/components/Signup/SignupForm.tsx
```diff
@@ -0,0 +1,108 @@
...
+          props.setSignupState({
+            codeRecipientEmail: {
+              raw: props.donor.email,
+              obscured: nextStep.codeDeliveryDetails.destination,
+            },
+          });
+        } catch (err) {
+          const message =
+            err instanceof AuthError ? err.message : GENERIC_ERROR_MESSAGE;
+          handleError(message);
```
misicnenad, line 74:
```suggestion
          handleError(err, message);
```

src/components/Signup/SignupForm.tsx
```diff
@@ -0,0 +1,108 @@
...
+              userAttributes: {
+                given_name: props.donor.firstName,
+                family_name: props.donor.lastName,
+              },
+              autoSignIn: false,
+            },
+          });
+
+          //per cognito config
+          if (nextStep.signUpStep !== "CONFIRM_SIGN_UP") throw "";
```
misicnenad, line 61: Shouldn't we throw some kind of error message, at least for debugging purposes? We don't have to show it to the users (the `catch` block is already set up correctly)

src/components/Signup/SignupForm.tsx
```diff
@@ -0,0 +1,108 @@
...
+      </div>
+      <p className="text-xs text-red text-right mt-1">
+        {errors.password?.message}
+      </p>
+
+      <button
+        type="submit"
+        className="btn-blue rounded-full normal-case px-4 w-full mt-4"
+      >
+        {isSubmitting ? "Submitting..." : "Create account"}
```
misicnenad, line 104: Suggestion: could use `LoadText` here

src/helpers/donation.ts
```diff
@@ -0,0 +1,29 @@
...
+export const persistDonor = (donor: Donor) => {
+  window.localStorage.setItem(storageKey, JSON.stringify(donor));
+};
+
+export const persistedDonor = (): Donor | null => {
+  try {
+    return object(shape).cast(
+      JSON.parse(window.localStorage.getItem(storageKey) ?? "")
+    ) as Donor;
+  } catch (_) {
```
misicnenad, line 26: Shouldn't log the error here just in case?

misicnenad (changes requested):

src/components/Signup/SignupForm.tsx
```diff
@@ -0,0 +1,108 @@
...
+        } catch (err) {
+          const message =
+            err instanceof AuthError ? err.message : GENERIC_ERROR_MESSAGE;
+          handleError(message);
+        }
+      })}
+    >
+      <div className='grid grid-cols-[auto_1fr_auto] border border-prim rounded items-center px-3 has-[:focus]:ring-2 ring-blue ring-offset-1  has-[input[aria-invalid="true"]]:border-red has-[:disabled]:bg-gray-l3'>
+        <Icon type="Padlock" className="mr-3 text-gray" />
+        <input
```
misicnenad, line 80: Input should have height of 52px as per Figma

src/components/Signup/SignupForm.tsx
```diff
@@ -0,0 +1,108 @@
...
+          {...register("password")}
+          type={isPasswordShown ? "text" : "password"}
+          className="h-full focus:outline-none bg-transparent"
```
misicnenad, line 83: Input should have a lightly tinted background as per Figma

misicnenad (changes requested):

src/components/Signup/Signup.tsx
```diff
@@ -0,0 +1,26 @@
+import { useState } from "react";
+import ConfirmForm from "./ConfirmForm";
+import SignupForm from "./SignupForm";
+import Success from "./Success";
+import { Donor, SignupState } from "./types";
+
+type Props = { classes?: string; donor: Donor };
+
+export default function Signup({ classes = "", donor }: Props) {
```
misicnenad, line 9: I assume it was created to be shared between `DonateFiatThanks` and the Sign up page. This can be done, but need to clarify some things first.

The problem is that the sign up page has too many different components that are interspersed between the ones in this components:
- The "You are signing up as" section
- first name, last name, email fields

Basically the only common fields are: 1. password field & 2. sign up button

We could handle this like the following:
1. if `donor` is defined, treat this as a subcomponent of DonateFiatThanks
2. otherwise, treat this as a subcomponent of the actual Sign up page with all the other fields

In the former case we would hide all the unnecessary fields/components, while in the latter we would display them. This would result in many `{!!donor && <SomeFieldOrComponent />}` instances.

If this is ok with you, we can leave this as-is. If not, we should move this component to be a part of the `pages/DonateFiatThanks/` folder and I'll create a completely new Signup page with some copy/pasted from here.

misicnenad (changes requested):

SovereignAndrey:
> > @ap-justin Was able to do a complete donation with fiat. 👍 Found an issue with how we check if a user exists. In the below screenshot I managed to do the following: Not logged, in completed the donations email with `andrey@better.giving` (an email that exists in the DB - google auth). The final success screen shows me the signup form to set a password. This is not good, as it would then have setup two accounts for the same email if I confirmed it(I suspect) with possibly differing passwords.
> > Screenshot from 2024-02-29 15-34-24
> > We should check if the email entered is in the UsersV2 DB as part of deciding which success version to show.
> > 
> > * [ ]   If the email is in the DB, then show the logged-in version (w/o signup option)
> 
> @SovereignAndrey, I believe this is a cognito issue (though not sure if really an issue) in general - I noticed before that account created with `username + password` is different with federated account (from google). Could verify this on our`/signin` page.
> 
> I view it as: an entiy `x@mail.com` can either login by providing password (`username+password` account), or just by federated signin (no need for password - as already taken care of by provider) 

@ap-justin Yes, it's def a Coginto thing. No arguments there! :) This scenario would also apply if they had created both account types (google & password-based) or if that just used a password-based account from the start. By not checking if the passed email is in the Users DB we would present them with a signup form that would then fail because they already had an account in the DB.

It's not good UX to ask someone who signed up already with one method to make a second account w/ a password. It's possible to do so, yes, but should we openly flaunt that? Probably not. Better to just have a check step with the DB and treat them like a logged in user for the purposes of that final display screen instead of making them go through the hassle again. 

**COS Scenarios for a non-logged in donor:**
- [ ] User has no accounts with supplied email --> see signup prompt @ success screen
- [ ] User signed up supplied email previously --> see see logged in user success screen

ap-justin:
> > > @ap-justin Was able to do a complete donation with fiat. 👍 Found an issue with how we check if a user exists. In the below screenshot I managed to do the following: Not logged, in completed the donations email with `andrey@better.giving` (an email that exists in the DB - google auth). The final success screen shows me the signup form to set a password. This is not good, as it would then have setup two accounts for the same email if I confirmed it(I suspect) with possibly differing passwords.
> > > Screenshot from 2024-02-29 15-34-24
> > > We should check if the email entered is in the UsersV2 DB as part of deciding which success version to show.
> > > 
> > > * [ ]    If the email is in the DB, then show the logged-in version (w/o signup option)
> > 
> > 
> > @SovereignAndrey, I believe this is a cognito issue (though not sure if really an issue) in general - I noticed before that account created with `username + password` is different with federated account (from google). Could verify this on our`/signin` page.
> > I view it as: an entiy `x@mail.com` can either login by providing password (`username+password` account), or just by federated signin (no need for password - as already taken care of by provider)
> 
> @ap-justin Yes, it's def a Coginto thing. No arguments there! :) This scenario would also apply if they had created both account types (google & password-based) or if that just used a password-based account from the start. By not checking if the passed email is in the Users DB we would present them with a signup form that would then fail because they already had an account in the DB.
> 
> It's not good UX to ask someone who signed up already with google to make a second account w/ a password. It's possible to do so, yes, but should we openly flaunt that? Probably not. Better to just have a check step with the DB and treat them like a logged in user for the purposes of that final display screen instead of making them go through the hassle again.
> 
> **COS Scenarios for a non-logged in donor:**
> 
> * [ ]  User has no accounts with supplied email --> see signup prompt @ success screen
> * [ ]  User signed up supplied email previously --> see see logged in user success screen

@SovereignAndrey, got these additional requirements - will convert this to draft and create first an endpoint to check if user is logged in

ap-justin:
@SovereignAndrey 
> COS Scenarios for a non-logged in donor:

>  User has no accounts with supplied email --> see signup prompt @ success screen
> User signed up supplied email previously --> see see logged in user success screen

> @SovereignAndrey, got these additional requirements - will convert this to draft and create first an endpoint to check if user is logged in

started working on a `user/{email}` endpoint realized that this endpoint would be public, as donation pages are public. Even with limitations on what gets return, would still end up to some `isUserSignedUp` info. I'm not sure if it's okay for us to have an endpoint like this that exposes whether an email has an account in our app our not. 

I also looked on `aws-amplify/auth` but no api to like some `checkUserIsSignedIn`.  On the otherdhand, a hacky approach could be 
```js
useEffect(() => {
 try {
  await awsAmplifyAuth.signIn(persistedDonor.email)
 } catch(err){
   //inspect err and look for some `UserDoesNotExist`
 }
},[])
```
Let me know ser what you think

misicnenad:
> started working on a `user/{email}` endpoint realized that this endpoint would be public, as donation pages are public. Even with limitations on what gets return, would still end up to some `isUserSignedUp` info. I'm not sure if it's okay for us to have an endpoint like this that exposes whether an email has an account in our app our not.

@ap-justin @SovereignAndrey you're right, this is a security vulnerability. 

What do you think about updating donations status checker logic in AWS (`paypal-proxy > captureOrder` and `stripe-payment-status`) to have an additional responsibility - to check that the user associated with the donation exists in our DB?
That way we do not expose any sensitive data to the public and we get the info whether a user is already registered.
- anon donates
- check donation status & whether the supplied email already exists in our DB
- return `{ status, isUserSignedUp }` to the webapp
- update UI accordingly

src/components/Signup/ConfirmForm.tsx
ap-justin: I think only unexpected errors should be captured by sentry. but looking at current `handleError` seems like all sorts of error is captured even though some errors are made and can be corrected by user e.g. incorrect confirmation code, signing-up with taken email. 

Perhaps `handleError` would need to distinguish between unexpected (app failed) and expected (user failed)
e.g.
```
catch (err) {
if (err instanceof AuthError){
   return showModal(Error, err.message)
}
handleUnexpectedError(error)
}
```

src/components/Signup/SignupForm.tsx
```diff
@@ -0,0 +1,108 @@
...
+  const {
+    register,
+    formState: { errors, isSubmitting },
+    handleSubmit,
+  } = useForm({
+    resolver: yupResolver(
+      object({
+        password: requiredString
+          .min(8, ({ min }) => `must have at least ${min} characters`)
+          .matches(/[a-z]/, "must have lowercase letters")
```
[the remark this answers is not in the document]
ap-justin: We should abide by what cognito sets (per our cognito config) (used by [authenticator-ui](https://github.com/aws-amplify/amplify-ui/blob/main/packages/ui/src/machines/authenticator/defaultServices.ts) ). Trying the same password in `/sign-in` would fail 
image

src/components/Signup/SignupForm.tsx
ap-justin: fixed

ap-justin:
> > started working on a `user/{email}` endpoint realized that this endpoint would be public, as donation pages are public. Even with limitations on what gets return, would still end up to some `isUserSignedUp` info. I'm not sure if it's okay for us to have an endpoint like this that exposes whether an email has an account in our app our not.
> 
> @ap-justin @SovereignAndrey you're right, this is a security vulnerability.
> 
> What do you think about updating donations status checker logic in AWS (`paypal-proxy > captureOrder` and `stripe-payment-status`) to have an additional responsibility - to check that the user associated with the donation exists in our DB? That way we do not expose any sensitive data to the public and we get the info whether a user is already registered.
> 
> * anon donates
> * check donation status & whether the supplied email already exists in our DB
> * return `{ status, isUserSignedUp }` to the webapp
> * update UI accordingly

@misicnenad, thanks for your input. @SovereignAndrey, I would go on to implement this: takes away additional loading (query isUserSignedUp) from the success page

ap-justin:
> > > started working on a `user/{email}` endpoint realized that this endpoint would be public, as donation pages are public. Even with limitations on what gets return, would still end up to some `isUserSignedUp` info. I'm not sure if it's okay for us to have an endpoint like this that exposes whether an email has an account in our app our not.
> > 
> > 
> > @ap-justin @SovereignAndrey you're right, this is a security vulnerability.
> > What do you think about updating donations status checker logic in AWS (`paypal-proxy > captureOrder` and `stripe-payment-status`) to have an additional responsibility - to check that the user associated with the donation exists in our DB? That way we do not expose any sensitive data to the public and we get the info whether a user is already registered.
> > 
> > * anon donates
> > * check donation status & whether the supplied email already exists in our DB
> > * return `{ status, isUserSignedUp }` to the webapp
> > * update UI accordingly
> 
> @misicnenad, thanks for your input. @SovereignAndrey, I would go on to implement this: takes away additional loading (query isUserSignedUp) from the success page



UPDATE: 
realized adding donor email in endpoint  `v2/fiat/stripe-proxy/?payment_intent=${paymentIntentId}` e.g. `?donor=mail@mail.com` and make it return `isUserSignedUp` is basically just the same with a standalone `user/{email}` endpoint

another way (for stripe) is to [include donor in stripe-proxy](https://github.com/AngelProtocolFinance/devops/pull/493) and include in `createPaymentIntent.metadata` so when `stripe-payment-status` retrieves the intent, metadata is already included and not need be passed when querying stripe payment status. However, this (attaching metadata) is not possible with paypal `createOrder` and `captureOrder`

ap-justin:
@SovereignAndrey, though we're supposed to develop things and not talk about UX (as this must have been already talked through before the design even came out), I would like point out things if I were to be a donor of this platform, and perhaps biased to simplify engineering implementation. 

first signup form is vague as to why only password is required - It is not effectively communicated that information from `summary` step is used to fill out `name`, and `email`. 
image

I would only know which email was used once confirm pops out. 
image

Worse is that I would get some error that account already exist but doesn't know which account it is I'm trying to use to register. 
image




An alternative to this is just a simple instruction e.g.
image
(todo: elaborate what `to manage` means: get receipt, download list of donations, see donation history, or what `your account` is supposed to be: user dashboard / my donations)


* If user is logged in, simply goes to this `your account` (dashboard/ donations etc.)
* if not logged in but have account (user knows this better than we do, and if they are not logged in when they donate, we should give them reason why they need to login)
image

* If not logged in and doesnt have account, when they click `go to your account`. They would just be redirected to `signin`,(they know better than we do that they are not registered) so the next logical step is for them to `/signup` - and they would likely do so if they are given reasons upfront why an account is beneficial (see donation history, request receipt, etc. )

src/components/Signup/SignupForm.tsx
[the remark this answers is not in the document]
misicnenad: Makes sense to align these :+1: Weird that they would narrow the number of allowed characters to just the English alphabet.

ap-justin:
> > UPDATE: realized adding donor email in endpoint `v2/fiat/stripe-proxy/?payment_intent=${paymentIntentId}` e.g. `?donor=mail@mail.com` and make it return `isUserSignedUp` is basically just the same with a standalone `user/{email}` endpoint
> > another way (for stripe) is to [include donor in stripe-proxy](https://github.com/AngelProtocolFinance/devops/pull/493) and include in `createPaymentIntent.metadata` so when `stripe-payment-status` retrieves the intent, metadata is already included and not need be passed when querying stripe payment status. However, this (attaching metadata) is not possible with paypal `createOrder` and `captureOrder`
> 
> @ap-justin you may not even need to pass the donator email as a query parameter, just the `payment_intent` for Stripe and `orderId` for PayPal should be enough - these fields are used to populate the `transactionId` field in on-hold-donations table, so you can get user's email by querying it.
> 
> * hit endpoint `v2/fiat/stripe-proxy/?payment_intent=${paymentIntentId}`
> * extract the payment intent ID
> * check payment status
> * query on-hold-donations by filtering those where `transactionId === paymentIntentId`
> * get email associated with the donation
> * check user DB
> * return `{ status, isUserSignedUp }`
> 
> Similar thing can be done for PayPal, just swap "payment intent" with "order".

right, would go for this

ap-justin:
@SovereignAndrey, user signup checking is now included 🤞

SovereignAndrey:
> @SovereignAndrey, user signup checking is now included 🤞

Cool! Will pull and test after reviewing the code changes. Thanks!

SovereignAndrey (approved):

PR #2756, opened 2024-03-01
Towards BG-1133: New brand colors included 

ap-justin:
* Include new brand colors (can now be used on new UIs after this PR) 
* adjust existing shades to preserve color hierarchy of existing UIs not included in new designs. i.e. replacing `green.DEFAULT` all lights (`-l-n`) should be adjusted to the new base color

* replaced `bluegray (old)` with `navy-blue (new)`, mapped to preserve hierarchy
```
bluegray-d2 -> navy-d4 
bluegray-d1 -> navy-d3
bluegray-default -> navy-default
```

Color pallete from `tailwind.config.json` can be view using [extension](https://marketplace.visualstudio.com/items?itemName=KalimahApps.tailwind-config-viewer)

-

## UI changes for review
* UI should still remain coherent


SovereignAndrey (changes requested):
@ap-justin I think we'll need to tweak the blues a bit. From the Figma, the default blue seems like it ought to be `#2D89C8` 400 and the `blue-l1` could be set as `#2D89C899` (ie. default color @ 60 opacity)? We could also stand to try and reduce the number of blues overall. At the least shoot to get dark/light blues in balance w/ 5 shades or less each? From the more limited selection we really don't need quite as many colors in either direction from DEFAULT.

ap-justin:
@SovereignAndrey 
> I think we'll need to tweak the blues a bit. From the Figma, the default blue seems like it ought to be #2D89C8 400 and the blue-l1 could be set as #2D89C899 (ie. default color @ 60 opacity)

To me, `DEFAULT` has no semantic value (other than to tell its relative position in the set)  - it could have been just `500` (e.g. `green-500`, but used `DEFAULT` instead as it translates to only `green` when used in `className`, if we want, we could totally omit `DEFAULT` and opt for `0` as in `l2,l1,0,d1,d2` to completely eliminate the misconception around its meaning ) to denote that it's the middle of the range. 

In case of blue `#2D89C8`, it being retained in `blue-d2` is to simply preseve that some UI might be using it as a darker shade e.g. `bg-blue hover:bg-blue-l1, active:bg-blue-d2`. However, for the donation redesign, we would be using `bg-blue-d2 #2D89C8` as required (I believe some of the new UIs I shipped, I used `blue-d2` when applicable )

Another example: our current text color. Our `"default"` is actually not `text-gray` but two shades darker `text-gray-d2`. and secondary text is `text-gray-d1` (1 shade lighter from `text-gray-d2`)

If you look closely to other colors, say `red`, the new `#C81818`, is not mapped to our existing `red`, instead it's set at equivalent to `red-d2`, 

image

so that when expanded, its `500/DEFAULT/0` would map exactly to what we currently used as `red`. This is the same method used to include new brand colors (shades are generated to match existing colors )



> We could also stand to try and reduce the number of blues overall. At the least shoot to get dark/light blues in balance w/ 5 shades or less each? From the more limited selection we really don't need quite as many colors in either direction from DEFAULT.

I actually tried to just YOLO on this, remove our old colors, and replace strictly with what's listed on figma. I would start then working on donations page to apply the new set (as this is where these colors came from anyway) - but I started to think, what would be of other pages then: `admin, applications, marketplace, profile`, Would simply squashing existing color hierarcy to the new limited colors

e.g. `BLUE: d7,d6,d5,d4,d3,d2,d1,DEFAULT,l1,l2,l3,l4` -> `blue,50,100,400`
`RED:d4,d3,d2,d1,DEFAULT,l1,l2,l3,l4,l5 -> `red, red/opacity60`

 would make them look the way the designers would have intended for it to look had they designed it? 



> From the more limited selection we really don't need quite as many colors in either direction from DEFAULT.

Contrary to this, I think we need to expand on what the design team has given to us, As most of the time, we are left to decide on things that isn't very specific.
e.g. this piece of UI

image

previous design just says button must `orange`. and other stuff is left to us
1. what's the color if hovered?
2. what's the color if disabled?
3. what's the outline if tabbed?
4. should error message be just `red` in` blue` background? - we are able to dial it lighter because we have a lighter red upfront. 

By expanding on a brand color upfront (brand color doesn't need to be in the middle), we'll always have the flexibility that we thought we might not need when we need it. Also, the added flexibility doesn't necessarily mean we stray away from the brand colors design had chosen - we are merely expanding on it

this is an example I worked on navy -  design team provided `200, 400, 1000`, 
image

However, existing UIs uses `500,600,700`, so I just expand (lightness and saturation) on `200,400,1000` (maintaining the `hue` - which represents the brand feel)

misicnenad:
Looks good code-wise, will gonna let Andrey make the final call on whether color palette is ok

SovereignAndrey (approved):

ap-justin:
> > To me, DEFAULT has no semantic value (other than to tell its relative position in the set) - it could have been just 500 (e.g. green-500, but used DEFAULT instead as it translates to only green when used in className, if we want, we could totally omit DEFAULT and opt for 0 as in l2,l1,0,d1,d2 to completely eliminate the misconception around its meaning ) to denote that it's the middle of the range.
> 
> @ap-justin Makes sense when put that way. Was interpreting default as "the color we most often use". Not a huge issue, that we need to worry about changing default to 0. Thanks for explaining it.
> 
> Regarding the color changes, for now, we can push and l will tell Tim, Tejaswini and Chauncey to start reviewing and getting us feedback on ASAP. I don't want to make calls for design stuff that are not needed or might deviate from the design plans they have in mind.

got this - I have included all colors in the figma file and would start using those as they appear in design. For stuff that is not included (hover, active, disabled, we'll probably use generated shades derived from given colors)

PR #2718, opened 2024-02-11
Decouple react-hook-form from `<input/>`

ap-justin:
When creating new forms which fields are not even deep in the component tree still requires [additional context wrapping](https://github.com/AngelProtocolFinance/angelprotocol-web-app/blob/2d06dedad8a404746bd3f774db8aaf25d9758cc1/src/components/donation/Steps/DonateMethods/PayPal/Form.tsx#L61) to make reusable components in (`/components/form/* `) work. 

As an example, `components/form/Field` wasn't reused in [RecipientDetailsForm](https://github.com/AngelProtocolFinance/angelprotocol-web-app/blob/2d06dedad8a404746bd3f774db8aaf25d9758cc1/src/components/BankDetails/RecipientDetails/RecipientDetailsForm/RecipientDetailsForm.tsx#L204) as form context is not provided there.

## Explanation of the solution
* extract `<NativeField/>` from `<Field/>` - but retain `<Field/>` context implementation to not affect our forms - do the same for `<CheckField/>` , `<Radio/>`, `<DateInput/>`
* reuse `<NativeField/>` in [`RecipientDetailsForm`](https://github.com/AngelProtocolFinance/angelprotocol-web-app/blob/2d06dedad8a404746bd3f774db8aaf25d9758cc1/src/components/BankDetails/RecipientDetails/RecipientDetailsForm/RecipientDetailsForm.tsx#L204) and verify it's as functional as native `<input />`

## Instructions on making this work

- run `yarn` or `yarn install` to install npm dependencies
- run `yarn run test --watchAll` to verify all tests still pass
- (optional) run `yarn run build` to verify the build passes
- run `yarn start` to start the webapp
-

## UI changes for review

When major UI changes are introduced with a PR, please include links to URLS to compare or screenshots demonstrating the difference and notify on design changes

PR #2766, opened 2024-03-05
Signup page rework

misicnenad:
Closes BG-1159
 
Borrowed a bunch of helpful logic from ser Justin's PR https://github.com/AngelProtocolFinance/angelprotocol-web-app/pull/2753

Tried making SignUp a shared component to be shared as intended in the aforementioned PR, but the differences in UI are big enough to warrant making the page a standalone component.

- go to http://localhost:4200/signup

## UI changes for review


src/pages/SignUp/SignUp.tsx
```diff
@@ -0,0 +1,33 @@
...
+      <ConfirmForm
+        userType={state.userType}
+        codeRecipientEmail={state.codeRecipientEmail}
+        setSignupState={setState}
+      />
+    );
+  })();
+
+  return (
+    <div className="grid place-items-center py-14 text-[#1D3C51]/80">
```
misicnenad, line 29: @ap-justin @SovereignAndrey not sure how to handle this color `text-[#1D3C51]/80` that appears in Figma as the default text color, but it doesn't match any color we have in *tailwind.config.js*. Left it as defined in Figma until instructed to change to a different color or to add it to the config file.

misicnenad:
@SovereignAndrey @ap-justin I have been working on finishing and polishing this the whole day and am feeling tired, so will update the error handling logic tomorrow. Wanted to create the PR so you can review the UI and logic, but please **don't merge the PR (if approved) until I update error handling** (error logging can be improved).

src/pages/SignUp/SignUp.tsx
ap-justin: its not defined into their color variables but I'll check if it fits in the gray shades they provided - If not, we could just set is as lone color e.g `gray-page`

SovereignAndrey: Probably just a missed detail from the designer. We can fit it in, as Justin mentioned, to something matching or very close that exists the existing palette.

misicnenad:
Error handling is now updated.

src/pages/SignUp/ConfirmForm.tsx
```diff
@@ -0,0 +1,117 @@
...
+      <button
+        type="submit"
+        className="flex-center bg-blue-d1 disabled:bg-gray text-white enabled:hover:bg-blue enabled:active:bg-blue-d2 h-[52px] rounded-full normal-case text-lg font-bold w-full"
+      >
+        Verify account
+      </button>
+
+      <span className="grid grid-cols-[4fr_3fr] items-center justify-items-start gap-1 text-sm font-medium mt-5">
+        <span className="justify-self-end">Resend code in</span>
+        <button
```
ap-justin, line 106: Resend inhibitor works - just not sure the numbers be the clickable 
image

would leave to internal team if this is telling enough

ap-justin (dismissed):
UI matches figma and wiring works as intended. nice work!

src/pages/SignUp/Field.tsx
```diff
@@ -0,0 +1,49 @@
...
+  name: Path<T>;
+  placeholder: string;
+  icon?: IconType;
+};
+
+export default function Field<T extends FieldValues>(props: Props<T>) {
+  const {
+    register,
+    formState: { errors },
+  } = useFormContext<T>();
```
ap-justin, line 18: we should steer away from baked in form context for new fields - as use case now doesn't need fields to be deep unlike that of crypto UIs before

misicnenad: I'm not sure I understand, can you expand a bit on what you were trying to say?

src/pages/SignUp/ConfirmForm.tsx
misicnenad: @SovereignAndrey need your input ser.

src/pages/SignUp/Field.tsx
ap-justin: existing component `components/Field` has `useFormContext` inside and is often inside `FormProvider` - this made sense before as form inputs would often placed way deep in a complex UI (crypto admin UIs)

```js
<FormProvider>
     {....}
</FormProvider>
```
but new `SignUp/Field` could be abstracted in such a way that its api is just that of native `<input/>` and can be used as

```
//no need for FormProvider
<Field {...register("password")}/>
////no need for FormProvider
```
as [demonstrated here](https://github.com/AngelProtocolFinance/angelprotocol-web-app/pull/2718/files#diff-f176c879a94c9ef8e62a3aca7830390215dcb0cba4751b9f53019b3281a58fafR198)


new forms don't have to use `FormProvider`
no need for additional generics `Field<Additional>` as `useForm.register` already has autocompletion 

```js
<Field {...register("...field name selection...")}/>
```

src/pages/SignUp/SignUp.tsx
misicnenad: `navy-l1` seems to be the closest color:
image
@ap-justin @SovereignAndrey we can use it for text instead of this Figma color.

misicnenad:
Updated the terms and policy message at the bottom of the screen (https://github.com/AngelProtocolFinance/angelprotocol-web-app/pull/2766/commits/20893785e25b4356890f9c60bd6a74e63e85c028) as per [this requirement](https://linear.app/better-giving/issue/BG-1159/signup-page#comment-2e934b5c).

src/pages/SignUp/Field.tsx
```diff
@@ -0,0 +1,49 @@
...
+          type="text"
+          className={`w-full h-full placeholder:font-medium placeholder:font-heading placeholder:text-navy-l3 focus:outline-none bg-transparent ${
+            props.icon ? "pr-5" : "px-5"
+          } ${input}`}
+          placeholder={props.placeholder}
+          aria-invalid={!!errorMsg}
+        />
+      </div>
+      {errorMsg && (
+        <p className={`text-xs text-[#C52828] mt-1.5 ${error}`}>{errorMsg}</p>
```
ap-justin, line 45: @SovereignAndrey, looked at the colors, and I think we can use the standard red they provided in place of this stray `#C52828`
image

they have very small difference in `HSL` values so I don't think we need two reds for the same purpose
image

src/pages/SignUp/SignupForm/UserTypeSelector.tsx
```diff
@@ -0,0 +1,43 @@
+import { useController, useFormContext } from "react-hook-form";
+import { FormValues } from "../types";
+
+const btnClass = (isSelected: boolean) =>
+  `flex items-center justify-center btn-outline-2 h-[42px] ${
+    isSelected
+      ? "bg-[#EAF3FA] border-[1.25px] enabled:border-blue-d1 enabled:hover:border-blue-d1 hover:cursor-default enabled:text-blue-d1 font-bold"
```
ap-justin, line 7: @SovereignAndrey, existing `blue-l4` (from standard provided) could also be used in place of this stray `#EAF3FA` as they really just look the same

image

src/pages/SignUp/ConfirmForm.tsx
SovereignAndrey: @misicnenad Will try to find other example and see how they do it and have an example to show. 

My first thought is that the numbers shouldn't be clickable, but that the "resend code" text before it would be clickable (disabled until countdown expires).

~~ OR ~~

It's all text and a "resend code" link/button appears below the text after the timer reaches zero and disappears after clicked when timer restarts.

src/pages/SignUp/SignUp.tsx
ap-justin: > `navy-l1` seems to be the closest color: image, it could fit as `blue-d5`, But I am leaning on just treating it the way it looks `navy-l1`(NavyBlue-400 in design; intended for secondary text ) as @misicnenad suggested

src/pages/SignUp/Field.tsx
misicnenad: @ap-justin you're suggesting we use `red-d3` instead, right?

src/pages/SignUp/ConfirmForm.tsx
misicnenad: @SovereignAndrey something like this?
image

src/contexts/ErrorContext/ErrorContext.tsx
```diff
@@ -10,18 +10,26 @@ import {
 } from "react";
 import { useModalContext } from "../ModalContext";
 
-type State = { handleError: (error: any, displayMessage?: string) => void };
+type State = {
+  handleError: (
+    error: any,
+    displayMessage?: string,
+    options?: { log: boolean }
```
misicnenad, line 17: @ap-justin using this `options` param can allow us to reuse the error prompt logic without logging the error (referencing the problem you raised here https://github.com/AngelProtocolFinance/angelprotocol-web-app/pull/2753#discussion_r1509096323

src/pages/SignUp/Field.tsx
misicnenad: @ap-justin what do you think about renaming this component to `Input` and moving to */components/form/\** folder as part of this PR? Later on we can refactor it to use a slightly different logic as you described.
I'll need this component for my work to rework Login page and you'll probably need it too as part of your own work.

src/App/Header/index.tsx
```diff
@@ -20,6 +20,7 @@ export default function Header({ links }: Props) {
           <div className="flex gap-2 md:gap-4 justify-self-end items-center">
             {!(
               location.pathname === appRoutes.signin ||
```
SovereignAndrey, line 22:
```suggestion
              location.pathname === appRoutes.login ||
```
Nit: didn't we settle on login & signup?

misicnenad: Will change this as part of Login rework https://linear.app/better-giving/issue/BG-1158/login-page

src/pages/SignUp/ConfirmForm.tsx
SovereignAndrey: @misicnenad This is exactly what I had in mind. :100: I'd say place the re-send countdown & link _above_ the verity button.

SovereignAndrey (approved):

src/pages/SignUp/ConfirmForm.tsx
misicnenad: image

PR #2769, opened 2024-03-07
Login page rework

misicnenad:
Closes BG-1158

- make sure you're signed out
- go to http://localhost:4200/signin

## UI changes for review
Sign in:
image

Reset password - confirm email:
image

Set new password + enter OTP (for the reason why these were merged see this explanation):
image
**NOTE: since we don't have a design for these 2 steps being merged, I created an ad-hoc combination like in the image above, but I'm open to suggestions @ap-justin @SovereignAndrey **

Reset successful:
image


SovereignAndrey (approved):
@misicnenad looks great overall. Nice work! I am torn on the ordering of the merged step fields. Password entry first, then confirm code field? :thinking: It is fine as is though.  We can let design and operations folks drive any follow-up changes there.

PR #2772, opened 2024-03-08
Prefer package.json formatting to biome formatting

ap-justin:
When running `yarn install`, package.json is auto formatted, (sorts dependencies), biome should just let this be

* added `package.json` in biome ignores

-

## UI changes for review

### DIFF
--- a/package.json
+++ b/package.json
@@ -14,7 +14,11 @@
     "preview": "rsbuild preview"
   },
   "browserslist": {
-    "production": [">0.2%", "not dead", "not op_mini all"],
+    "production": [
+      ">0.2%",
+      "not dead",
+      "not op_mini all"
+    ],
     "development": [
       "last 1 chrome version",
       "last 1 firefox version",
--- a/biome.json
+++ b/biome.json
@@ -53,7 +53,7 @@
     }
   },
   "formatter": {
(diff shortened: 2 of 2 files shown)


package.json
```diff
@@ -14,7 +14,11 @@
     "preview": "rsbuild preview"
   },
   "browserslist": {
-    "production": [">0.2%", "not dead", "not op_mini all"],
```
misicnenad, line 17: First of all - THANK YOU, this has been bugging me for some time now.

----
Problem: opening `package.json` and saving the file (with no changes) still formats this with Biome :confused: It seems the reason for this is that the Biome extension doesn't abide by the *biome.json* rules; switching the default VS Code formatter to "JSON language features" solves the problem. 
Are you encountering the same problem?

ap-justin: > Problem: opening package.json and saving the file (with no changes) still formats this with Biome 😕

🤔this doesn't occur on my side

SovereignAndrey: @misicnenad This doesn't happen on my side, so I can only assume it's something at play in your VS Code configs.

misicnenad: Interesting :thinking: Will dig in a bit when I feel bored.

misicnenad (approved):

PR #2775, opened 2024-03-11
Hide tip checkbox

0xNeshi:
Towards BG-1140

Related PR https://github.com/AngelProtocolFinance/devops/pull/509

- implemented the ticket's requirements
- removed the redundant `Divider` component and replaced it with `Separator` (previously used only in *src/components/registration*)
- separated "update profile" and "delete program" endpoints to align with new lambda logic

- sign in with an account that is an admin of some endowment
- go to endowment's donation page
- start donation process
- **verify tip screen appears** :heavy_check_mark: 
- go to the same endowment's admin edit profile page
- **verify checkbox exists** :heavy_check_mark:
- tick the checkbox
- click submit
- **verify the profile updated so that the checkbox value is now `true` even when you reload the page** :heavy_check_mark:
- go back to endowment's donation page
- start donation process
- **verify tip screen no longer appears** :heavy_check_mark:
- go to Programs page
- **verify "create program" functionality still works** :heavy_check_mark:
- go back to Programs page
- delete the newly created program
- **verify the program was actually deleted** :heavy_check_mark:

## UI changes for review

image

### DIFF
--- a/src/pages/Admin/Charity/EditProfile/HideBGTipCheckbox.tsx
+++ b/src/pages/Admin/Charity/EditProfile/HideBGTipCheckbox.tsx
@@ -9,10 +9,10 @@ export default function HideBGTipCheckbox() {
       </CheckField>
       <span className="text-xs sm:text-sm italic text-navy-l1">
         During the donation flow, there is a step in which users can choose to
-        tip us any amount they desire alongside their donation to you - the
-        amount they tip will not affect the donation amount you receive. You may
-        choose to turn this step off in the donation flow by ticking the
-        checkbox below and we will instead apply a fixed 1.5% fee to any
+        tip Better Giving any amount they desire alongside their donation to you
+        - the amount they tip will not affect the donation amount you receive.
+        You may choose to turn this step off in the donation flow by ticking the
+        checkbox above and we will instead apply a fixed 1.5% fee to any
         donation amount you receive.
       </span>
     </div>


src/components/donation/Steps/Summary/Summary.tsx
```diff
@@ -37,7 +37,11 @@ export default function Summary({
       Amount={Amount}
       amount={amount}
       splitLiq={liquidSplitPct}
-      onBack={() => dispatch(setStep("tip"))}
+      onBack={() =>
+        recipient.hide_bg_tip
+          ? dispatch(setStep("splits"))
+          : dispatch(setStep("tip"))
+      }
```
misicnenad, line 44: An alternative in this particular case would be to move this check inside the `setStep` action itself, but that would cause confusion when a dev intends to set the step to `"tip"`, but sees that the app doesn't allow him to do that. I opted for making it `setStep`'s responsibility to just set the step and not worry about which exact step to show.

src/components/donation/Steps/Summary/Summary.tsx
```diff
@@ -37,7 +37,11 @@ export default function Summary({
...
       tip={
```
misicnenad, line 45: The current setup requires this conditional logic or more generally, it requires devs to *know* what is the next and previous step for each component. 
Maybe it would be more practical to replace `setStep` with something like `goToNextStep` & `goToPreviousStep`, making those actions' responsibility to decide which step to go to, removing the need from devs to *remember* which step to navigate to. 
@ap-justin @SovereignAndrey let me know what you think.

src/pages/Admin/Charity/EditProfile/index.tsx
```diff
@@ -42,7 +42,7 @@ function FormWithContext(props: TProfile & { id: number }) {
   //
   const init: EndowmentProfileUpdate = toProfileUpdate({
     type: "initial",
-    data: { ...props, id: props.id },
+    data: { ...props },
```
misicnenad, line 45: `...props` already sets `id` correctly

src/services/aws/aws.ts
```diff
@@ -134,21 +133,28 @@ export const aws = createApi({
...
           headers: { authorization: TEMP_JWT },
           body: payload,
         };
       },
     }),
+    deleteProgram: builder.mutation<EndowmentProfile, ProgramDeleteMsg>({
+      invalidatesTags: (_, error) => (error ? [] : ["profile"]),
+      query: ({ id, program_id }) => {
+        return {
+          url: `/${v(1)}/endowments/${id}/programs/${program_id}`,
```
misicnenad, line 152: See https://github.com/AngelProtocolFinance/devops/pull/509#discussion_r1517847861

src/slices/donation/donation.ts
```diff
@@ -60,6 +60,13 @@ const donation = createSlice({
     },
 
     setSplit: (state, { payload }: PayloadAction<number>) => {
+      if (state.recipient?.hide_bg_tip) {
```
misicnenad, line 63: In reference to https://github.com/AngelProtocolFinance/angelprotocol-web-app/pull/2775/files#r1519591055, `setSplit` is already responsible for deciding on the next step, so adding the condition here is natural.

src/components/donation/Steps/Summary/Summary.tsx
ap-justin: > it requires devs to know what is the next and previous step for each component.

the implementation of 
> with something like goToNextStep & goToPreviousStep, 

still requires knowing which step to set the state

src/pages/Admin/Charity/EditProfile/Form.tsx
```diff
@@ -121,6 +122,8 @@ export default function Form() {
           label="Website of your organization"
           placeholder="https://website.org"
         />
+
+        <HideBGTipCheckbox />
```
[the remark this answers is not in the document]
SovereignAndrey: It can go here for now, as we don't as many specific design reqs for this feature. Just need to get it live for now and we can always shuttle it elsewhere based on operations/design/users feedback dictates.

SovereignAndrey:
@misicnenad Looks good to me overall! Please fix the merge conflicts and I'll take a final pass later on today. Thanks!

ap-justin (dismissed):

src/components/donation/Steps/Summary/Summary.tsx
misicnenad: > the implementation of

Not sure I understand this, did you mean to write something more here?

> still requires knowing which step to set the state

Reducer has access to `state.step`, so these actions could be written like:
```ts
    goToNextStep: (state) => {
      if (state.step === "init") {
        state.step = "donate-form";
      } else if (...) {
      ...
    },
```
Data would be set as is currently the case (`setSplit`, `setDetails`, etc.).

misicnenad:
@SovereignAndrey @ap-justin fixed the checkbox description, see https://github.com/AngelProtocolFinance/angelprotocol-web-app/pull/2775/commits/69f54a893955262435c7048767f766b91aa609e5 (`"...by ticking the checkbox [below->above]..."` & `"...users can choose to tip [us->Better Giving]..."` - let me know if you think this can be worded even better).

src/components/donation/Steps/Summary/Summary.tsx
ap-justin: >The current setup requires this conditional logic or more generally, it requires devs to know what is the next and previous step for each component.

> still requires knowing which step to set the state

In your example, dev still has to know that the next step to `init` is `donate-form`, only difference is that this is handled in the reducer 
```js
if (state.step === "init") {
        state.step = "donate-form";
      }
```
by handling this on the dispatch, caller already has context (current step)
```
//dispatcher is in splits step (context)
--> continue: tip if enabled
<-- back: donate-form

```
contrary to handling this in the reducer, additional checks required  to know what the current step first before going to the next
```js

switch(state.currentStep){
 case "form" --> split
 case "split": --> tip if not skipped
 case "tip": --> summary
 case "summary" --> checkout 
}
```
same goes for going back
```js
switch(state.currentStep){
}
```

misicnenad: > In your example, dev still has to know that the next step to init is donate-form, only difference is that this is handled in the reducer

That is true, but the difference is that the step flow is clearly defined in a single place, making it unnecessary to "remember" which step goes next/before.

> contrary to handling this in the reducer, additional checks required to know what the current step first before going to the next

This is also true, luckily with our current requirement all the necessary checks can be performed with just `state`, as it contains all the data necessary.

Anyway this was just an idea, we can table this as it's not important.

PR #2785, opened 2024-03-13
BG-1179: Adjust page paddings affected by header positioning

misicnenad:
Added `pb-20` where appropriate, as it makes distance from the bottom of the page to the top of the footer look a bit more pronounced. 

- **verify all pages' top padding is aligned with the new header**

### DIFF
--- a/src/pages/Registration/index.tsx
+++ b/src/pages/Registration/index.tsx
@@ -14,7 +14,7 @@ const Success = lazy(() => import("./Success"));
 
 function Registration() {
   return (
-    <div className="flex justify-center items-center mt-20">
+    <div className="flex justify-center items-center my-20">
       <Suspense fallback="Loading page...">
         <Seo
           title={`Registration Portal - ${APP_NAME}`}
@@ -23,24 +23,16 @@ function Registration() {
         <Routes>
           <Route
             path={regRoutes.welcome}
-            element={<Welcome classes="my-10 md:my-32 mx-6" />}
+            element={<Welcome classes="mx-6" />}
           />
           <Route
             path={regRoutes.steps + "/*"}
-            element={<Steps classes="md:my-20" />}
+            element={<Steps classes="max-sm:-my-20" />}
           />
-          <Route path={regRoutes.resume} element={<Resume classes="my-20" />} />
-          <Route
-            path={regRoutes.success}
-            element={<Success classes="mt-10 sm:mt-[5.5rem] mb-6 sm:mb-20" />}
-          />
-          <Route
-            path={regRoutes.sign_result}
-            element={
-              <SignResult classes="mt-10 sm:mt-[5.5rem] mb-6 sm:mb-20" />
-            }
-          />
-          <Route index element={<Signup classes="my-20" />} />
+          <Route path={regRoutes.resume} element={<Resume />} />
+          <Route path={regRoutes.success} element={<Success />} />
+          <Route path={regRoutes.sign_result} element={<SignResult />} />
+          <Route index element={<Signup />} />
           <Route path="*" element={<Navigate to={regRoutes.index} />} />
         </Routes>
       </Suspense>
--- a/src/pages/Registration/index.tsx
+++ b/src/pages/Registration/index.tsx
@@ -27,7 +27,7 @@ function Registration() {
           />
           <Route
             path={regRoutes.steps + "/*"}
-            element={<Steps classes="max-sm:-my-20" />}
+            element={<Steps classes="max-md:-my-20" />}
           />
           <Route path={regRoutes.resume} element={<Resume />} />
           <Route path={regRoutes.success} element={<Success />} />
--- a/src/pages/Application/Application.tsx
+++ b/src/pages/Application/Application.tsx
@@ -11,7 +11,7 @@ function Application() {
   const queryState = useApplicationQuery(id, { skip: !id });
 
   return (
-    <div className="grid content-start gap-y-4 lg:gap-y-8 lg:gap-x-3 relative padded-container pt-8 lg:pt-20 pb-8">
+    <div className="grid content-start gap-y-4 lg:gap-y-8 lg:gap-x-3 relative padded-container py-20">
       <h1 className="text-center text-3xl col-span-full max-lg:mb-4">
         Applications Review - Details
       </h1>
--- a/src/pages/Applications/Applications.tsx
+++ b/src/pages/Applications/Applications.tsx
@@ -22,7 +22,7 @@ function Applications() {
   const isLoadingOrError = isLoading || isLoadingNextPage || isError;
 
   return (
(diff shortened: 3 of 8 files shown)


src/pages/Registration/index.tsx
```diff
@@ -23,24 +23,16 @@ function Registration() {
         <Routes>
           <Route
             path={regRoutes.welcome}
-            element={<Welcome classes="my-10 md:my-32 mx-6" />}
+            element={<Welcome classes="mx-6" />}
           />
           <Route
             path={regRoutes.steps + "/*"}
-            element={<Steps classes="md:my-20" />}
+            element={<Steps classes="max-sm:-my-20" />}
```
ap-justin, line 30: too much space on mobile
image


per  the old design
image

we should leave this as is, until design got new modifications for the existing pages

ap-justin (changes requested):

src/pages/Registration/index.tsx
SovereignAndrey: @misicnenad I would say we should not leave this as it stands here, nor should we revert back to what's on staging either as changes are needed to make it workable, and rather should spend a little bit of time to adjust the mobile views paddings some more to look "decent enough" even though we don't have specific mobile guidelines. Wdyt?

ap-justin: > @misicnenad I would say we should not leave this as it stands here, nor should we revert back to what's on staging either as changes are needed to make it workable for users.

@SovereignAndrey, @misicnenad, ticket only requires to fix inconsistent top spacing due to recent header changes and I believe some parts of this PR fixed this.

But i'm not sure why suddenly we are changing bottom spacing as well for no objective reason other than it is somewhat `"approriate"` and to make to it more `"pronounced"`

> Added pb-20 where appropriate, as it makes distance from the bottom of the page to the top of the footer look a bit more pronounced.

There are no new designs to warrant changing what previous designers intended the page to be laid out.

ap-justin: @SovereignAndrey, @misicnenad, additional context for this PR:
before, Header is `transparent` allowing page with banners background to flow and be seen underneath it when not scrolled - this resulted to pages with banners to have extra top spacing that could go underneath the header

image

Now that this is no longer the case, those extra spacing should be removed.
e.g. previous
```
pt-28 pb-8 
```
could now be  just
```
py-8
```
as the extra `pt-20` is no longer needed.

misicnenad: > too much space on mobile

fixed this, see https://github.com/AngelProtocolFinance/angelprotocol-web-app/pull/2785/commits/bdac7d357af5fffeea6c49b9fc749683671fb67d

misicnenad: > @misicnenad I would say we should not leave this as it stands here, nor should we revert back to what's on staging either as changes are needed to make it workable for users. Rather we should spend a little bit of time(couple hours tops) to adjust the mobile views padding some more to look "decent enough", even though we don't have specific mobile guidelines. Wdyt?

> @SovereignAndrey, @misicnenad, ticket only requires to fix inconsistent top spacing due to recent header changes and I believe some parts of this PR fixed this.
> 
> But i'm not sure why suddenly we are changing bottom spacing as well for no objective reason other than it is somewhat "approriate" and to make to it more "pronounced"

@SovereignAndrey @ap-justin this is once again a discussion about how strict/focused we want our PRs to be with regards to ticket requirements. We've had similar discussions before (we've had one 2 weeks ago, see https://github.com/AngelProtocolFinance/angelprotocol-web-app/pull/2744) and they seem to be a recurrent cause of conflict. 
What makes things even more confusing is that positions on this topic seem to change from PR to PR and I guess this is the result of _not having a clear policy on the subject_, or in other other words _we don't have an answer to the question: "do we strictly respect the ticket requirements and leave unrelated changes to follow-up PRs/tickets OR do we allow ad hoc changes?"_.
I advocated the former until the PR linked above (https://github.com/AngelProtocolFinance/angelprotocol-web-app/pull/2744) and opted for the latter in all my PRs since. Although I like arguing with you guys (joke ofc), this topic is becoming a bit boring and confusing, so we really need to have a clear answer to the above question - strict PRs or not?

ap-justin: @misicnenad, @SovereignAndrey, 
> I advocated the former until the PR linked above (https://github.com/AngelProtocolFinance/angelprotocol-web-app/pull/2744) and opted for the latter in all my PRs since.

I believe we are confusing and mixing terms like unrelated/surface-area/unnecessary. I hope I could break it down to you.

First to focus on this PR:
`About`: header changes made top extra paddings unneccessary causing too much spacing between. this PR should opt to remove that.
`Changes`:  Instead of removing those extra top spacing, more spaces are added in the bottom instead citing this as "approriate" and "pronouncing". This to me is `UNNECESSARY` and shoudn't be done and is nowhere `RELATED` to the problem being fixed

I'll write about #2744 shortly

ap-justin: With regards to #2744, 
`About:` User complains num chars validation is off, citing video where can't submit even though haven't reached the char limit
`Changes`:
1. Move validation from callback `onError` to standalone schema `richText`.
2. `RELATED` to validation functionality of said schema, found that it allows submission even though character limit is reached. 
3. `RELATED` to validation functionality of said schema, found that when user breaches limit ( errors out ) and goes back within limits, error is not removed 

turned out that related bugs `2,3` could be fixed by just going back to `1` and removing the old `onError` in favor of the new `richText`. But @misicnenad repeatedly refused to do attributing to `UNRELATED` and increases `SURFACE AREA`

ap-justin: @SovereignAndrey @misicnenad , to add to these, 

> loose PRs -> leave things more or less as-is, updating paddings where necessary to improve UI

we should not take it to ourselves to judge weather unplanned changes we make would `"improve"` UI especially if such UI is backed by design works. 

An example to this is the designed UX interactivity of registration progress indicator. It just works just fine, [until for some unnecessary reason we somehow subjectively felt it's off](https://linear.app/better-giving/issue/BG-987/registration-when-contact-details-step-is-not-complete-its-completed). 

Which instead of doing any good, brought more bugs to fix and discuss upon
- https://linear.app/better-giving/issue/BG-1122/staging-registration-marker-not-filled-even-when-step-is-complete
- https://linear.app/better-giving/issue/BG-1188/prod-and-staging-registration-progress-indicator-not-reverting-to

which doesn't exist in the first place

SovereignAndrey (approved):
This was a simple, needed change driven by other changes that had been made in other recent PRs to page header/nav fixes. Way too much back and forth and overall time sunk on what ought to have been a trivial PR review w/ a change suggestion(s), that suggestion(s) alone debated and possibly pulled in and we'd be set. This in NOT the place for discussing working processes/etc which should be taken up in Discord chat.

Approving this and closing.

PR #2788, opened 2024-03-14
BG-1188: registration progress indicator must revert to desktop position

misicnenad:

- perform repro steps (see ticket)
- **verify the issue no longer appears**

### DIFF
--- a/src/pages/Registration/Steps/ProgressIndicator.tsx
+++ b/src/pages/Registration/Steps/ProgressIndicator.tsx
@@ -32,6 +32,7 @@ export default function ProgressIndicator({ step, classes = "" }: Props) {
 
       if (ref.isDesktop !== isOnDesktop) {
         setDesktop(isOnDesktop);
+        ref.isDesktop = isOnDesktop;
       }
     },
     {
--- a/src/pages/Registration/Steps/ProgressIndicator.tsx
+++ b/src/pages/Registration/Steps/ProgressIndicator.tsx
@@ -25,7 +25,7 @@ export default function ProgressIndicator({ step, classes = "" }: Props) {
   useHandleScreenResize(
     (screen, ref) => {
       const isOnDesktop = screen >= SCREEN_BREAKPOINTS.md;
-      if ((isOnDesktop && !ref.isOpen) || (!isOnDesktop && ref.isOpen)) {
+      if (isOnDesktop !== ref.isOpen) {
         setIsOtherStepsShown(isOnDesktop);
         ref.isOpen = isOnDesktop;
       }


src/pages/Registration/Steps/ProgressIndicator.tsx
```diff
@@ -25,13 +25,14 @@ export default function ProgressIndicator({ step, classes = "" }: Props) {
   useHandleScreenResize(
     (screen, ref) => {
       const isOnDesktop = screen >= SCREEN_BREAKPOINTS.md;
-      if ((isOnDesktop && !ref.isOpen) || (!isOnDesktop && ref.isOpen)) {
```
ap-justin, line 28: @misicnenad, there's still noticeable layout shift (progress indicator lags in the middle then jumps to the top) present here that's not present [in how the progress indicator is originally developed](https://github.com/AngelProtocolFinance/angelprotocol-web-app/pull/2789).

misicnenad: @ap-justin @SovereignAndrey I've been thinking about this, do we need to spend time fixing this layout shift lag at all?
The lag only appears when a user is manually resizing their screen or switching from mobile to desktop view using the browser console, so a question arises: how often really will users be manually resizing their screens?
I'm not sure the problem is worth the time necessary to debug and fix it.

ap-justin: > The lag only appears when a user is manually resizing their screen or switching from mobile to desktop view using the browser console, so a question arises: how often really will users be manually resizing their screens and do so during registration?

Just to be entirely clear: 

`A.` pertaining to the bug reported in the ticket ( a UI edge case ) - a detail considered when this UI is first implemented and a bug not present before changes in ProgressIndicator was done (again we are talking about a bug, not styling nor a change in general)

`B.` pertaining to the (lag) - a flaw in the fix being implemented to solve  `A.`

`B` could be ignored if you guys want, but `A` is an edge case that's been handled before and should be kept handled.

misicnenad: To be fair the problem wasn't there before because the progress indicator was stuck in the mobile position.

misicnenad: Ah, yes, maybe should've been more specific. I was talking about `B`.

misicnenad: @SovereignAndrey pinging you ser about https://github.com/AngelProtocolFinance/angelprotocol-web-app/pull/2788#discussion_r1532012344

SovereignAndrey: We'll ignore the transitional artifacts that arise when changing screen size changes. This is not a case that will affect 99% of our target users, as few will be fiddling with the sizes mid visit I expect. If we get complaints reported, then we'll spend more time on refining this, but right now we got bigger fish to fry.

SovereignAndrey (approved):

PR #2791, opened 2024-03-18
BG-1190 & BG-1191 Sentry fixes: env & pkg ver

SovereignAndrey:
- Use our env var (`PUBLIC_ENVIRONMENT`) to set Sentry's `environment` on init call
- Downgrades Sentry to production version (`v7.89.0`)
- Only log errors to Sentry for Production environment

## UI changes for review
None

### DIFF
--- a/src/index.tsx
+++ b/src/index.tsx
@@ -23,6 +23,7 @@ const root = createRoot(container as Element);
 
 Sentry.init({
   dsn: process.env.PUBLIC_SENTRY_DSN,
+  environment: process.env.PUBLIC_ENVIRONMENT,
 });
 
 root.render(
--- a/package.json
+++ b/package.json
@@ -41,7 +41,7 @@
     "@paypal/react-paypal-js": "8.1.3",
     "@radix-ui/react-slider": "1.1.2",
     "@reduxjs/toolkit": "2.2.1",
-    "@sentry/react": "7.106.1",
+    "@sentry/react": "7.89.0",
     "@stripe/react-stripe-js": "2.4.0",
     "@stripe/stripe-js": "2.4.0",
     "@terra-money/terra.js": "3.1.10",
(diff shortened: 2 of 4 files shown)


src/index.tsx
```diff
@@ -23,6 +23,7 @@ const root = createRoot(container as Element);
 
 Sentry.init({
   dsn: process.env.PUBLIC_SENTRY_DSN,
+  environment: process.env.PUBLIC_ENVIRONMENT,
```
ap-justin, line 26: @SovereignAndrey, to add to `BG-1191:staging: sentry error should only be captured in production`
in `helpers/logger.ts`

`logger.error` won't log in production (`IS_TEST = false`)
```js
export const logger: Logger = IS_TEST
  ? {
      // using console.log for .error and .info funcs as console.error and
      // console.info are undefined while testing for some reason
      error: (message) => {
        Sentry.captureException(message);
        console.log(message);
      },
      info: (...data) => console.log(...data),
      log: (...data) => console.log(...data),
    }
  : {
      error: (_) => {},
      info: (_) => {},
      log: (_) => {},
    };

```
should be changed to 
```js
export const logger: Logger = IS_TEST
  ? {
      // using console.log for .error and .info funcs as console.error and
      // console.info are undefined while testing for some reason
      error: (message) => {
        console.log(message);
      },
      info: (...data) => console.log(...data),
      log: (...data) => console.log(...data),
    }
  : {
      error: (_) => {
         Sentry.captureException(message); //move sentry capture here
      },
      info: (_) => {},
      log: (_) => {},
    };

```

SovereignAndrey: @ap-justin Yeah, that's true with regards to flooding sentry with staging/dev errors. Hmm... :thinking: 
I guess we're safe to capture _only_ production level changes in Sentry. We already have dev logging to debug local issues & staging app bugs that Operations catches can always be triaged internally. Will make those needed changes.

ap-justin (approved):

PR #2797, opened 2024-03-20
Closes BG-1193: Endowment slug

ap-justin:
* create slug field 
image
* update endpoint version

-

## UI changes for review

### DIFF
--- a/src/pages/Admin/Charity/EditProfile/Form.tsx
+++ b/src/pages/Admin/Charity/EditProfile/Form.tsx
@@ -14,6 +14,7 @@ import { Link } from "react-router-dom";
 import { EndowDesignation } from "types/aws";
 import { UNSDG_NUMS } from "types/lists";
 import HideBGTipCheckbox from "./HideBGTipCheckbox";
+import Slug from "./Slug";
 import { getSDGLabelValuePair } from "./getSDGLabelValuePair";
 import { MAX_CHARS, MAX_SIZE_IN_BYTES, VALID_MIME_TYPES } from "./schema";
 import { FV } from "./types";
@@ -127,12 +128,7 @@ export default function Form() {
 
         <HideBGTipCheckbox />
 
-        <Field<FV>
-          classes="field-admin"
-          name="slug"
-          label="Slug"
-          placeholder="your-endowment-slug"
-        />
+        <Slug />
       </Group>
 
       <Group title="Organization">
--- a/src/pages/Admin/Charity/EditProfile/Form.tsx
+++ b/src/pages/Admin/Charity/EditProfile/Form.tsx
@@ -126,6 +126,13 @@ export default function Form() {
         />
 
         <HideBGTipCheckbox />
+
+        <Field<FV>
+          classes="field-admin"
+          name="slug"
+          label="Slug"
+          placeholder="your-endowment-slug"
+        />
       </Group>
 
       <Group title="Organization">
--- a/src/pages/Admin/Charity/EditProfile/HideBGTipCheckbox.tsx
+++ b/src/pages/Admin/Charity/EditProfile/HideBGTipCheckbox.tsx
@@ -7,7 +7,7 @@ export default function HideBGTipCheckbox() {
       <CheckField<FV> name="hide_bg_tip" classes="text-sm">
         Opt out of tip-based donations
       </CheckField>
-      <span className="text-xs sm:text-sm text-navy-l1">
+      <span className="text-xs sm:text-sm italic text-navy-l1">
         During the donation flow, there is a step in which users can choose to
         tip Better Giving any amount they desire alongside their donation to you
         - the amount they tip will not affect the donation amount you receive.
--- a/src/pages/Admin/Charity/EditProfile/HideBGTipCheckbox.tsx
+++ b/src/pages/Admin/Charity/EditProfile/HideBGTipCheckbox.tsx
@@ -7,7 +7,7 @@ export default function HideBGTipCheckbox() {
       <CheckField<FV> name="hide_bg_tip" classes="text-sm">
         Opt out of tip-based donations
       </CheckField>
-      <span className="text-xs sm:text-sm italic text-navy-l1">
+      <span className="text-xs sm:text-sm text-navy-l1">
         During the donation flow, there is a step in which users can choose to
         tip Better Giving any amount they desire alongside their donation to you
         - the amount they tip will not affect the donation amount you receive.
--- a/src/pages/Admin/Charity/EditProfile/schema.ts
+++ b/src/pages/Admin/Charity/EditProfile/schema.ts
@@ -48,5 +48,5 @@ export const schema = object<any, SchemaShape<FV>>({
     tiktok: url,
   }),
   overview: richTextContent({ maxChars: MAX_CHARS }),
-  slug: string().matches(/\D/, "can't use id as slug"),
+  slug: string().trim().matches(/\D|^$/, "can't use id as slug"),
 }) as ObjectSchema<FV>;
--- a/src/pages/Admin/Charity/EditProfile/schema.ts
+++ b/src/pages/Admin/Charity/EditProfile/schema.ts
@@ -48,5 +48,8 @@ export const schema = object<any, SchemaShape<FV>>({
     tiktok: url,
   }),
   overview: richTextContent({ maxChars: MAX_CHARS }),
-  slug: string().trim().matches(/\D|^$/, "can't use id as slug"),
+  slug: string()
+    .trim()
+    .max(30, "max 30 characters")
+    .matches(/\D|^$/, "can't use id as slug"),
 }) as ObjectSchema<FV>;
--- a/src/pages/Admin/Charity/EditProfile/schema.ts
+++ b/src/pages/Admin/Charity/EditProfile/schema.ts
@@ -5,7 +5,7 @@ import { optionType, richTextContent } from "schemas/shape";
 import { url, requiredString } from "schemas/string";
 import { SchemaShape } from "schemas/types";
 import { ImageMIMEType } from "types/lists";
-import { ObjectSchema, array, object } from "yup";
+import { ObjectSchema, array, object, string } from "yup";
 import { FV } from "./types";
 
 export const VALID_MIME_TYPES: ImageMIMEType[] = [
@@ -48,4 +48,5 @@ export const schema = object<any, SchemaShape<FV>>({
     tiktok: url,
   }),
   overview: richTextContent({ maxChars: MAX_CHARS }),
+  slug: string().matches(/\D/, "can't use id as slug"),
 }) as ObjectSchema<FV>;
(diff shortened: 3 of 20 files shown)


src/types/aws/ap/index.ts
```diff
@@ -88,7 +89,7 @@ export type EndowmentOption = Pick<Endowment, "id" | "name">;
 
 //most are optional except id, but typed as required to force setting of default values - "", [], etc ..
 export type EndowmentProfileUpdate = Except<
-  Ensure<Endowment, "hide_bg_tip">,
+  Ensure<Endowment, "hide_bg_tip" | "slug" | "card_img">,
```
ap-justin, line 92:
```suggestion
  Ensure<Endowment, "hide_bg_tip" | "slug">,
```
unrelated to problem: `card_img` attribute is optional attribute of `type Endowment` which is not enforced in `type EndowmentProfileUpdate` - so not initialized and creates false diff (makes edit profile submit even nothing is changed - technically something changed under the hood `initial.card_image` is `undefined` while `fv.card_image` is initialized to `""`, change: `undefined` -> `""`)

A more sustainable definition would be `Required<Endowment>` so not to missing additing in `Ensure` when new optional attribute is added

src/pages/Marketplace/index.tsx
```diff
@@ -1,9 +1,16 @@
...
 import ActiveFilters from "./ActiveFilters";
 import Cards from "./Cards";
 import Hero from "./Hero";
 import Toolbar from "./Toolbar";
 
 export default function Marketplace() {
+  const [params] = useSearchParams();
+  const slug = params.get("slug");
+
+  if (slug) return <Profile slug={slug} />;
```
ap-justin, line 12: `marketplace?params..` is the same route as `marketplace` so distinguishing this needs to be inside Marketplace where search params can be parsed

src/pages/Admin/Charity/EditProfile/Form.tsx
```diff
@@ -126,6 +126,13 @@ export default function Form() {
         />
 
         <HideBGTipCheckbox />
+
+        <Field<FV>
+          classes="field-admin"
+          name="slug"
+          label="Slug"
```
misicnenad, line 133: @SovereignAndrey @ap-justin we should add a bit more detail about the purpose of "Slug", as not all users know what it even is.

src/pages/Admin/Charity/EditProfile/schema.ts
```diff
@@ -48,4 +48,5 @@ export const schema = object<any, SchemaShape<FV>>({
     tiktok: url,
   }),
   overview: richTextContent({ maxChars: MAX_CHARS }),
+  slug: string().trim().matches(/\D|^$/, "can't use id as slug"),
```
misicnenad, line 51: Not sure what the backend logic will be, but do we want to allow users to be able to remove their slug once it has been set?
`user has no slug -> user sets some slug like "some-slug" -> user updates profile to remove the slug (sets to empty string)`

src/pages/Admin/Charity/EditProfile/useEditProfile.ts
```diff
@@ -51,6 +53,17 @@ export default function useEditProfile() {
         return showModal(TxPrompt, { error: "No changes detected" });
       }
 
+      if (update.slug !== initial.slug) {
+        const result = await endowment({ id: update.slug });
```
misicnenad, line 57: Important: just a note that the same check must be performed in the backend as well (`update-endow-profile`), as it's possible some (malicious) actor sends a direct curl request to update profile to our lambda.

src/services/aws/aws.ts
```diff
@@ -117,13 +117,14 @@ export const aws = createApi({
     }),
     endowment: builder.query<
       Endowment,
-      { id: number; fields?: (keyof Endowment)[] }
+      { id: number | string; fields?: (keyof Endowment)[] }
```
misicnenad, line 120: This setup allows a small chance that id would be passed as a string by accident, but treated like a slug (`useEndowomentQuery("1")`), which would result in unintended behavior (although valid, it would return "endowment not found").

Suggestion: accept *either* `id` or `slug` _explicitly_:
```ts
    endowment: builder.query<
      Endowment,
      ({ id: number; slug?: never } | { id?: never; slug: string }) & {
        fields?: (keyof Endowment)[];
      }
    >({
      providesTags: ["profile"],
      query: ({ fields, id, slug }) => ({
        url: !!id ? `v7/endowments/${id}` : "v7/endowments", // might even write just `v7/endowments/${id}`
        params: {
          env: apiEnv,
          slug, // if `id` is passed, then this is undefined and will not be set
          ...(fields ? { fields: fields.join(",") } : {}),
        },
      }),
    }),
```
and update `useEndowment` wrapper to something similar:
```ts
export function useEndowment<T extends K[]>(
  idOrSlug: { id: number; slug?: never } | { id?: never; slug: string },
  fields?: T,
  options?: { skip: boolean }
) {
  const query = useEndowmentQuery(
    {
      ...idOrSlug, // pass on whatever was set
      fields,
    },
    options
  );

  return query as QueryState<
    T extends K[] ? Pick<Endowment, ArrayValues<T>> : Endowment
  >;
}
```

misicnenad (changes requested):

src/pages/Admin/Charity/EditProfile/useEditProfile.ts
ap-justin: you could test that this is already the case, remove this check and submit an existing slug

src/pages/Admin/Charity/EditProfile/schema.ts
ap-justin: user setting slug to `""` is simply treated as endow doesn't have slug at all. If no slug is passed on `GET` endow isn't find by `slug`.

src/services/aws/aws.ts
ap-justin: fixed: https://github.com/AngelProtocolFinance/angelprotocol-web-app/pull/2797/commits/3284a74febdbda19c9885f2a39e416bfa93985fd

src/pages/Admin/Charity/EditProfile/Form.tsx
ap-justin: chose to defer the verbiage to marketing - I should have put this in some comment - Unless you really want this in this PR @SovereignAndrey

src/pages/Admin/Charity/EditProfile/useEditProfile.ts
misicnenad: Ah, missed the changes that introduces the slug in the first place, so all is good

misicnenad (commented):
Tentative approval, waiting on Andrey to respond to https://github.com/AngelProtocolFinance/angelprotocol-web-app/pull/2797#discussion_r1531941673

src/pages/Admin/Charity/EditProfile/Form.tsx
SovereignAndrey: @ap-justin I would agree we can change that a bit to be more understandable(at least an attempt at it) within this PR. This was my bad. I forgot to add the UI facing wording for that field/help text. :thinking: 

For the UI facing text, let's use the following for now. Marketing can always chime in with teaks, if needed, to the copy:
- [ ] Field Label: `Profile "Vanity" URL`
- [ ] Help Tex below label: `Set your nonprofit's profile URL ending segment from the default numeric ID to a more human-readable value for better SEO and donor recognition. <strong>Only numbers and letters are permitted!</strong> Your numeric ID URL will continue to work still, so there is no risk of breaking existing links that have been shared.</br>Example: Setting this field as "myNonprofit" would change your nonprofit's profile URL to {{baseURL}}/marketplace/myNonprofit.`

SovereignAndrey (changes requested):
Please see change item here: https://github.com/AngelProtocolFinance/angelprotocol-web-app/pull/2797#discussion_r1533127459

Also, we'll want to think about char limits for this. Maybe something like 30 chars max? Enough for some semblance of most charity names, but not enough for an essay. Wdyt?
- [ ] Char limit for `slug` field set to 30 or less

src/pages/Admin/Charity/EditProfile/Form.tsx
ap-justin: @SovereignAndrey, AWS were setup for 
```
{{baseURL}}/marketplace?slug=myNonProfit
```
should I change this so that slug is treated as path parameter instead of search param?

src/pages/Admin/Charity/EditProfile/HideBGTipCheckbox.tsx
```diff
@@ -7,7 +7,7 @@ export default function HideBGTipCheckbox() {
       <CheckField<FV> name="hide_bg_tip" classes="text-sm">
         Opt out of tip-based donations
       </CheckField>
-      <span className="text-xs sm:text-sm italic text-navy-l1">
+      <span className="text-xs sm:text-sm text-navy-l1">
```
ap-justin, line 10:
```suggestion
      <span className="text-xs sm:text-sm italic text-navy-l1">
```
slug description contains bold words which won't show under italic - so slug desscription is not italic. Made tip help text also not italic for uniformity

src/pages/Admin/Charity/EditProfile/Form.tsx
SovereignAndrey: @ap-justin I think it's fine if the web app takes either, in the cleaner /marketplace/{idOrSlug} setup. :+1: Web App can check if `slugOrId` is an integer. If we have an integer, use the ID lookup method AWS, else use the slug query method: `parseInt(slugOrId) ? endowments/{slugOrId} : endowments?slug={slugOrId}`

ap-justin: @SovereignAndrey 
> I think it's fine if the web app takes either, 

just to confim: we should allow web-app urls: 
```
marketplace?slug={slug}
marketplace/id
marketplace/slug
```

and in the field description 
`...would create a new vanity profile URL of {{baseURL}}/marketplace/myNonprofit.`

SovereignAndrey: @ap-justin I don't think there's a need for the first option: `marketplace?slug={slug}`.
Description looks OK.

src/pages/Admin/Charity/EditProfile/HideBGTipCheckbox.tsx
SovereignAndrey: I've been staring at the new slug help text for a bit and I _think_ that we should prob:
- [ ] drop the bold text on slug helptext/description
- [ ] revert to using italic

It just seems too similar to the labels if we remove the italic. Wdyt?

ap-justin: > It just seems too similar to the labels if we remove the italic. Wdyt?

To me smaller and lighter helper text is enough to distinguish it from the labels. Nonetheless I've applied to changes.

image

src/pages/Admin/Charity/EditProfile/Form.tsx
ap-justin: > @ap-justin I don't think there's a need for the first option: `marketplace?slug={slug}`. Description URL looks OK.

removed route `marketplace?slug={slug}`

SovereignAndrey (approved):

PR #2816, opened 2024-03-27
BG-1218 Donation splits slider: swap display ordering

SovereignAndrey:
Display needed to be swapped for split component ordering to avoid confusion over colors and %s. 
image

- Go to any nonprofit donation page
- Verify the splits work as desired still. Only the display order has changed

## UI changes for review
image

### DIFF
--- a/src/components/donation/Steps/Splits/Splits.tsx
+++ b/src/components/donation/Steps/Splits/Splits.tsx
@@ -50,12 +50,12 @@ export default function Split({
 
       {/** percentages */}
       <div className="flex justify-between mt-6 mb-3">
-        <div className="grid gap-1 grid-cols-[1fr_auto] items-center">
+        <div className="grid gap-1 grid-cols-[auto_1fr] items-center">
           <p className="text-xl text-right">{liqSplitPct}%</p>
           <Image src={sendMoney} className="ml-1" />
           <p className="uppercase text-xs col-span-full">Instantly Available</p>
         </div>
-        <div className="grid gap-1 grid-cols-[auto_1fr] items-center">
+        <div className="grid gap-1 grid-cols-[1fr_auto] items-center">
           <Image src={leaf} className="mr-1" />
           <p className="text-xl">{lockedSplitPct}%</p>
           <p className="uppercase text-xs col-span-full">Sustainable fund</p>
--- a/src/components/donation/Steps/Splits/Splits.tsx
+++ b/src/components/donation/Steps/Splits/Splits.tsx
@@ -50,15 +50,15 @@ export default function Split({
 
       {/** percentages */}
       <div className="flex justify-between mt-6 mb-3">
-        <div className="grid gap-1 grid-cols-[auto_1fr] items-center">
-          <Image src={leaf} className="mr-1" />
-          <p className="text-xl">{lockedSplitPct}%</p>
-          <p className="uppercase text-xs col-span-full">Sustainable fund</p>
-        </div>
         <div className="grid gap-1 grid-cols-[1fr_auto] items-center">
           <p className="text-xl text-right">{liqSplitPct}%</p>
           <Image src={sendMoney} className="ml-1" />
           <p className="uppercase text-xs col-span-full">Instantly Available</p>
+        </div>
+        <div className="grid gap-1 grid-cols-[auto_1fr] items-center">
+          <Image src={leaf} className="mr-1" />
+          <p className="text-xl">{lockedSplitPct}%</p>
+          <p className="uppercase text-xs col-span-full">Sustainable fund</p>
         </div>
       </div>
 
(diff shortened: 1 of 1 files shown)


src/components/donation/Steps/Splits/Splits.tsx
```diff
@@ -51,15 +51,15 @@ export default function Split({
...
-          <Image src={leaf} className="mr-1" />
-          <p className="text-xl">{lockedSplitPct}%</p>
-          <p className="uppercase text-xs col-span-full">Sustainable fund</p>
-        </div>
-        <div className="grid gap-1 grid-cols-[1fr_auto] items-center">
           <p className="text-xl text-right">{liqSplitPct}%</p>
           <Image src={sendMoney} className="ml-1" />
           <p className="uppercase text-xs col-span-full">Instantly Available</p>
         </div>
+        <div className="grid gap-1 grid-cols-[1fr_auto] items-center">
```
ap-justin, line 58:
```suggestion
        <div className="grid gap-1 grid-cols-[auto_1fr] items-center">
```
and the other container (liquid) be `[1fr_auto]`

request change: to get the intended middle alignment
image

comment: personally, alignment looks off to me though, prefers everythng aligned against the container
image

ap-justin (changes requested):

src/components/donation/Steps/Splits/Splits.tsx
SovereignAndrey: @ap-justin Good suggestion. I will run this idea by Tejaswini. Can always make this change separately though in new tiny ticket, as it's unrelated to the fix at hand (ie. swap the labels).

ap-justin: @SovereignAndrey, requested changes to fix icon alignment 

> request change: to get the intended middle alignment

image

```suggestion
        <div className="grid gap-1 grid-cols-[auto_1fr] items-center">
```
> and the other container (liquid) be `[1fr_auto]`

ap-justin (changes requested):

src/components/donation/Steps/Splits/Splits.tsx
SovereignAndrey: Sorry, closed that too quickly w/o looking closely enough at the fix and only the images and second block of text about the container alignment. :+1:

SovereignAndrey: @ap-justin Fixed by reverting col grid change (see: [ec7a9d5](https://github.com/AngelProtocolFinance/angelprotocol-web-app/pull/2816/commits/ec7a9d516734f5f6699f35a91f30b6b7aaf0f99b))

src/components/donation/Steps/Splits/Splits.tsx
```diff
@@ -50,16 +50,16 @@ export default function Split({
...
           <p className="uppercase text-xs col-span-full">Instantly Available</p>
         </div>
+        <div className="grid gap-1 grid-cols-[auto_1fr] items-center">
+          <Image src={leaf} className="mr-1" />
+          <p className="text-xl">{lockedSplitPct}%</p>
+          <p className="uppercase text-xs col-span-full">Sustainable fund</p>
+        </div>
       </div>
 
       {/** slider */}
```
misicnenad, line 65: There's a bug - the slider is increasing, but the displayed liquid value is decreasing:
image

The problem is that the slider is still tracking the locked value instead of the liquid one.
- update local state to `const [liquidSplitPct, setLiquidSplitPct] = useState(persistedLiqSplit);`
- remove `lockedSplitPct` (to display the value above the slider simply use `100 - liquidSplitPct`)
- update `Slider.Root` to:
```ts
      <Slider.Root
        value={[liquidSplitPct]}
        onValueChange={([pct]) => setLiquidSplitPct(pct)}
        className="relative flex items-center select-none touch-none my-2"
      >
```

misicnenad (changes requested):

src/components/donation/Steps/Splits/Splits.tsx
SovereignAndrey: @misicnenad When the slider moves to the right, the user is increasing the locked % amount desired, so it follows that there _should_ be a complementary decrease the liquid % and liquid amount values. I don't think there's a need to adjust that logic. We've only switched the display order, so that the slider color matches the color of the icon it ends up underneath at the end of either range. Lmk if that tracks, thanks!

ap-justin (approved):

src/components/donation/Steps/Splits/Splits.tsx
misicnenad: @SovereignAndrey I think donators would find this confusing - the left-hand side of the slider should represent whatever % is displayed _on the left_. As a donator I'd expect the left % match the size of the green (left) part of the slider.

IMO the current setup is the same as if the user hit "Page Up" key on their keyboard, but their screen scrolled *down*.

PR #2823, opened 2024-04-01
Dynamically set PayPal button's currency

jp-mariano:
Replaced the hard coded `"USD"` when initializing the PayPal button with `details.currency.code.toUpperCase()` so that both the initial order creation and button initialization processes use the same currency.

- Create a single fiat donation for USD & a non-USD currency
- Verify that PayPal Express checkout button does not throw an error and opens the pop-up window for checkout

## UI changes for review
No major UI changes.


misicnenad (approved):

SovereignAndrey (approved):
thanks for the fix! 🙂

PR #2839, opened 2024-04-03
BG-1213: URI validation bug: should encode all URIs before submitting to AWS to be RFC-compliant

misicnenad:

- **verify issue no longer appears**

### DIFF
--- a/src/helpers/uploadFiles.ts
+++ b/src/helpers/uploadFiles.ts
@@ -6,6 +6,7 @@ export type Bucket = "endow-profiles" | "endow-reg";
 export const bucketURL = "s3.amazonaws.com";
 
 const SPACES = /\s+/g;
+
 export async function uploadFiles(
   files: File[],
   bucket: Bucket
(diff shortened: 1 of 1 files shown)


src/helpers/uploadFiles.ts
```diff
@@ -36,7 +37,7 @@ export async function uploadFiles(
 }
 
 export function getFullURL(baseURL: string, fileName: string) {
-  return `${baseURL}-${fileName.replace(SPACES, "_")}`;
+  return encodeURI(`${baseURL}-${fileName.replace(SPACES, "_")}`);
```
misicnenad, line 40: As far as I could gather, every `publicUrl` field, which we use to submit files to AWS, gets constructed with this `getFullURL` function, thus the safest way to ensure all URLs are encoded prior to submission is to do it here.

The alternative would be to have a convention where we *must remember* to call `encodeURI` before any submission, which we are sure to forget at some point.

Open to suggestions anyway :+1:

SovereignAndrey (approved):

PR #2872, opened 2024-04-11
Slug preview

ap-justin:
* add slug preview
image

-

## UI changes for review

### DIFF
--- a/src/pages/Admin/Charity/EditProfile/Slug.tsx
+++ b/src/pages/Admin/Charity/EditProfile/Slug.tsx
@@ -1,8 +1,13 @@
 import { Field } from "components/form";
 import { BASE_URL } from "constants/env";
+import { appRoutes } from "constants/routes";
+import { Link } from "react-router-dom";
 import { FV } from "./types";
 
-export default function Slug() {
+export default function Slug({ initSlug = "" }) {
+  const link =
+    initSlug && `${window.origin}${appRoutes.marketplace}/${initSlug}`;
+
   return (
     <div>
       <Field<FV>
@@ -11,7 +16,15 @@ export default function Slug() {
         label="Custom Profile URL"
         placeholder="myNonprofit"
       />
-      <p className="text-xs sm:text-sm text-navy-l1 italic mt-4">
+      {link && (
+        <div className="text-sm italic mt-1">
+          <span>Current URL:</span>{" "}
+          <Link to={link} className="text-blue-d1 hover:underline">
+            {link}
+          </Link>
+        </div>
+      )}
+      <p className="text-xs sm:text-sm text-navy-l1 italic mt-2">
         Change your nonprofit's profile URL ending segment from the default
         numeric ID to a more human-readable value for better SEO and donor
         recognition. Only numbers and letters are permitted! Your numeric ID URL
--- a/src/pages/Admin/Charity/EditProfile/Slug.tsx
+++ b/src/pages/Admin/Charity/EditProfile/Slug.tsx
@@ -1,12 +1,17 @@
 import { Field } from "components/form";
-import { BASE_URL } from "constants/env";
 import { appRoutes } from "constants/routes";
+import { useFormContext } from "react-hook-form";
 import { Link } from "react-router-dom";
 import { FV } from "./types";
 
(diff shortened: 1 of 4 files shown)


src/pages/Admin/Charity/EditProfile/Slug.tsx
```diff
@@ -1,8 +1,13 @@
 import { Field } from "components/form";
 import { BASE_URL } from "constants/env";
+import { appRoutes } from "constants/routes";
+import { Link } from "react-router-dom";
 import { FV } from "./types";
 
-export default function Slug() {
+export default function Slug({ initSlug = "" }) {
+  const link =
```
misicnenad, line 8: Once I saved a slug, I was not able to delete it, got error:
`"One or more parameter values are not valid. The update expression attempted to update a secondary index key to a value that is not supported. The AttributeValue for a key attribute cannot contain an empty string value."`

Is this expected?

src/pages/Admin/Charity/EditProfile/Slug.tsx
```diff
@@ -1,8 +1,13 @@
...
+    initSlug && `${window.origin}${appRoutes.marketplace}/${initSlug}`;
```
misicnenad, line 9: Judging by the wording of the Linear ticket, I think the requirement was to show what the slug would look like as the input value is changed, examples below:
*user starts typing*
slug: `e`
Current URL: `http://localhost:4200/marketplace/e`
slug: `ex`
Current URL: `http://localhost:4200/marketplace/ex`
slug: `exa`
Current URL: `http://localhost:4200/marketplace/exa`
slug: `example`
Current URL: `http://localhost:4200/marketplace/example`

I.e. show in the URL whatever is in the slug input field.

Wasn't this the idea?

misicnenad (changes requested):

src/pages/Admin/Charity/EditProfile/Slug.tsx
ap-justin: this is bug in the edit profile handler, I have set `slug` as pk in a GSI, perphaps instead of setting to `""`, attribute slug should be removed when web-app sets it to `""`

src/pages/Admin/Charity/EditProfile/Slug.tsx
SovereignAndrey: @ap-justin @misicnenad Ah, english... nebulous, unlike computer code 😆 In my head, I thought of it as applying to submitted, existing slugs only:
- IF `slug != ""` >> show current slug link
- ELSE >> show nothing there. Help text would start where the link sit.

 When a user opens that page, and the slug was set they see the link, if not set no link. After submission, when slug is set, can show the link.

src/pages/Admin/Charity/EditProfile/useEditProfile.ts
```diff
@@ -53,9 +53,9 @@ export default function useEditProfile() {
         return showModal(TxPrompt, { error: "No changes detected" });
       }
 
-      if (update.slug !== initial.slug) {
+      //dont check hit if unsetting
```
ap-justin, line 56: fixes bug trying to query empty slug `endowment?slug=`

src/pages/Admin/Charity/EditProfile/Slug.tsx
ap-justin: @SovereignAndrey, @misicnenad, implemented live example 
no slug provided
image

slug provided
image

SovereignAndrey: @ap-justin After playing with the live version a bit I'd say let's revert that commit (4f408404) and just go with how it was. Tim can provide more clarity as he gets feedback from nonprofits using the form if there's still confusion or ways to improve it. 👍

SovereignAndrey: @ap-justin After playing with the live version a bit I'd say let's revert that commit (4f408404) and just go with how it was. Tim can provide more clarity as he gets feedback from nonprofits using the form if there's still confusion or ways to improve it. 👍

SovereignAndrey (approved):

PR #2878, opened 2024-04-15
Include tip amount to crypto donation

jp-mariano:
Resolves BG-1291

When donating crypto and a tip is included, the actual transfer transaction's amount does not include the tip amount. However, the DB record does include the tip amount. This might introduce some "not enough tokens error" later on if left unchecked.

- try to donate with tip included and verify that the gross amount for the transfer includes the tip

## UI changes for review
No major UI changes.


misicnenad (approved):
> This might introduce some "not enough tokens error" later on if left unchecked.

As well as leaving BG without a source of revenue :smile:

ap-justin (approved):

PR #2891, opened 2024-04-16
BG-1295: Donation increment buttons

misicnenad:
I tried to avoid using `watch, setValue, trigger` and instead to use `useController`, but it then wouldn't update the amount input value. I guess `register(field)` and `useController` don't play nicely when targeting the same field.

- go to http://localhost:4200/donate/53
- switch to USD
- **verify increment buttons appear**
- switch to any other currency
- **verify no increment buttons appear**

## UI changes for review

USD:
image

Non-USD:
image

### DIFF
--- a/src/components/donation/Steps/DonateMethods/Stripe/Incrementers.tsx
+++ b/src/components/donation/Steps/DonateMethods/Stripe/Incrementers.tsx
@@ -15,6 +15,7 @@ function Incrementer({ value }: { value: number }) {
   const { setValue, trigger, watch } = useFormContext<FormValues>();
   return (
     <button
+      type="button"
       className="text-sm font-medium border border-gray-l4 hover:border-gray-l3 rounded-full w-20 h-10"
       onClick={() => {
         const amount = Number(watch("amount"));
--- a/src/components/donation/Steps/DonateMethods/Stripe/Incrementers.tsx
+++ b/src/components/donation/Steps/DonateMethods/Stripe/Incrementers.tsx
@@ -1,0 +1,31 @@
+import { useFormContext } from "react-hook-form";
+import { FormValues } from "./types";
+
+export default function Incrementers() {
+  return (
+    <div className="flex justify-center gap-3">
+      <Incrementer value={40} />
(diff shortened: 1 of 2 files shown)


src/components/donation/Steps/DonateMethods/Stripe/Incrementers.tsx
```diff
@@ -0,0 +1,31 @@
...
+      <Incrementer value={100} />
+      <Incrementer value={200} />
+    </div>
+  );
+}
+
+function Incrementer({ value }: { value: number }) {
+  const { setValue, trigger, watch } = useFormContext<FormValues>();
+  return (
+    <button
```
ap-justin, line 17:
```suggestion
    <button
       type="button"
```
so that doesn't go to next step (submits) - specify `type="submit"` if the it's the intention

src/components/donation/Steps/DonateMethods/Stripe/Incrementers.tsx
```diff
@@ -0,0 +1,31 @@
...
+    <div className="flex justify-center gap-3">
+      <Incrementer value={40} />
+      <Incrementer value={100} />
+      <Incrementer value={200} />
+    </div>
+  );
+}
+
+function Incrementer({ value }: { value: number }) {
+  const { setValue, trigger, watch } = useFormContext<FormValues>();
```
ap-justin, line 15:
```suggestion
  const { setValue, trigger, getValues } = useFormContext<FormValues>();
```
for one time (on user click) retrieval, it's correct (in context of `react-hook-form` - not necessarily better ) to use [`getValues("path")`](https://react-hook-form.com/docs/useform/getvalues)

ap-justin (dismissed):

src/components/donation/Steps/DonateMethods/Stripe/Incrementers.tsx
misicnenad: This is just what I needed to:
- read the value, but
- not validate the whole form

Thanks!

PR #2896, opened 2024-04-18
Get/set cookie in client

ap-justin:
Towards BG-1283
server set cookie value e.g.
```
...value ; domain: apes-api.better.giving; path="/dev"
```
can't be modified (contrary to the initial expectation) in the client running on domain `staging.better.giving`. 
Instead a distinct cookie is created
```
...value ; domain: apes-api.better.giving; path="/dev"
...value ; domain: staging.better.giving; path="/donate"
```
in which case, cookie `...value ; domain: staging.better.giving; path="/donate"` is never included in succeeding calls with `{credentials:include}` and therefore value is not used. 

* on first visit (no pref yet), save to cookie `prefCode` based on IP
* on succeeding visit, use saved `prefCode` as search param (as cookie with domain `staging.better.giving` is not included when requesting `apes-api.better.giving` ) so ip query need not be run

- [x] verify that selected currency is persisted on page reload for both `/donate` and `widget (embedded on standalone html document)

-

## UI changes for review

### DIFF
--- a/src/components/donation/Steps/DonateMethods/Stripe/Form.tsx
+++ b/src/components/donation/Steps/DonateMethods/Stripe/Form.tsx
@@ -2,6 +2,7 @@ import { yupResolver } from "@hookform/resolvers/yup";
 import CurrencySelector from "components/CurrencySelector";
 import QueryLoader from "components/QueryLoader";
 import { Field, Form as FormContainer } from "components/form";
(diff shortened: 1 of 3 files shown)


ap-justin:
@SovereignAndrey, I think modification of server set cookie value is only possible if the document is rendered in the same domain the cookie is set, that is, for server-rendered UI,

SovereignAndrey (approved):

PR #2944, opened 2024-05-02
Rolled back stripe donation endpoint

jp-mariano:
Related to BG-1318

Rolling back the endpoint used for `stripe-proxy` in APES AWS to `/fiat-donation/stripe` because it already uses API Gateway's stage variables. This stage variable allows us to work on `staging` without overwriting the code in `production`. Setting the API endpoint is also a one-time thing because we don't have to update the endpoint's versions. We would only need to update the Lambda's aliases to point into the correct Lambda versions.

- create a `one-time` donation intent using Stripe
- verify that the intent appears in the `Awaiting Payment` tab
- pay the donation intent
- verify that the donation now appears in the `Pending` tab
- if donating as anon, verify that the record changes `status` and `transactionId` in the holding DB table in APES AWS

## UI changes for review
No major UI changes.

### DIFF
--- a/src/services/apes/apes.ts
+++ b/src/services/apes/apes.ts
@@ -106,7 +106,7 @@ export const apes = createApi({
     }),
     stripePaymentIntent: builder.query<string, StripePaymentIntentParams>({
       query: (data) => ({
-        url: "fiat-donation/stripe/v2",
+        url: "fiat-donation/stripe",
         method: "POST",
         body: JSON.stringify(data),
       }),


SovereignAndrey (approved):

ap-justin:
@jp-mariano @SovereignAndrey, 
Since `$LATEST` previously is used in staging site. I first created a version `x` out of it and pointed `alias:staging`
.i.e.,
```rs
$LATEST --> new endpoint /v2 integration:$LATEST
ver x --> alias: staging (integration:${stageVariables.alias})
ver prod:  --> alias: prod (integration:${stageVariables.alias})
```
to make sure that I don't break staging site as I'm not sure how long would I be testing or how long my PR's going to stay in review. 

Once PR gets merged to staging, I then switched `/v2` integration 
```
$LATEST --> new endpoint /v2 integration:${stageVariables.alias}
```
and pointed `alias:staging` back to `$LATEST`.

QUESTIONS:
1. is overwriting `$LATEST` even if used in staging site is okay? 
2. if `1` is not okay, perhaps I'm doing something wrong with my process (versioning staging code and creating extra `/v2` endpoint to be pointed to `$LATEST` (for changing)). Is there a better way to do this.
3. if creating ephemeral version `(2.)` is the best way fo this (if `(1.)` is not okay. Is reverting back to `/` the same thing as not having stages at all (separate PR for web-app to revert)

ap-justin (approved):

jp-mariano:
@ap-justin,

> 1. is overwriting `$LATEST` even if used in staging site is okay?
> 
> 2. if `1` is not okay, perhaps I'm doing something wrong with my process (versioning staging code and creating extra `/v2` endpoint to be pointed to `$LATEST` (for changing)). Is there a better way to do this?
> 
> 3. if creating ephemeral version `(2.)` is the best way fo this (if `(1.)` is not okay. Is reverting back to `/` the same thing as not having stages at all (separate PR for web-app to revert from temp `/v2` version)

For me, if the change/update is minor and would not result in breaking the `staging` site then it's alright to overwrite directly the `$LATEST` version of the Lambda. If the change/update would break the `staging` site, publishing a new version of that Lambda and pointing it to `staging` alias and using `$LATEST` as a test bed with a *temporary* endpoint is a good way to go, just what you did previously. The point of using stage variables is that once we set an API endpoint, we don't have to change/bump its version every time incremental/big updates are made to that endpoint.

Staying true to the purpose of stage variables, what I would propose is for us to create a third stage like `test`. We currently have `staging` (but uses "dev" in API path due to naming overlap) and `production`. That third `test` stage is where we can do all the breaking changes and still not worry about updating or changing the previously set endpoint unless we're going to rename or nuke it for a better endpoint.

PR #2970, opened 2024-05-08
BG-1331: Move donate page inside user dashboard

ap-justin:
* convert screen media query to container queries
* move page to admin section
image

-

## UI changes for review


src/App/Header/UserMenu/UserMenu.tsx
```diff
@@ -35,6 +35,9 @@ export default function UserMenu() {
   if (user === "loading") {
     return <LoaderRing thickness={10} classes="w-6" />;
   }
+
+  if (location.pathname.startsWith(appRoutes.user_dashboard)) return <></>;
```
SovereignAndrey, line 39: Please revert this. I don't think we should hide the user icon in the header nav, just because the user in the user_dashboard area. Most sites don't do that, largely because I think it causes some disonance for the user when the site's top-most/core layouts change based on user's location on the site.

SovereignAndrey (approved):

SovereignAndrey (changes requested):
See my icon hiding comment reversal request. Thanks!

src/App/Header/UserMenu/UserMenu.tsx
ap-justin: reverted ser https://github.com/AngelProtocolFinance/angelprotocol-web-app/pull/2970/commits/0b282d28df2d60019dfd34d7245c445fab18db2e

SovereignAndrey (approved):

PR #2975, opened 2024-05-09
Endowment link: error marker

ap-justin:
Towards BG-1334 
since user can still access admin, given url `admin/:endowId` it means, user token contains appropriate endowment access. The only reason the link won't show is if the logo and name fetch fails (which is likely not since querying reported endow ids in the same endpoint works)

* show fallback link when fetching logo,name fails
(failed endows here are residual endows in my email) 
image

* check from user if they see fallback link (means they are encountering error fetching link metadata )

-

## UI changes for review

### DIFF
--- a/src/App/Header/UserMenu/EndowmentLink.tsx
+++ b/src/App/Header/UserMenu/EndowmentLink.tsx
@@ -13,7 +13,7 @@ export default function EndowmentLink({ endowId }: Props) {
       queryState={query}
       messages={{
         loading: <Skeleton />,
-        error: <_Link id={endowId} classes="text-red" />,
+        error: <_Link id={endowId} />,
       }}
     >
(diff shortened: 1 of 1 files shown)


SovereignAndrey:
@ap-justin If we're going to show the Link on error, is there a reason to make it red? If the user can click the "dummy" link to get to their admin area OK, then I think that having the different color would confuse them more than help. :thinking:

ap-justin:
> @ap-justin If we're going to show the Link on error, is there a reason to make it red? If the user can click the "dummy" link to get to their admin area OK, then I think that having the different color would confuse them more than help. 🤔

removed ser https://github.com/AngelProtocolFinance/angelprotocol-web-app/pull/2975#:~:text=remove%20red%20color-,568f7a6,-ap%2Djustin%20requested

PR #2976, opened 2024-05-10
BG-1341 Widget adaptations

SovereignAndrey:
Widget/Donations logic changes needed per Chauncey:
- [x] hide direct/SF breakdown lines on summary if SF portion is == 0
- [x] add SF explainer and Laira character to summary view if SF portion > 0
- [x] allow Split & BG Tip screen to be visited by Stocks and DAF for a more uniform donor flow
- [x] if `tip & SF portion === 0`, hide top total line item

### Notes:
- Need to replace the "waving" with "pointing" Laira character image, but can't find it. Will swap in when we have it. Noted a TODO comment in code addressing this.

-

### No Tip && No SF
no-tip-no-sf

### Yes Tip && No SF
yes-tip-no-sf

### No Tip && Yes SF
no-tip-yes-sf

### Yes Tip && Yes SF
yes-tip-yes-sf

### DIFF
--- a/src/components/donation/Steps/common/Summary.tsx
+++ b/src/components/donation/Steps/common/Summary.tsx
@@ -48,28 +48,36 @@ export default function Summary({
       <dl
         className={`text-navy-l1 py-3 gap-y-2 grid grid-cols-[1fr_auto] items-center justify-between border-y border-gray-l4 ${splitClass}`}
       >
-        <dt className="mr-auto text-navy-d4">
-          {props.tip
-            ? `Donation for ${props.tip.charityName}`
-            : "Total donation"}
-        </dt>
-        <Amount amount={props.amount} classes="text-navy-d4" />
-        <div className="flex items-center justify-between col-span-full">
-          <div className="mr-auto flex">
-            <dt className="text-sm mt-2">Sustainability Fund</dt>
-            <Image src={character} className="inline-block px-1 h-8" />
-          </div>
-          <Amount classes="text-sm" amount={locked} />
-        </div>
-        <div className="flex items-center justify-between col-span-full">
-          <dt className="mr-auto text-sm">Direct Donation</dt>
-          <Amount classes="text-sm" amount={liq} />
-        </div>
-        {props.tip && (
-          <div className="col-span-full grid grid-cols-[1fr_auto] border-y border-gray-l4 py-3">
-            <dt className="mr-auto">Donation for Better Giving</dt>
-            <Amount classes="text-sm" amount={props.tip.value} />
-          </div>
+        {(props.tip || locked > 0) && (
+          <>
+            <dt className="mr-auto text-navy-d4">
+              {props.tip
+                ? `Donation for ${props.tip.charityName}`
+                : "Total donation"}
+            </dt>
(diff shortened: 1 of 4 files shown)


src/components/donation/Steps/common/Summary.tsx
```diff
@@ -51,29 +54,43 @@ export default function Summary({
...
-        <div className="flex items-center justify-between col-span-full">
-          <dt className="mr-auto text-sm">Sustainability Fund</dt>
-          <Amount classes="text-sm" amount={locked} />
-        </div>
-        <div className="flex items-center justify-between col-span-full">
-          <dt className="mr-auto text-sm">Direct Donation</dt>
-          <Amount classes="text-sm" amount={liq} />
-        </div>
-        {props.tip && (
+        {locked > 0 && (
```
ap-justin, line 57: comment: 
for me, summary is a step that lays out prior choices I've made 
e.g. opting for `100%` liquid  and `0%` locked and it's justifiable that this choice is presented to me again (summary)

image

compared to (the intent of donating is lost) and it's just assumed that user remembers that this donation is all-in liquid
image

src/components/donation/Steps/common/Summary.tsx
```diff
@@ -51,29 +54,43 @@ export default function Summary({
...
+            </div>
+          </>
+        )}
+        {props.tip && props.tip.value > 0 && (
           <div className="col-span-full grid grid-cols-[1fr_auto] border-y border-gray-l4 py-3">
             <dt className="mr-auto">Donation for Better Giving</dt>
             <Amount classes="text-sm" amount={props.tip.value} />
           </div>
         )}
-        {props.tip && (
```
ap-justin, line 68: the reason a separate `total` row is shown only when there's tip is that collation if only sensible if you have two distinct rows (og & bg)

image

so for donations that doesn't have tip, the UI is redundant
image

src/components/donation/Steps/common/Summary.tsx
SovereignAndrey: This is a fair point.

src/components/donation/Steps/common/Summary.tsx
SovereignAndrey: Per your other comment, if we always show split breakdown then this scenario can'tt happen. I think this is prob what we can do as the best balance of things.

ap-justin: redundant `Total` is unncessary imo as there's only one item (even if that item's detail/partition is enumerated)
image

SovereignAndrey: It's a balance between not showing the breakdown or having it, but having no tip means it's restating the first line. It's like buying a VAT free product in the store. The subtotal and total are the same numbers. One thing the second total line adds is the frequency with which the donation will occur (monthly/one-time). 

I think we can freeze this PR scope for now for review/merge and refine further, only once we know what other chagnes will be needed as a separate ticket.

ap-justin: > It's like buying a VAT free product in the store. The subtotal and total are the same numbers. 

makes sense

PR #3001, opened 2024-05-16
Add back logo fallback

ap-justin:
* update card type
* include logo in fields retrieve

-

## UI changes for review

### DIFF
--- a/src/pages/Marketplace/Cards/Card.tsx
+++ b/src/pages/Marketplace/Cards/Card.tsx
@@ -17,6 +17,7 @@ export default function Card({
   active_in_countries = [],
   name,
   card_img,
+  logo,
   id,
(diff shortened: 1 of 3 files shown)


SovereignAndrey:
@ap-justin I think we can hold off on making this change, now that all the published nonprofits have logos copied over as a quick fix. It's better in the long-run if we can enforce the logos to be used only on profile or for purely logo needs vs marketplace cards.

ap-justin:
> @ap-justin I think we can hold off on making this change, now that all the published nonprofits have logos copied over as a quick fix. It's better in the long-run if we can enforce the logos to be used only on profile or for purely logo needs vs marketplace cards.

got this ser, i'll also revert adding logo as fallback in algolia records

PR #3006, opened 2024-05-17
BG-1354: new account logins is not proceeding

ap-justin:
`endows` claim was incorrectly typed to be `string` where it is actually `string | undefined` (only set when there is an endowment credential)

``` js
//pre-token-gen-trigger.js
if (items.length > 0) {
            event.response = {
                claimsOverrideDetails: {
                    claimsToAddOrOverride: { endows: items.map(({ endowID }) => endowID).join(",") },
                },
            };
        }
```

* change type and update 

-

## UI changes for review

### DIFF
--- a/src/slices/auth.ts
+++ b/src/slices/auth.ts
@@ -30,7 +30,7 @@ export const loadSession = createAsyncThunk<User, AuthUser | undefined>(
 
       type Payload = {
         /** csv */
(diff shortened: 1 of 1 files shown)


SovereignAndrey (approved):
Great catch with the split failure! :muscle:

PR #3069, opened 2024-06-17
Custom tokens UI

ap-justin:
Towards BG-1392
add not listed option (for chains with `erc20` and `cw20` only for now - future: solana `spl`)
image

clicking opens search modal
image

image

cliking `proceed` sets the custom token as the selected token (formatted as `type TokenWithAmount`)
image

-

## UI changes for review


src/components/TokenField/TokenOptions.tsx
```diff
@@ -78,6 +93,37 @@ export default function TokenOptions({ classes = "", selectedChainId }: Props) {
...
+        <button
+          onClick={() =>
+            showModal(TokenSearch, {
+              coingeckoPlatformId,
+              onSubmit: (token, details) => {
+                const option: TokenWithAmount = {
+                  approved: false,
+                  decimals: details.decimals,
+                  logo: details.logo,
+                  min_donation_amnt: 0,
```
ap-justin, line 106: since this token is not in our official tokens list, `min_donation_amnt` is unknown hence temp `0`

ap-justin:
@SovereignAndrey @jp-mariano, opening UI for initial review though not mergeable as custom tokens aren't supported yet in our endpoints. Also coingecko doesn't have testnet apis so need to test this with `PUBLIC_ENVIRONMENT=production`

src/components/TokenField/TokenOptions.tsx
SovereignAndrey: @ap-justin The `min_donation_amnt` for unsupported tokens needs to be higher than normal tokens I think. At the very least, we should set the min tokens required based on the current USD value of the token. All of that would mean that we need a price API though to enable it. 

CoinMarketCap has a free tier we could use solely for unlisted coin price checks (after they are chosen)? https://coinmarketcap.com/api/pricing/  Wdyt?

ap-justin: @SovereignAndrey 
> We should set the min tokens required based on the current USD value of the token based on some cutoff amount(ex. US$25), so if a coin price was US$0.10 we would have a min tokens amnt of: (25 / 0.10) => 250

this makes sense - I implemented this using our existing coingecko `usdRate` endpoint https://github.com/AngelProtocolFinance/angelprotocol-web-app/pull/3069/commits/b415126cc3992d7bab3441ad6fdaad404afb41ce
image

> Total value must be at least ${min_custom_usd} at the time of donation. Failure to send enough tokens may result in your donation being lost or not being processed.

I guess the `min amount` need not more explanation just like our supported tokens

Please note that we also need to update donation intent creation as server validation `getTokenFromDbBySymbol(customSymbol)` would fail as it is not listed

SovereignAndrey: Perfect!

SovereignAndrey (approved):

PR #3235, opened 2024-08-14
Prep to delegate QR flow to 3rd party

ap-justin:
Towards BG-1547

* remove hardcoded extra coingecko tokens, as they are not present in our db tables and would just fail [check](https://github.com/AngelProtocolFinance/devops/blob/main/src/APES/lambdas/crypto-intent-creator/index.mjs#L148). propagating `coingecko id` thru intent creation, scraping, resume  would be unpractical work and rather best spent into integration to 3rd party provider
* new chain terms `tier1`, `tier2` instead of supported/unsupported 

-

## UI changes for review


src/schemas/string.ts
```diff
@@ -51,7 +51,7 @@ export const segment = Yup.string()
     (v) => !v?.startsWith(".") && !v?.endsWith(".")
   );
 
-export function walletAddrPatten(chainId: SupportedChainId) {
+export function walletAddrPatten(chainId: Chain.Id.Tier1) {
```
jp-mariano, line 54: Just a chance to correct the typo here to `walletAddrPattern`.

jp-mariano (dismissed):
Aside from the typo correction ~~and bringing this branch up to date with `master`~~, all's looking good! Using tiers as labels for chains is great because we technically all have "support" for them. 🥇

PR #3575, opened 2025-02-27
Hot Fix - Inconsistent search results

jp-mariano:
Resolves BG-1751

## Bug
When a search query is made in the home page, there is an additional param added `fund_opt_in` which essentially filters out some NPOs.

## Fix
Removed the `fund_opt_in` param which should make search results in the home page and marketplace return the same. To test, search for `"test"` in the marketplace first then try the same with home page. With the fix implemented the results should be the same.


ap-justin (approved):
⭐️ nice

PR #3592, opened 2025-04-03
Donation message updates for crypto and Stripe pipeline

jp-mariano:
Resolves BG-1786

## Description
wip


src/routes/api.nowpayments-webhook.$stage/handlers/settled.ts
```diff
@@ -205,5 +206,22 @@ export const handleSettled = async (payment: NP.PaymentPayload) => {
     } as OnHoldDonation.PrimaryKey,
   });
 
+  /** donation message */
+  if (order.kycEmail && order.donor_public) {
```
ap-justin, line 210: @jp-mariano, we need to create donation message per npo + extra copy for the fund
https://linear.app/better-giving/issue/BG-1624/add-donor-message-option-in-form-add-donor-wall-listing-donations-and#comment-0d26c593

src/.server/donation-messages.ts
```diff
@@ -0,0 +1,37 @@
...
+  });
+  const { Items, LastEvaluatedKey } = await apes.send(cmd);
+
+  const items = (Items || []) as DonationMessage.DBRecord[];
+  const nextPageKey = LastEvaluatedKey
+    ? Buffer.from(JSON.stringify(LastEvaluatedKey)).toString("base64")
+    : undefined;
+
+  return {
+    items: items.map(({ PK, SK, gsi1PK, gsi1SK, ...i }) => i),
```
ap-justin, line 34:
```suggestion
    items: items.map(({ PK, SK, gsi1PK, gsi1SK, donor_id, ..i }) => i),
```
we should make sure not to expose user email addresses - but don't worry about the client, it is handled in separate PR #3593

ap-justin (changes requested):

src/routes/api.nowpayments-webhook.$stage/handlers/settled.ts
jp-mariano: > for stripe,chariot, would this mean that donation messages are sent in final-recorder?

Yes, with the new request it would be better to handle donation message/s creation in the `final-recorder` directly. Will update this accordingly.

ap-justin (approved):

PR #3599, opened 2025-04-23
Cognito signup - new attribute `referral_id`

jp-mariano:
Resolves BG-1801

## Updates
- The `cognito.signup` function is updated to include the optional custom attribute `custom:referral_id`.
- AWS Cognito User pools for both `prod` and `dev` are now also updated to include this custom attribute.


src/.server/auth/cognito.ts
```diff
@@ -174,20 +174,27 @@ class Cognito extends Storage {
...
       lastName: string;
+      "custom:referral_id"?: string;
       "custom:user-type": string;
     }
   ) {
+    const userAttributes = [
+      { Name: "family_name", Value: attributes.lastName },
+      { Name: "given_name", Value: attributes.firstName },
+      { Name: "email", Value: username },
+      attributes["custom:referral_id"] && {
```
ap-justin, line 185: @jp-mariano, `referral_id` is internally generated and included as custom attribute on user signup

ap-justin (changes requested):
@jp-mariano, the referral id should be internal generated and would not come fromusez

src/.server/auth/cognito.ts
jp-mariano: Got it, will update accordingly. 👍🏾

ap-justin (approved):

PR #3606, opened 2025-05-12
Email newsletter subscription - HubSpot

jp-mariano:
Related to BG-1802

## Description
Lambda function `hubspot-email-subs` has been migrated from AWS to web app via `newsletterSubscribe()` API.


src/api/action/newsletter-subscribe.ts
```diff
@@ -0,0 +1,16 @@
+import {
+  HUBSPOT_NEWSLETTER_FORM_ID as formId,
+  HUBSPOT_NEWSLETTER_PORTAL_ID as portalId,
+  HUBSPOT_ACCESS_TOKEN as token,
+} from "constants/env";
+import { hsNewsletterSubs } from "../api";
+
+export async function newsletterSubscribe(email: string) {
```
ap-justin, line 8: since `newsletterSubscribe` is specific to news letter subscription, can hardcode `form-id` and `portal-id` (no need for env ). also `HUBSPOT_NEWSLETTER_PORTAL_ID` should be server side env variable - this in turn would make `.server` directory better place for `function newsletterSubscribe`

ap-justin (changes requested):

